PR #4, opened 2018-03-03
fix: export NgxsPlugin interface

SebastianM:


amcdnl:
Thanks for your contribution!!!

PR #6, opened 2018-03-03
feat: support redux-devtools

SebastianM:
Adds support for the redux-devtools.

This fixes also the type for the `plugins` array in the options. It was not possible to provide a class name in the configuration.

Open issue: I'm not able to inject `Ngxs` into the Plugin (it's always `undefined`). So I'm not able to set the initial state.

### DIFF
--- a/README.md
+++ b/README.md
@@ -479,6 +479,24 @@ describe('Zoo', () => {
 });
 ```
 
+### Redux Devtools
+
+To enable support for the [Redux Devtools extension](http://extension.remotedev.io/), add the following plugin to your `forRoot` configuration:
+
+```javascript
+import { ReduxDevtoolsPlugin } from 'ngxs';
+
+@NgModule({
+    imports: [
+        NgxsModule.forRoot([
+            ZooStore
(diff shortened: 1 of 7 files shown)


amcdnl:
WOAH! YOU ARE EPIC! I WAS LITERALLY JUST ABOUT TO START THIS!

SebastianM:
Thanks man! 😄I really like the design! I'm sure this will be a popular project.

amcdnl:
Maybe its undefined because its circular?

SebastianM:
Hmm no don't think so. Got no error and I also tried to inject Injector - was also undefined without errors. Haven't had the time yet to do more research.

amcdnl:
Maybe I need to export the services in my ngxs module?

amcdnl:
Fixed it :)

SebastianM:
Cool! Btw: I think that's the line that prevents DI in plugins: https://github.com/amcdnl/ngxs/blob/master/src/plugin-manager.ts#L12

PR #7, opened 2018-03-03
chore(testing): sample-test

deebloo:
update some tsconfigs and adds a sample unit test


amcdnl:
I just added some too!

amcdnl:
Looks like that test is failing. I pushed a commit excluding it for now.

PR #10, opened 2018-03-04
fix: Spelling and Capitalization

BeastCode:
Most of the fixes are small spelling mistakes. One thing you might want to review is I changed LocalStorage to be pascal case, example:  
LocalstoragePluginOptions  -> LocalStoragePluginOptions
so it's not mixed case.  /Mike

### DIFF
--- a/README.md
+++ b/README.md
@@ -165,7 +165,7 @@ Those are reserved for `Action`. Similarly actions should not mutate state.
 Lets say for our `NewAnimal` event we want to reach out to the backend and save the
 new animal before we add it to the UI.
 
-Our stores can also participate in depedency injection so when we want to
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -401,7 +401,7 @@ export class LoggerPlugin implements NgxsPlugin {
 }
 ```
 
-This pattern allows us to define options while presevering the constructor
+This pattern allows us to define options while persevering the constructor
```
duanemck: I think this should be 'preserving'

BeastCode: Oh, I see what you're saying

BeastCode: Reverted back to the original line and amended the commit.

amcdnl:
No one ever called me a word smith hahaha. Thanks!

PR #33, opened 2018-03-05
chore: event vs action

BeastCode:
If I understand ngxs correct, you're always dispatching an event and then you'll use the action/mutation decorators to handle the events. Which is different from ngrx, where you dispatch an action.

It seems like the concept is mixed in the source code. I renamed some variables based on the above statement. If I'm wrong, please just ignore this PR. /Mike

### DIFF
--- a/docs/concepts/README.md
+++ b/docs/concepts/README.md
@@ -2,8 +2,8 @@
 There are 4 major concepts to NGXS:
 
 - Events: Classes dispatched to stores with optional payload
-- Actions: Events that perform async operations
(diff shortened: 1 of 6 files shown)


kylecordes:
I did not scrutinize the lines of code change, but I think it is an excellent idea to clarify the language and naming around this. Naming is infamously one of the handful of hard problems in computer science, and thinking about events versus actions is easier if they are named consistently.

amcdnl:
This is great! Thanks so much.

PR #48, opened 2018-03-06
chore: refactor tooling

### LINKED ISSUE #20: commitlint +  conventional-changelog
automatize changelog generation?
I could add the required config files + devDeps, if there is any interest.

jotatoledo:
Closes #20


jotatoledo:
One thing to notice is the prettier config for ts files. The old prettier script had “e5“ in it. Not sure which option was that supposed to be

PR #55, opened 2018-03-08
feature/generic-store-types

deebloo:
Makes the store decorator generic allowing you to define the state for an individual store


amcdnl:
@deebloo - Can you update the docs too?

deebloo:
@amcdnl Yes I can do that. From my testing this in particular SHOULDN'T break anything but I will certainly update

amcdnl:
Ya I just want to make sure our docs are reflecting latest

PR #57, opened 2018-03-09
feature/plugins: update plugins with proposed api

### LINKED ISSUE #16: forRoot plugins won't work with AoT
I don't think having plugins that call methods to pass options will work with AoT.

deebloo:
@amcdnl putting the work here so you can see it. closes #16 

The API will use a plugins DI token for the plugin classes and each plugin will use it's own *_OPTIONS token to configure the plugin. The plugin can then be provided either manually (not recommended) or use the plugin module with a *.forRoot() method.

```TS
@NgModule({
  imports: [
    NgxsModule.forRoot([TodoStore]),
    LocalStoragePluginModule.forRoot({ ... })
  ]
})
```

the example forRoot implementation for the localstorage plugin looks like:

```TS
export function serialize(val: any) {
  return JSON.stringify(val);
}

export function deserialize(val: any) {
  return JSON.parse(val);
}

@NgModule()
export class LocalStoragePluginModule {
  static forRoot(options: LocalStoragePluginOptions = {}): ModuleWithProviders {
    return {
      ngModule: LocalStoragePluginModule,
      providers: [
        {
          provide: NGXS_PLUGINS,
          useClass: LocalStoragePlugin,
          multi: true
        },
        {
          provide: LOCAL_STORAGE_PLUGIN_OPTIONS,
          useValue: {
            key: options.key || '@@STATE',
            serialize: options.serialize || serialize,
            deserialize: options.deserialize || deserialize
          }
        }
      ]
    };
  }
}
```

NOTE: the important thing here is those Injection Tokens. It doesn't matter how the plugin author provides them. (ex: PluginModule, providePlugin(), etc)

### DIFF
--- a/docs/plugins/README.md
+++ b/docs/plugins/README.md
@@ -1,49 +1,68 @@
 # Plugins
+
 Next lets talk about plugins. Similar to Redux's meta reducers, we have
-a plugins interface that allows you to build a global plugin for your state. 
+a plugins interface that allows you to build a global plugin for your state.
 
-Lets  take a basic example of a logger:
+All you have to do is provide a class to the NGXS_PLUGINS token.
+If your plugins has options associated with it, we suggest defining an injection token and then a forRoot method on your module
 
-```javascript
-import { NgxsPlugin } from 'ngxs';
+Lets take a basic example of a logger:
 
+```TS
+import { Injectable, Inject, NgModule } from '@angular/core';
+import { NgxsPlugin, NGXS_PLUGINS } from 'ngxs';
+
+export const LOGGER_PLUGIN_OPTIONS = new InjectionToken('LOGGER_PLUGIN_OPTIONS');
+
+@Injectable()
 export class LoggerPlugin implements NgxsPlugin {
+  constructor(@Inject() private options: any) {}
+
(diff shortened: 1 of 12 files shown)


deebloo:
@amcdnl now we at least have something concrete to look at when we talk about it

deebloo:
the other thought I had was to provide plugins the way you had it before but plugin options have to be manually provided.

```TS
@NgModule({
  imports: [NgsxModule.forRoot([MyStore], { plugins: [MyPlugin] })],
  providers: [
    {
      provide: LOCAL_STORAGE_PLUGIN_OPTIONS,
      useValue: { ... }
    }
  ]
})
```

SebastianM:
Hmm I've never seen this pattern before 😄 just kidding

deebloo:
😁

amcdnl:
So as a user of a prebuilt plugin, what would it look like to pass custom options?

deebloo:
```TS
imports: [MyPluginModule.forRoot({ config1: 1, config2: 2 })]
```

amcdnl:
@deebloo - Lets update the docs and call it good!

PR #83, opened 2018-03-18
fix(core): class selector

deebloo:
Fixes issue where selecting substate with a class would return undefined. A state "path" is now added to metadata when children are present

The StateFactory now adds the path property to each class on init.

NOTE: may have uncovered another issue with sub states if you include the substate in the forRoot method BEFORE the root


amcdnl:
> NOTE: may have uncovered another issue with sub states if you include the substate in the forRoot method BEFORE the root

That should not be happening because of the topologoical sort.

PR #108, opened 2018-03-20
 feature(core): store select accepts class

deebloo:


src/store.ts
```diff
@@ -62,6 +72,13 @@ export class Store {
     return this.select(mapFn).pipe(take(1));
   }
 
+  /**
+   * Allow the user to subscribe to the root of the state
+   */
+  subscribe(fn?: any): Subscription {
+    return this._stateStream.subscribe(fn);
```
amcdnl: I like this!

PR #116, opened 2018-03-21
chore(core): patch state should be partial

deebloo:
use typescript "Partial" type for patchState method


amcdnl:
Whoa, didn't know about this!

PR #119, opened 2018-03-21
fix(docs): updated bad or missing imports

codediodeio:
Fixed some import issues in the docs.


amcdnl:
Thanks for contributing!

PR #124, opened 2018-03-21
NgxsLoggerPlugin - log action payload if present

bossqone:
Fixes https://github.com/amcdnl/ngxs/issues/110


amcdnl:
Thanks for your contribution!

PR #126, opened 2018-03-21
feat: improved module init

leon:
I was having troubles when having `NgxsModule.forFeature` in the main app.

It seems that the provider store token was being injected multiple times which caused feature's to be loaded twice.

By separating out the feature states into it's own token it seems to be working better.

I Also added some units tests to test out the different setups we can encounter.

The pull request is also tested against `--aot` and everything seems to be working.
@amcdnl  Could you have a look and see if this improves things?

### DIFF
--- a/src/module.ts
+++ b/src/module.ts
@@ -1,13 +1,13 @@
-import { NgModule, ModuleWithProviders, Optional, Inject, SkipSelf } from '@angular/core';
+import { NgModule, ModuleWithProviders, Optional, Inject } from '@angular/core';
 
-import { STORE_TOKEN } from './symbols';
+import { ROOT_STATE_TOKEN, FEATURE_STATE_TOKEN } from './symbols';
 import { StateFactory } from './state-factory';
 import { Actions } from './actions-stream';
 import { Store } from './store';
 import { SelectFactory } from './select';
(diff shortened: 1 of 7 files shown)


deebloo:
so far looks good to me. looks like some minor stuff is breaking in ci. looks like the linter which is odd since prettier should take care of that for you.

src/module.ts
```diff
@@ -83,25 +70,26 @@ export class NgxsModule {
...
-        stores,
         StateFactory,
         PluginManager,
+        ...states,
         {
-          provide: STORE_TOKEN,
-          useValue: stores
+          provide: FEATURE_STATE_TOKEN,
+          multi: true,
+          useValue: states
```
deebloo: does this need to be multi since we are providing it with an array?

src/spec/module.spec.ts
```diff
@@ -0,0 +1,101 @@
+import { NgModule } from '@angular/core';
```
deebloo: yay more tests

amcdnl:
@leon - Can you fix the lint error - `ERROR: src/module.ts[31, 1]: Exceeds maximum line length of 140`

PR #130, opened 2018-03-22
feat: add overloading to select method in Store

jonbretman:
Hey!

I'm currently evaluating using ngxs for a large Angular app and so far it looks great. I'm particularly interested in the type-safety of selecting data from the store and I think this improves it for the `store.select()` approach.

A simple example:
```ts
interface User {
  name: string;
}

interface State {
  users: User[];
}

// this will type check but is wrong as actually userNames$ will be Observable<number>
const userNames$: Observable<string> = store.select((state: State) => state.users.length);
```

With this additional function signature I get this error:
```
[ts]
Type 'Observable<number>' is not assignable to type 'Observable<string>'.
  Type 'number' is not assignable to type 'string'.
```

### DIFF
--- a/src/store.ts
+++ b/src/store.ts
@@ -50,6 +50,8 @@ export class Store {
   /**
    * Selects a slice of data from the store.
    */
+  select<T>(selector: (state: any) => T): Observable<T>;
+  select(selector: string | any): Observable<any>;
   select(selector: any): Observable<any> {
     if (selector[META_KEY] && selector[META_KEY].path) {
       const getter = fastPropGetter(selector[META_KEY].path.split('.'));


leon:
Wouldn't it work if you return a string from the select?

You are trying to force a select that is an Observable<number> to be a Observable<string>
So if you convert the .length into a string it will work.

```
const userNames$: Observable<string> = store.select((state: State) => ''+state.users.length);
```

deebloo:
Yeah it looks like the.  Compiler error is valid in this case

deebloo:
Oh oh I'm sorry I misread.

deebloo:
the points is atm we aren't getting the correct error message.

src/store.ts
```diff
@@ -50,6 +50,8 @@ export class Store {
   /**
    * Selects a slice of data from the store.
    */
+  select<T>(selector: (state: any) => T): Observable<T>;
```
deebloo: what if instead of the return type, "T" was the state and we remove the return signature so the compiler could infer the correct type.

```
select<T>(selector: (state: T) => T);
```

deebloo:
@leon I think this looks good. it woudln't be a breaking change and adds more type safety which I like

leon:
@deebloo I agree.

Do you want to do the merge?

deebloo:
@leon I am checking on thing real quick and then I will

deebloo:
@jonbretman thanks for the contribution!

jonbretman:
@deebloo @leon thanks for the super quick response and for merging - looking forward to using this library more, it seems great so far!

PR #131, opened 2018-03-22
fix: add empty options object, to prevent errors

labs-scnm:
Check if options falsy (eg. `.forRoot()`), than create an empty object to prevent errors. So you can call `.forRoot()` instead of `.forRoot({})` .


leon:
Looks good to me 👍 

What do you think @amcdnl?

PR #135, opened 2018-03-22
fix(core):feature-modules

deebloo:
This should fix issues based on requiring a parent state stream for a feature module.
Now what happens is the state stream is injected and IF there is a parent it uses that stream.


deebloo:
@amcdnl before we merge this in could you give it a try in your project?

yarrgh:
👍 Works as far as I can tell (for both lazy and non lazy loaded modules)

leon:
@deebloo Why did you remove ChromeHeadless?

deebloo:
@leon it isn't needed for circle

amcdnl:
Works! :)

PR #137, opened 2018-03-22
feat: npm auto build

leon:
Publish dev builds to npm so that people can follow along with nightly builds.

The format for the version will follow semver and publish under the current version present in package.json and append `+dev.buildsha1`

example:
```
yarn install @ngxs/core@2.0.0-rc.19+dev.ae34ea2
```

**Note**
This pull request will also switch to the new npm repo `@ngxs/core`

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -68,12 +68,17 @@ jobs:
       - *restore_cache
       - *yarn_install
       - *save_cache
+      - run:
+          name: Compile NGXS
+          command: yarn build
+      - run:
+          name: Create Pack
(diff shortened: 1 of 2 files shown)


deebloo:
I definitely want Austin's feedback before changing the package name

leon:
I don't want to start spiting out build versions to npm before we decide if it's a good idea.

What do you think?

leon:
@amcdnl I think store is more in line with that we do.
Yes I added you to the npm org, did you get an invite?

amcdnl:
Can you resolve conflicts and we can merge this.

PR #142, opened 2018-03-23
Add Angular 6.0.0 and rxjs 6.0.0 support

xmlking:
1. Angular compatible 6.0.0 and 5.x
2. rxjs compatible 6.0-forward-compat.0 || ^6.0.0

for Angular v5 users, suggest Option 1
for Angular v6 users, suggest Option 2

1. Option 1:
    rxjs: "5.6.0-forward-compat.1"

2. Option 2:
    "rxjs": "6.0.0-beta.1",
    "rxjs-compat": "6.0.0-beta.1",

https://github.com/ngxs/core/issues/128


amcdnl:
@xmlking - Does this require users to be running ng6 or is it cross compat with 5?

package.json
```diff
@@ -37,22 +37,22 @@
   },
   "homepage": "https://github.com/amcdnl/ngxs#readme",
   "peerDependencies": {
-    "@angular/core": "^5.0.0",
-    "rxjs": "^5.5.6"
+    "@angular/core": ">=5.0.0 <7.0.0 || >6.0.0-beta <7.0.0",
+    "rxjs": ">=6.0.0 || ^5.6.0-forward-compat.0 || ^6.0.0-beta.1"
```
amcdnl: Can we keep `5.5.6` here for current users or is that a hard req?

src/store.ts
```diff
@@ -1,11 +1,7 @@
...
-import { Observable } from 'rxjs/Observable';
-import { Subscription } from 'rxjs/Subscription';
-import { distinctUntilChanged, catchError, take } from 'rxjs/operators';
-import { forkJoin } from 'rxjs/observable/forkJoin';
-import { map } from 'rxjs/operators/map';
-import { fromPromise } from 'rxjs/observable/fromPromise';
-import { of } from 'rxjs/observable/of';
+import { Observable, Subscription, forkJoin, of } from 'rxjs';
+import { map, distinctUntilChanged, catchError, take } from 'rxjs/operators';
+import { fromPromise } from 'rxjs/internal/observable/fromPromise';
```
amcdnl: Is the `internal` really supposed to be used?

package.json
xmlking: yep it can be 5.5.6

amcdnl: Can you update to include that then?

xmlking: oops , they have to at least go to 5.6.0-forward-compat.0 which is compatible with old libs

amcdnl:
@xmlking - Can you fix the yarn file too?

xmlking:
will fix yarn, let me also do some tests with various versions of angular and rxjs.

src/store.ts
leon: fromPromise has been moved to `from` so we need to change all fromPromise

xmlking: i see this import pattren [here](https://github.com/ReactiveX/rxjs/blob/master/compat/add/observable/fromPromise.ts)
```ts
import { fromPromise } from 'rxjs/internal-compatibility';
```
`from` is also available  from `rxjs`
```ts
import { Observable, Subscription, forkJoin, of, from  } from 'rxjs';
```
@leon is this `from` can convert to ` Observable` ?

amcdnl:
@xmlking - ok good deal. This is a big change so I wanna make sure to not break anyone. Can you rebase when you get a second?

xmlking:
Testing and working with Angular 5.x and 6.x
https://stackblitz.com/edit/angular-ngxs-rxjs6-test

amcdnl:
@deebloo - I think I'd like to put this in 2.1 beta release and publish that in parallel to our 2.0

xmlking:
Agree

deebloo:
@xmlking we just did some significant update to the repo. can you rebase? thanks!

xmlking:
Sure give 1hour

xmlking:
tested with angular 5.x  https://stackblitz.com/edit/angular-ngxs-rxjs6-test

amcdnl:
@deebloo - how do you wanna handle merging this?

amcdnl:
@leon - doesn't this require users to upgrade rxjs now?

kylecordes:
@amcdnl Based on the test StackBlitz above, it appears to work fine with Angular 5 and RxJS 5 (5.6 forward-compat though...)

It remains to be seen exactly how the bulk of the community is going to handle this.  It's certainly going to be a more noticeable transition than the last few major Angular versions. One possible answer: popular libraries upgrade to RxJS 6, but continue to work with forward-compat for quite a while, so that folks who are not yet ready to use RxJS 6 can keep using the latest libraries.

amcdnl:
Right, it requires you to upgrade to 5.6. I suppose this would make it a major version?

kylecordes:
Gitter - Thanks for the tip, will do.

About the versioning, that is an excellent question. I don't know about this nuance of semver and it's not clear from a quick skim of the spec. What we're talking about here is not an API change (the thing that obviously triggers a major version). And it's not even a dependency on a new major version of an underlying library. If a new version of my library takes a peer dependency on version 5.6 instead of 5.7 of another library, does that trigger a new major version of my library? I don't think so, but maybe? Any semver lawyers here? :-)

Koslun:
Not a semver lawyer but I think peer dependencies should be seen as part of the API surface while regular dependencies shouldn't. As regular dependencies will just get updated automatically while peer dependencies will not. After upgrading to a new version I'd still be on version 5.5 and thus getting broken and with some npm warnings.

Think it's more or less the same as compatability with browsers or node engine where I've seen this interpretation a few times before. One example would be Mocha has which essentially did two relatively close releases due to this with [version 4](https://github.com/mochajs/mocha/blob/master/CHANGELOG.md#400--2017-10-02) and especially in [version 5](https://github.com/mochajs/mocha/blob/master/CHANGELOG.md#500--2018-01-17).

Think you already came to the same consensus but figured I'd mention it if not.

Ideally I think all libraries should try to become rxjs version 6 compliant as soon as possible so that those of us that move find it meaningful to move and that most projects aren't stuck using one of the [compat solutions](https://docs.google.com/document/d/12nlLt71VLKb-z3YaSGzUfx6mJbc34nsMXtByPUN35cg/preview) year out or longer. And instead only those that can't move to rxjs 6 are stuck with some kind of compat solution.

arjunyel:
Is there an Agnular 6 example using NGXS? I've never used rxjs-compat before

Issue #138, opened 2018-03-22
Actions Subscribe


mailok:
### Versions

```
* ngxs: 2.0.0-rc.18
* @angular/core: 5.2.9
```

### Repro steps

* Step 1: Defining Async Actions

```
@Action(SearchPeoples)
  serarchPeoples({ getState, setState }: StateContext<ParkStateModel>) {
    const state = getState();
    this.http.get<Search>('https://randomuser.me/api/?results=5').subscribe(value => {
      console.log('updating the state...');
      setState({
        ...state,
        search: value.results
      });
      console.log('status updated.');
    });
  }
```

* Step 2: Creating function to dispatch action and subscribe

```
search() {
    this.store.dispatch(new SearchPeoples()).subscribe(value => {
      console.log('The action is completed.');
    });
  }
```

* Step 3: Call function search

### Observed behavior

In the browser console displays:
![captura](https://user-images.githubusercontent.com/3769277/37792514-5aca8b64-2de2-11e8-9cac-eadf2079a40a.PNG)

### Desired behavior

I hoped log first : **updating the state...**
then : **status updated.**
and finally:  **The action is completed.**

I thought to subscribe only receive a result when the action will update the status.
It is in this way that should work or not?

yarrgh:
The problem here is that the dispatch observable finishes before the observable inside your action since it isn't waiting for the observable inside the action.

What happens if you return the Observable from the action and instead of subscribing to it use a rxjs operator, like `tap()` to do the work inside the observable?

mailok:
@yarrgh sorry, I did not understand, You can give me an example?

yarrgh:
In your action:

```typescript
@Action(SearchPeoples)
  serarchPeoples({ getState, setState }: StateContext<ParkStateModel>) {
    const state = getState();
    return this.http.get<Search>('https://randomuser.me/api/?results=5').pipe(
      tap(value => {
        console.log('updating the state...');
        setState({
          ...state,
          search: value.results
        });
        console.log('status updated.');
      })
    );
  }
```

I've changed the action to return the Observable and doing the logic inside the `tap()` rxjs operator.

mailok:
@yarrgh I get it!. But that gives another problem:

![captura](https://user-images.githubusercontent.com/3769277/37793689-c2620f06-2de5-11e8-842b-d89878758896.PNG)

**it appears that executes the action several times**

yarrgh:
[...]

PR #149, opened 2018-03-24
fix(core): fix dispatch issue

### LINKED ISSUE #145: Async action handler requires subscription in order to execute pipeline
### Versions
```
* ngxs: 2.0-rc20
```
### Repro steps
See the adapted sample:
https://stackblitz.com/edit/ngxs-simple-v2rc20-async-issue?file=app/app.component.ts
All I changed was to make the action handler async and return an observable
* Click the button labelled "Click Me (with subscribe)"
* Observe that the number increments
* Click the button labelled "Click Me (no subscribe)"
* Observe that the number doesn't increment

### Desired behavior
It should not be necessary to subscribe to an dispatch in order for the action handler's observable to be executed.

It seems that this is a regression introduced by the change made here: https://github.com/ngxs/store/commit/0de542362573ac117f0ed31639e6193787dfa0a6

I think that the solution to this should include some unit tests so that it does not regress again.

deebloo:
should fix issue requiring you to subscribe and also where dispatch would fire multiple times. closes #145 #138


src/spec/dispatch.spec.ts
```diff
@@ -0,0 +1,89 @@
...
+
+    store.dispatch(new Increment()).subscribe();
+    store.dispatch(new Increment()).subscribe();
+    store.dispatch(new Increment());
+    store.dispatch(new Increment());
+    store.dispatch(new Decrement());
+
+    store.select(MyState).subscribe(res => {
+      expect(res).toBe(3);
+    });
```
deebloo: quick note here. If the dispatch was broken it would either not dispatch at all or dispatch twice. so the state will be incorrect if these don't work

Issue #140, opened 2018-03-22
NgxsLocalStoragePluginModule always picks localStorage as storage option


perjansson:
### Versions

```
* ngxs: 2.0.0-rc.18
* @angular/core: 5.2.1
```

### Repro steps

When configuring `NgxsLocalStoragePluginModule` to use sessionStorage as storage engine like this is does not work.

```
NgxsLocalStoragePluginModule.forRoot({
	key: STORAGE_KEY_NGXS_STORE,
	storage: sessionStorage,
})
```

### Observed behavior

It still saves state into `localStorage`.

### Desired behavior

Saves state into `sessionStorage`

amcdnl:
Can you see whats getting set here? https://github.com/ngxs/core/blob/master/src/plugins/localstorage/localstorage.plugin.ts#L14

perjansson:
@amcdnl It's the localStorage getting set there. I added one item to `sessionStorage` with the name of `"sessionStorage"` and one item to `localStorage` with the name of "`localStorage"`. When I debug I can see that it is the Storage that has the item "localStorage" that comes in option.storage and is assigned the engine.

<img width="1640" alt="engine" src="https://user-images.githubusercontent.com/1557938/37851470-49a9ccde-2edf-11e8-86de-ba7c7e77b39a.png">

un33k:
@amcdnl It turns out that Angular passes `null` for anything that cannot be serialized.  In this case, localStorage and sessionStorage seem not to be serializable.

Since I recommended this approach, I went ahead and fixed the issue and while at it, I cleanup storage a bit as well.  Since we support `localStorage` & `sessionStorage`, I renamed the module `storage` instead of `localstorage`.

Docs were modified to reflect the changes.
Added tests for sessionStorage as well.

Hope this helps.

Here is the PR:  https://github.com/ngxs/store/pull/152

PR #152, opened 2018-03-25
Refactor(storage-plugin): Rework storage

un33k:
Restrict storage types to localStorage and sessionStorage.
Reintroduce NgxsStorageStrategy.

Fix for https://github.com/ngxs/store/issues/140


amcdnl:
We want to allow you to pass in the storage strategy you want so that you can swap it out for various different platforms like mobile.

amcdnl:
We wanted to expose it where you can pass the different storage engines if you are on different platforms, for instance mobile.

Reading through the issue, I wonder if is another way to handle this.

un33k:
I put inline instructions in the options. 

For mobile there is no localStorage or sessionStorage. 

So a compatible store engine would be used like:

Import {customLocalStoage} from ‘some-storage-package’;

global.localStorage = customLocalStorage. 

Ditto for sessionStorage. 

That is done in the app module. 

The PR takes care of everything else. 

Val

deebloo:
I don't like having to globally set the storage engine. that feels very hacky. I think we can provide a way to make this happen but i don't think this is it

deebloo:
I think we are better suited providing an injection token and interface for what storage looking like that a user can then provide whatever solution they want and we can still provide an enum for the defaults of localstorage and sessionstorage.

deebloo:
@un33k I am closing this for now since the repo is going through some significant restructuring and we can address after that get's merged in

un33k:
@deebloo @amcdnl Sounds good.

PR #153, opened 2018-03-25
Break into separate packages for store and plugins

leon:
We might need to do some extra work with the build pipeline, but this should make it easier for building out the plugins with the form and publish each plugin as it's own npm package


deebloo:
after the ci stuff is working we should be able to resolve any issues (if there are any) and reset master to this

PR #154, opened 2018-03-25
refactor(storage): update storage plugin

deebloo:
Updates storage plugin to allow you to pass in your own storage implementation.

@un33k wanted your input since you had opened a PR for this previously. wanted to know what you thought. should address #140


deebloo:
@un33k wanted your input since you had opened a PR for this previously. wanted to know what you thought

packages/storage-plugin/src/storage.module.ts
```diff
@@ -1,19 +1,40 @@
...
   key: '@@STATE',
-  storage: localStorage,
+  storage: StorageOption.LocalStorage,
   serialize: JSON.stringify,
   deserialize: JSON.parse
 };
 
+export function engineFactory(engine: StorageOption): StorageEngine {
+  if (engine === StorageOption.LocalStorage) {
+    return localStorage;
```
un33k: We are still bound to the global context.  

```
return localStorage 
```

is the same as 
```
return global.localStorage
```

Where is the handling of the custom storage engine, let's say on mobiles via NativeScript.

packages/storage-plugin/src/symbols.ts
```diff
@@ -24,3 +30,11 @@ export interface NgxsStoragePluginOptions {
...
 
 export const NGXS_STORAGE_PLUGIN_OPTIONS = new InjectionToken('NGXS_STORAGE_PLUGIN_OPTION');
+
+export const STORAGE_ENGINE = new InjectionToken('STORAGE_ENGINE');
+
+export interface StorageEngine {
+  getItem(key): any;
+  setItem(key, val): void;
+  removeItem(key): void;
+}
```
un33k: we can add `length`, `key` and `clear` as well to make it fully compatible.

un33k: Everything else looks great. We just need to test the non-default option, for example sessionStorage.

docs/plugins/storage.md
```diff
@@ -0,0 +1,62 @@
...
+    NgxsModule.forRoot([]),
+    NgxsLocalStoragePluginModule.forRoot({
+      /**
+       * Default key to persist. You can pass a string or array of string
+       * that can be deeply nested via dot notation.
+       */
+      key: '@@STATE',
+
+      /**
+       * Storage strategy to use. Thie defaults to LocalStorage but you
```
un33k: simple typo  `This` for `Thie`.

packages/storage-plugin/src/symbols.ts
deebloo: Ill make sure it is tested. Just wanted your input before I went to far

packages/storage-plugin/src/storage.module.ts
deebloo: See StorageEngine

docs/plugins/storage.md
```diff
@@ -0,0 +1,62 @@
...
+       * can pass SessionStorage or anything that implements the localStorage API.
+       */
+      storage: StorageOption.LocalStorage,
+
+      /**
+       * Custom deseralizer. Defaults to JSON.parse
+       */
+      deserialize: JSON.parse,
```
amcdnl: @un33k - If passing a different storage engine breaks, wouldn't this break too?

docs/plugins/storage.md
```diff
@@ -0,0 +1,62 @@
...
+
+      /**
+       * Custom serializer, defaults to JSON.stringify
+       */
+      serialize: JSON.stringify,
```
amcdnl: Trailing comma

deebloo:
@amcdnl don't merge yet. still adding tests and docs around it

docs/plugins/storage.md
deebloo: this is actually something I am looking into now. writing tests around a simple custom storage option now

deebloo:
@amcdnl @un33k I feel pretty good about this now. wanted to know if you all wanted to take one more look at it

docs/plugins/storage.md
deebloo: @amcdnl so it will break depending on what you do with your implementation. it is easy enough to pass a custom serializer/deserialzer though so im not to concerned. maybe a documentation issue more then anything else

un33k:
@deebloo Looks fantastic.  One quick note on the `storage?: StorageOption;`.  It wonder if this is redundant for custom storage engines. If so, the comment may need to reflect it.

Great job guys.

deebloo:
It would be yes. I can certainly add clarification in the docs

PR #163, opened 2018-03-26
fix(storage): fix regression with AOT

deebloo:


packages/storage-plugin/src/storage.module.ts
```diff
@@ -30,26 +35,27 @@ export function engineFactory(engine: StorageOption): StorageEngine {
...
         {
           provide: NGXS_PLUGINS,
           useClass: NgxsStoragePlugin,
           multi: true
         },
         {
           provide: NGXS_STORAGE_PLUGIN_OPTIONS,
-          useValue: combinedOptions
+          useFactory: storageOptionsFactory,
+          deps: ['USER_OPTIONS']
```
amcdnl: Can we make this string a const?

deebloo:
updated

PR #164, opened 2018-03-26
added form-plugin and docs. modified integration app. to test form-plugin

tsawan:
added form-plugin and docs. modified integration app. to test form-plugin


deebloo:
@tsawan can you add a unit test?

deebloo:
@amcdnl we should also decide if we want to pull more plugins into core or jsut have folks publish them individually

yarrgh:
My opinion is that if there's an obvious benefit that everyone or the majority can benefit from, then it should be considered being added to the core repository. 

Plugins that target a subset of users generally should be considered being maintained independently.

amcdnl:
@deebloo - This is something I wanted to add and I think its a pretty core item for people so I'd like to have this one in core.

tsawan:
@deebloo it will take me a day or two to get back to it, then can try to add some unit test(s).

tsawan:
@amcdnl does it need to pass all ci checks before it can be merged? not sure I can fix the failed lint test.

deebloo:
Yes all checks need to pass

integration/app/todo.state.ts
```diff
@@ -33,8 +34,21 @@ export class TodoState {
...
+    todo: [],
+    pizza: { model: undefined }
+  },
   children: [TodoState]
 })
-export class TodosState {}
+export class TodosState {
+  @Action(SetPrefix)
+  setPrefix({ getState, setState, patchState }) {
+    const state = getState();
```
deebloo: Remove unused function

integration/app/app.component.ts
```diff
@@ -1,13 +1,24 @@
...
   template: `
     <div class="todo-list">
+      <div>
+        <h3>Reactive Form</h3>
+        <form [formGroup]="pizzaForm" novalidate (ngSubmit)="onSubmit()" ngxsForm="todos.pizza">
+            <input type="text" formControlName="toppings" />
+            <br><hr>
+            <button type="submit">set Dirty false</button>
+            <button type="button" (click)="onPrefix()">Set Prfix</button>
+        </form>
```
yarrgh: Lint is complaining about the trailing whitespace on line 19. Just need to remove it.

integration/app/todo.state.ts
tsawan: do u mean setPrefix? The action setPrefix is used in app.component.ts?

PR #172, opened 2018-03-26
fix(storage): fix lazy loaded storage

### LINKED ISSUE #122: After reload state form local storage, state$.subscribe() give back an empty state
I use the `NgxsLocalStoragePluginModule`. After reload the app, the state will load from local storage correctly (state show correctly on dev tools), but the subscribe give me an null value.

```ts
@Select(TmpState) tmpState$: Observable<any>;

constructor() {
  this.tmpState$.subscribe((state) => {
    console.log(state); // null
  });
}
```

Any idea what I'm doing wrong?

amcdnl:
Fixes lazy loaded storages not populating on first load. closes #122

### DIFF
--- a/packages/storage-plugin/src/storage.plugin.ts
+++ b/packages/storage-plugin/src/storage.plugin.ts
@@ -13,7 +13,8 @@ export class NgxsStoragePlugin implements NgxsPlugin {
 
   handle(state, event, next) {
     const options = this._options || <any>{};
-    const isInitAction = getActionTypeFromInstance(event) === '@@INIT';
+    const actionType = getActionTypeFromInstance(event);
+    const isInitAction = actionType === '@@INIT' || actionType === '@@UPDATE_STATE';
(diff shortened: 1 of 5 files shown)


deebloo:
Or I guess we can use the static member on the action class

packages/storage-plugin/src/storage.plugin.ts
```diff
@@ -13,7 +13,8 @@ export class NgxsStoragePlugin implements NgxsPlugin {
 
   handle(state, event, next) {
     const options = this._options || <any>{};
-    const isInitAction = getActionTypeFromInstance(event) === '@@INIT';
+    const actionType = getActionTypeFromInstance(event);
+    const isInitAction = actionType === '@@INIT' || actionType === '@@UPDATE_STATE';
```
deebloo: Can the init stuff go in the construction rather then checking for the init event?

deebloo:
@amcdnl fix looks like it works! after updating to use the type check to use the static property on the class id say this is good to go!

PR #176, opened 2018-03-27
feat(plugin): router plugin (work in progress)

leon:
A first wip of the router plugin.
not working quite yet.

## Questions
Since this is the first plugin which should have it's own state, I'm not sure on how we should best attack this.

I'm thinking maybe it should have it's own `@State` which users can use and get selectors from.
But I'm currently trying to get it to work with the module setup.

Otherwise I've tried to keep it clean and a exact port of the ngrx version so that we can try to keep them in sync if bugs appear.

@amcdnl @deebloo How do you think the best way would be to incorporate a `@State` into a plugin?

### DIFF
--- a/packages/router-plugin/src/router.state.ts
+++ b/packages/router-plugin/src/router.state.ts
@@ -1,0 +1,42 @@
+import { RouterStateSnapshot } from '@angular/router';
+import { Action, Selector, State, StateContext } from '@ngxs/store';
+import { RouterAction, RouterCancel, RouterError, RouterNavigation } from './router.actions';
+
+export type RouterStateModel<T = RouterStateSnapshot> = {
+  state: T;
+  navigationId: number;
+};
+
+@State<RouterStateModel>({
+  name: 'router',
+  defaults: {
+    state: undefined,
+    navigationId: undefined
+  }
+})
+export class RouterState {
+  @Selector()
+  static state(state: RouterStateModel) {
+    return state && state.state;
+  }
+
+  @Selector()
+  static url(state: RouterStateModel) {
+    return state && state.state && state.state.url;
+  }
+
+  @Selector()
+  static paramMap(state: RouterStateModel) {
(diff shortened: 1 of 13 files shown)


packages/router-plugin/src/router.state.ts
```diff
@@ -0,0 +1,42 @@
+import { RouterStateSnapshot } from '@angular/router';
+import { Action, Selector, State, StateContext } from '@ngxs/store';
+import { RouterAction, RouterCancel, RouterError, RouterNavigation } from './router.actions';
+
+export type RouterStateModel<T = RouterStateSnapshot> = {
```
deebloo: what is that I see! does typescript allow for default generic values now!?!?!

deebloo:
@leon I think it should be fine as long as the plugin module uses .forFeature

amcdnl:
@leon did u see my gist yesterday on this?

amcdnl:
https://gist.github.com/amcdnl/1c935e9773242bd61a5e86fce9baf3ea

amcdnl:
I’d like to use the state manager and register the state automatically rather than force the user to pass it

amcdnl:
I’d like to only put the important info in the state, that route object is crazy big.

amcdnl:
Also we can merge this but I don’t want to publish til post 2.0 which is like soon

packages/router-plugin/src/router.state.ts
leon: @deebloo yes it has been in typescript since 2.3 :)

deebloo: Well son of a b

leon:
@amcdnl Totally missed your gist :)

We can totally wait for after 2.0 this was more a "get something started" pull request.

The ngrx version is quite complex, but it handles things like only navigating when needed, early instantiating so as to not miss a navigation event.

Maybe we could have a little workshop around it?

amcdnl:
Lets close this for now.

PR #187, opened 2018-03-28
use withLatestFrom instead of map

amitport:
no need to re-listen to the observable

### DIFF
--- a/docs/concepts/store.md
+++ b/docs/concepts/store.md
@@ -58,9 +58,9 @@ select in the chain like:
 
 ```javascript
 import { Store } from '@ngxs/store';
-import { AddAnimal } from './animal.events';
 import { withLatestFrom } from 'rxjs/oprators';
 import { Observable } from 'rxjs/Observable';
+import { AddAnimal } from './animal.events';
 
 @Component({ ... })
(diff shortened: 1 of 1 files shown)


docs/concepts/store.md
```diff
@@ -59,16 +59,21 @@ select in the chain like:
 ```javascript
 import { Store } from '@ngxs/store';
 import { AddAnimal } from './animal.events';
+import { withLatestFrom } from 'rxjs/oprators';
+import { Observable } from 'rxjs/Observable';
```
deebloo: Fix order of imports

docs/concepts/store.md
```diff
@@ -66,9 +68,12 @@ export class ZooComponent {
   constructor(private store: Store) {}
 
   addAnimal(name) {
-    this.store.dispatch(new AddAnimal(name)).map(() => this.animals$).subscribe((animals) => {
-      this.form.reset();
-    });
+    this.store.dispatch(new AddAnimal(name))
+      .pipe(withLatestFrom(this.animals$))
+      .subscribe(([,animals]) => {
```
amcdnl: leading comma here. Can you fix?

PR #203, opened 2018-04-01
feat: cancellable actions

amcdnl:
Adds ability to make actions cancellable.

### DIFF
--- a/packages/store/src/state-factory.ts
+++ b/packages/store/src/state-factory.ts
@@ -5,7 +5,7 @@ import { of } from 'rxjs/observable/of';
 import { shareReplay, takeUntil, map } from 'rxjs/operators';
 import { forkJoin } from 'rxjs/observable/forkJoin';
 
-import { META_KEY, StateContext } from './symbols';
+import { META_KEY, StateContext, ActionOptions } from './symbols';
 import { topologicalSort, buildGraph, findFullParentPath, nameToState, MetaDataModel, isObject } from './internals';
 import { getActionTypeFromInstance, setValue, getValue } from './utils';
 import { ofAction } from './of-action';
@@ -15,6 +15,7 @@ export class StateFactory {
   get states() {
     return this._parentFactory ? this._parentFactory.states : this._states;
   }
+
   private _states = [];
 
   constructor(
@@ -117,7 +118,9 @@ export class StateFactory {
 
           if (result instanceof Observable) {
             result = result.pipe(
-              actionMeta.options.cancellable ? takeUntil(actions$.pipe(ofAction(action.constructor))) : map(r => r)
+              (<ActionOptions>actionMeta.options).takeLast
+                ? takeUntil(actions$.pipe(ofAction(action.constructor)))
+                : map(r => r)
             ); // act like a noop
           }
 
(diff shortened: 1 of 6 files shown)


leon:
Looks good, but we need some unit tests to make sure it works.
what you say @amcdnl?

packages/store/src/state-factory.ts
```diff
@@ -93,7 +98,7 @@ export class StateFactory {
     }
   }
 
-  invokeActions(getState, setState, dispatch, action): any[] {
+  invokeActions(getState, setState, dispatch, actions$, action) {
```
deebloo: we should have type annotations

deebloo:
@amcdnl I wrote a test that proves that the  manual way works. when I try to swap it out for cancellable I get:

```
Failed: Cannot read property 'type' of undefined
        TypeError: Cannot read property 'type' of undefined
            at http://localhost:9876/_karma_webpack_/webpack:/packages/store/src/of-action.ts:12:51
            at Array.forEach (<anonymous>)
            at Object.ofAction packages/store/src/of-action.ts:12:16)
            at StateFactory.webpackJsonp../packages/store/src/state-factory.ts.StateFactory.invokeActions packages/store/src/state-factory.ts:120:72)
            at http://localhost:9876/_karma_webpack_/webpack:/packages/store/src/store.ts:108:12
            at http://localhost:9876/_karma_webpack_/webpack:/packages/store/src/compose.ts:22:14
            at Store.webpackJsonp../packages/store/src/store.ts.Store._dispatch packages/store/src/store.ts:117:7)
            at http://localhost:9876/_karma_webpack_/webpack:/packages/store/src/store.ts:34:45
            at Array.map (<anonymous>)
```

deebloo:
update. I think I fixed that error but the test is failing

packages/store/src/state-factory.ts
```diff
@@ -103,15 +108,25 @@ export class StateFactory {
...
+          } else if (result instanceof Promise) {
+            result = fromPromise(result);
+          }
+
+          if (result instanceof Observable) {
+            result = result.pipe(
+              actionMeta.options.cancellable ? takeUntil(actions$.pipe(ofAction(action.type))) : map(r => r)
+            ); // act like a noop
+
             results.push(result);
```
deebloo: it looks like results are only being pushed into the array if it is an observable? I am assuming it should always push into the results array?

packages/store/tests/dispatch.spec.ts
```diff
@@ -172,4 +172,53 @@ describe('Dispatch', () => {
       store.dispatch([new Increment()]);
     })
   );
+
```
deebloo: Below is the failing test. the result should be 1

packages/store/src/symbols.ts
```diff
@@ -62,3 +56,10 @@ export interface StoreOptions<T> {
    */
   children?: any[];
 }
+
+export interface ActionOptions {
+  /**
+   * Cancel the previous uncompleted request(s).
+   */
+  cancellable: boolean;
```
deebloo: what do you think of the "cancelPrevious"? it also avoids to British vs US spelling

PR #212, opened 2018-04-02
fix(store): always set action type

### LINKED ISSUE #200: Action undefined when running minified app code and app breaks
### Versions

```
* ngxs: 2.0.0
* @angular/core: 5.2.9
```

### Repro steps

My component:
```
export class LoginComponent {
	@Select(UserState) user$: Observable<any>

	onLogin(username, password) {
		this.store.dispatch(new Login({ username, password }))
	}
}
```

My action:
```
export class Login {
	constructor(public readonly payload: any) {}
}
```

My state:
```
@State<UserStateModel>({
	name: 'user',
	defaults: {},
})
export class UserState {
	constructor(private loginService: LoginService) {}

	@Action(Login)
	async login({ dispatch }: StateContext<UserStateModel>, { payload }: Login) {
		const { user, error } = await userActions.login(payload, this.loginService)
		if (error) {
			dispatch(new LoginFailure(error))
		} else {
			dispatch(new LoginSuccess(user))
		}
	}
}
```

### Observed behavior

1. LoggerPluginModule logs: `action @@INIT @ 18:47:04.473`
2. When triggering dispatch of Login action my app crashes
3. LoggerPluginModule logs: `action undefined @ 18:47:45.461`
4. Browser logs:
```
handleError TypeError: Cannot read property 'length' of null
    at map.js:2
    at Array.forEach (<anonymous>)
    at t.lazyInit (map.js:2)
    at t.init (map.js:2)
    at t.forEach (map.js:2)
    at t._subscribe (http.js:2106)
    at t._trySubscribe (tslib.es6.js:17)
    at t.subscribe (tslib.es6.js:17)
    at Object.e.subscribeToResult (tslib.es6.js:17)
    at e._innerSub (map.js:2)
    at e._tryNext (map.js:2)
    at e._next (map.js:2)
    at e.next (tslib.es6.js:17)
    at e._subscribe (http.js:2106)
    at e.t._trySubscribe (tslib.es6.js:17)
    at e.t.subscribe (tslib.es6.js:17)
    at t.call (map.js:2)
    at t.subscribe (tslib.es6.js:17)
    at t.call (http.js:2106)
    at t.subscribe (tslib.es6.js:17)
    at t.call (map.js:2)
    at t.subscribe (tslib.es6.js:17)
    at t.call (http.js:2106)
    at t.subscribe (tslib.es6.js:17)
    at t.call (http.js:2106)
    at t.subscribe (tslib.es6.js:17)
    at tslib.es6.js:17
    at new t (zone.js:653)
    at t.toPromise (tslib.es6.js:17)
    at Object.<anonymous> (bootstrap:2)
    at bootstrap:2
    at Object.next (bootstrap:2)
    at bootstrap:2
    at new t (zone.js:653)
    at o (bootstrap:2)
    at Object.requestPerson (bootstrap:2)
    at e.requestPerson (bootstrap:2)
    at u (ngxs-store.js:12)
    at t.invokeActions (ngxs-store.js:12)
    at t._dispatchActions (ngxs-store.js:12)
    at ngxs-store.js:12
    at ngxs-store.js:12
    at ngxs-store.js:12
    at t.handle (http.js:2106)
    at ngxs-store.js:12
    at ngxs-store.js:12
    at t.handle (http.js:2106)
    at ngxs-store.js:12
    at ngxs-store.js:12
    at t.handle (http.js:2106)
    at ngxs-store.js:12
    at t._dispatch (ngxs-store.js:12)
    at t.dispatch (ngxs-store.js:12)
    at e.onLogin (bootstrap:2)
    at Object.eval [as handleEvent] (e.html:12)
    at handleEvent (core.js:13446)
    at Js (core.js:14065)
    at Object.Hs [as handleEvent] (core.js:14065)
    at ei (core.js:9814)
    at core.js:10386
    at HTMLButtonElement.<anonymous> (platform-browser.js:2593)
    at t.invokeTask (zone.js:21)
    at Object.onInvokeTask (core.js:4158)
    at t.invokeTask (zone.js:21)
    at e.runTask (zone.js:21)
    at e.invokeTask [as invoke] (zone.js:21)
    at g (zone.js:1488)
    at HTMLButtonElement.b (zone.js:1488)
```

### Desired behavior

1. LoggerPluginModule logs: `action @@INIT @ 18:49:24.825`
2. LoggerPluginModule logs: `action Login @ 18:50:08.609`

I tested with a minified vendor file and not-minified app file and that works, so it’s only something in the client code of using ngxs that doesn’t work minified.

ngxs_minified

deebloo:
The goal of this PR is to mitigate any minification or name collision issues.

My First thought was to use Symbol but since that isn't supported by any IE browsers id rather not force folks to depend on a polyfill so I am thinking use a small Token class and use that as the token if none is provided. The hope here is to create completely unique actions without depending on the user to add a static property themselves.

Questions:
How would this impact plugins? - solved
Are there performance implications?

Fixes #200

@amcdnl @leon

### DIFF
--- a/packages/devtools-plugin/src/devtools.plugin.ts
+++ b/packages/devtools-plugin/src/devtools.plugin.ts
@@ -39,7 +39,8 @@ export class NgxsReduxDevtoolsPlugin implements NgxsPlugin {
           this.devtoolsExtension.init(state);
         } else {
           const type = getActionTypeFromInstance(action);
-          this.devtoolsExtension.send({ ...action, type }, newState);
+
+          this.devtoolsExtension.send({ ...action, type: type.desc || type }, newState);
         }
       })
     );
--- a/packages/devtools-plugin/src/devtools.plugin.ts
+++ b/packages/devtools-plugin/src/devtools.plugin.ts
@@ -1,5 +1,5 @@
 import { Injectable, Inject } from '@angular/core';
-import { NgxsPlugin, getActionTypeFromInstance, StateStream } from '@ngxs/store';
+import { NgxsPlugin, getActionTypeFromInstance, StateStream, getActionNameFromInstance } from '@ngxs/store';
 import { tap } from 'rxjs/operators';
 
 import { NgxsDevtoolsExtension, NgxsDevtoolsOptions, NGXS_DEVTOOLS_OPTIONS, NgxsDevtoolsAction } from './symbols';
@@ -38,9 +38,9 @@ export class NgxsReduxDevtoolsPlugin implements NgxsPlugin {
         if (isInitAction) {
           this.devtoolsExtension.init(state);
         } else {
-          const type = getActionTypeFromInstance(action);
+          const type = getActionNameFromInstance(action);
 
-          this.devtoolsExtension.send({ ...action, type: type.desc || type }, newState);
+          this.devtoolsExtension.send({ ...action, type }, newState);
         }
       })
     );
--- a/packages/store/src/action.ts
+++ b/packages/store/src/action.ts
@@ -35,14 +35,16 @@ export function Action(actions: any | any[], options?: ActionOptions) {
         action.type = new ActionToken(action.name);
       }
 
-      if (!meta.actions[action.type]) {
-        meta.actions[action.type] = [];
+      const type = action.type;
+
+      if (!meta.actions[type]) {
+        meta.actions[type] = [];
       }
 
-      meta.actions[action.type].push({
+      meta.actions[type].push({
         fn: name,
         options: options || {},
-        type: action.type
+        type
       });
     }
   };
--- a/packages/store/src/action.ts
+++ b/packages/store/src/action.ts
@@ -1,11 +1,18 @@
-import { ensureStoreMetadata, createId } from './internals';
+import { ensureStoreMetadata } from './internals';
 import { ActionOptions } from './symbols';
 
 export class ActionToken {
-  private readonly stamp = createId();
-  private readonly description = `ActionToken ${this.desc} ${this.stamp}`;
+  static counter = 0;
 
-  constructor(public readonly desc: string) {}
+  private readonly stamp: number;
+  private readonly description: string;
+
+  constructor(public readonly desc: string) {
+    ActionToken.counter++;
+
+    this.stamp = ActionToken.counter;
+    this.description = `ActionToken ${this.desc} ${this.stamp}`;
+  }
 
   toString(): string {
     return this.description;
--- a/packages/store/src/action.ts
+++ b/packages/store/src/action.ts
@@ -1,13 +1,14 @@
-import { ensureStoreMetadata, uuid } from './internals';
+import { ensureStoreMetadata, createId } from './internals';
 import { ActionOptions } from './symbols';
 
 export class ActionToken {
-  private readonly stamp = uuid();
+  private readonly stamp = createId();
+  private readonly description = `ActionToken ${this.desc} ${this.stamp}`;
 
   constructor(public readonly desc: string) {}
 
   toString(): string {
-    return `ActionToken ${this.desc} ${this.stamp}`;
+    return this.description;
   }
 }
 
--- a/packages/store/src/action.ts
+++ b/packages/store/src/action.ts
@@ -1,5 +1,15 @@
-import { ensureStoreMetadata, getTypeFromKlass } from './internals';
+import { ensureStoreMetadata, uuid } from './internals';
 import { ActionOptions } from './symbols';
+
+export class ActionToken {
+  private readonly stamp = uuid();
+
+  constructor(public readonly desc: string) {}
+
+  toString(): string {
+    return `ActionToken ${this.desc} ${this.stamp}`;
+  }
+}
 
 /**
  * Decorates a method with a action information.
@@ -13,15 +23,20 @@ export function Action(actions: any | any[], options?: ActionOptions) {
     }
 
     for (const action of actions) {
-      const type = getTypeFromKlass(action);
-      if (!meta.actions[type]) {
-        meta.actions[type] = [];
+      if (typeof action.type === 'string') {
+        action.type = new ActionToken(action.type);
+      } else if (!action.type) {
+        action.type = new ActionToken(action.name);
       }
 
-      meta.actions[type].push({
+      if (!meta.actions[action.type]) {
+        meta.actions[action.type] = [];
+      }
+
+      meta.actions[action.type].push({
         fn: name,
         options: options || {},
-        type
+        type: action.type
       });
     }
   };
--- a/packages/store/src/action.ts
+++ b/packages/store/src/action.ts
@@ -23,9 +23,7 @@ export function Action(actions: any | any[], options?: ActionOptions) {
     }
 
     for (const action of actions) {
-      if (typeof action.type === 'string') {
-        action.type = new ActionToken(action.type);
-      } else if (!action.type) {
+      if (!action.type) {
         action.type = new ActionToken(action.name);
       }
 
--- a/packages/store/src/internals.ts
+++ b/packages/store/src/internals.ts
@@ -9,14 +9,6 @@ export interface MetaDataModel {
   instance: any;
 }
 
-let idCounter = 0;
-
-export function createId(): string {
-  idCounter = idCounter + 1;
-
-  return idCounter.toString();
-}
-
 /**
  * Ensures metadata is attached to the klass and returns it.
  */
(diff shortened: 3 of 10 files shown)


packages/store/src/action.ts
```diff
@@ -1,6 +1,16 @@
...
 import { ActionOptions } from './symbols';
 
+export class ActionToken {
+  private readonly stamp = uuid();
+
+  constructor(public readonly desc: string) {}
+
+  toString(): string {
+    return `ActionToken ${this.desc} ${this.stamp}`;
+  }
```
deebloo: required in order for the ActionToken to be an object property

packages/store/src/action.ts
```diff
@@ -13,15 +23,18 @@ export function Action(actions: any | any[], options?: ActionOptions) {
     }
 
     for (const action of actions) {
-      const type = getTypeFromKlass(action);
-      if (!meta.actions[type]) {
-        meta.actions[type] = [];
+      if (!action.type) {
+        action.type = new ActionToken(action.name);
+      }
```
deebloo: if not action type is defined define one as an ActionToken

packages/store/src/internals.ts
```diff
@@ -8,6 +8,18 @@ export interface MetaDataModel {
   children: any[];
 }
 
+/* FROM: https://gist.github.com/jed/982883 */
+/* tslint:disable  */
+export function uuid(a?, b?) {
```
amcdnl: Why do we need this? Couldn't we just have a count increment?

```
let count = 0;
getId() { return (++count).toString() };
```

packages/store/src/action.ts
```diff
@@ -13,15 +23,18 @@ export function Action(actions: any | any[], options?: ActionOptions) {
...
+
+      if (!meta.actions[action.type]) {
```
amcdnl: How does this `action.type` not represent a object here?

packages/store/src/action.ts
```diff
@@ -1,6 +1,16 @@
...
+import { ensureStoreMetadata, uuid } from './internals';
 import { ActionOptions } from './symbols';
 
+export class ActionToken {
+  private readonly stamp = uuid();
+
+  constructor(public readonly desc: string) {}
+
+  toString(): string {
+    return `ActionToken ${this.desc} ${this.stamp}`;
```
amcdnl: Wouldn't it be faster to cache this?

packages/devtools-plugin/src/devtools.plugin.ts
```diff
@@ -39,7 +39,8 @@ export class NgxsReduxDevtoolsPlugin implements NgxsPlugin {
           this.devtoolsExtension.init(state);
         } else {
           const type = getActionTypeFromInstance(action);
-          this.devtoolsExtension.send({ ...action, type }, newState);
+
+          this.devtoolsExtension.send({ ...action, type: type.desc || type }, newState);
```
amcdnl: Can we make a helper method that will just get this for us?

packages/store/src/internals.ts
deebloo: yeah I think that would work. would we want one generic helper like this?

packages/store/src/action.ts
deebloo: it does represent an object. but when you do use an object as a key it calls the toString method and uses that as the actual key

deebloo: it works the same way as angular InjectionToken

packages/store/src/action.ts
deebloo: we sure could

packages/devtools-plugin/src/devtools.plugin.ts
deebloo: maybe getActionTypeFromInstance could return either or a string or and ActionToken.

packages/store/src/action.ts
amcdnl: Oh didn't realize it did that. I figured it would just stick the object there.

packages/store/src/internals.ts
amcdnl: Sure, a count tracker maybe?

amcdnl: Not sure why we would want to use it anywhere else but sure?

packages/devtools-plugin/src/devtools.plugin.ts
amcdnl: IDK, wouldn't u still want the raw type sometimes? Making that return that could be undesirable at some times.

packages/store/src/action.ts
deebloo: yeah I learned that as well.

deebloo:
@amcdnl updates made

packages/store/src/action.ts
deebloo: if you take the toString method off it breaks

packages/devtools-plugin/src/devtools.plugin.ts
deebloo: I added a method that returns just the name and another that returns the string or the ActionToken

packages/store/src/state-factory.ts
```diff
@@ -101,28 +101,29 @@ export class StateFactory {
   invokeActions(getState, setState, dispatch, actions$, action) {
     const results = [];
 
-    for (const metadata of this.states) {
+    for (const data of this.states) {
+      const metadata: MetaDataModel = data;
```
amcdnl: Why cast it like this? Why not set the states type and let it infer?

packages/store/src/state-factory.ts
```diff
@@ -101,28 +101,29 @@ export class StateFactory {
...
+
+      const actionMetas = metadata.actions[name.toString()];
 
       if (actionMetas) {
         for (const actionMeta of actionMetas) {
           const stateContext = this.createStateContext(getState, setState, dispatch, metadata);
           let result = metadata.instance[actionMeta.fn](stateContext, action);
 
-          if (result === undefined) {
+          if (!result) {
```
amcdnl: What if I return false or 0 here? Wouldn’t that break this?

packages/store/src/state-factory.ts
```diff
@@ -101,28 +101,29 @@ export class StateFactory {
...
-          if (result instanceof Observable) {
-            result = result.pipe(
-              (<ActionOptions>actionMeta.options).takeLast
-                ? takeUntil(actions$.pipe(ofAction(action.constructor)))
-                : map(r => r)
-            ); // act like a noop
-          }
+          result = result.pipe(
+            (<ActionOptions>actionMeta.options).takeLast
+              ? takeUntil(actions$.pipe(ofAction(action.constructor)))
```
amcdnl: Potential issue: if action is not a class, which happens when dev tools emit actions.

packages/store/src/state-factory.ts
```diff
@@ -101,28 +101,29 @@ export class StateFactory {
...
+              : map(r => r)
```
amcdnl: Curious if you have a better idea for a noop?

packages/store/src/utils.ts
```diff
@@ -1,17 +1,30 @@
...
+
+    return action.constructor.type as string;
+  } else if (action.type) {
     // events from dev tools are plain objects
-    return event.type;
+    return action.type;
   }
 }
 
+export function getActionNameFromInstance(action: any): string {
```
amcdnl: Add a comment to this please.

amcdnl: To the function that is

packages/store/src/action.ts
```diff
@@ -13,15 +31,18 @@ export function Action(actions: any | any[], options?: ActionOptions) {
...
     for (const action of actions) {
-      const type = getTypeFromKlass(action);
-      if (!meta.actions[type]) {
-        meta.actions[type] = [];
+      if (!action.type) {
+        action.type = new ActionToken(action.name);
+      }
+
+      if (!meta.actions[action.type]) {
+        meta.actions[action.type] = [];
```
amcdnl: You do action.type a lot, let’s lazy it.

packages/store/src/action.ts
```diff
@@ -1,6 +1,24 @@
...
+  static counter = 0;
+
+  private readonly stamp: number;
+  private readonly description: string;
+
+  constructor(public readonly desc: string) {
+    ActionToken.counter++;
+
+    this.stamp = ActionToken.counter;
+    this.description = `ActionToken ${this.desc} ${this.stamp}`;
```
amcdnl: How is this readonly and being set here. Odd.

packages/store/src/state-factory.ts
deebloo: it would just be undefined then. I don't think it will break

packages/store/src/state-factory.ts
deebloo: so that would be if you just returned a raw value from the state method?

PR #215, opened 2018-04-03
added form plugin unit and integration tests. fixed typos.

tsawan:


packages/form-plugin/src/form.plugin.ts
```diff
@@ -10,39 +10,39 @@ export class NgxsFormPlugin implements NgxsPlugin {
 
     let nextState = state;
 
-    if (type === 'UpdateFormValue' || type === 'UpdateForm') {
+    if (type === '[Forms] Update Form Value' || type === '[Forms] Update Form') {
```
amcdnl: Lets use the class type here vs strings.

PR #227, opened 2018-04-05
feat(actions): completions stream

amcdnl:
Adds new action completions stream


docs/SUMMARY.md
```diff
@@ -13,7 +13,7 @@
   * [Sub States](advanced/sub-states.md)
   * [Cancelation](advanced/cancelation.md)
   * [Caching](advanced/cache.md)
-  * [Pub Sub](advanced/pubsub.md)
+  * [Action Handlers](advanced/action-handlers.md)
```
deebloo: are we removing any docs on the action stream?

packages/store/src/actions-completion-stream.ts
```diff
@@ -0,0 +1,27 @@
...
+ * You can listen to this in services to react without stores.
+ */
+@Injectable()
+export class ActionCompletions extends Observable<any> {
+  private _stream: InternalActionCompletions;
+
+  constructor() {
+    super();
+    this._stream = new InternalActionCompletions();
+  }
```
deebloo: I don't think this will work. it isn't using the internal stream

packages/store/src/actions-completion-stream.ts
```diff
@@ -0,0 +1,27 @@
...
+
+  subscribe(fn?) {
+    return this._stream.subscribe(fn);
+  }
```
deebloo: if we are extending Observable we don't need this.

deebloo:
@amcdnl I made some fixes. take a look when you get a chance and let me know

docs/SUMMARY.md
amcdnl: No, just name update.

PR #232, opened 2018-04-06
fixed send error when socket is undefined

tsawan:
hold on to this till I submit another PR with unit tests.


amcdnl:
Thanks for your contribution!

PR #235, opened 2018-04-06
Fix dispatch observable issues (fixes #233)

### LINKED ISSUE #233: Dispatch observable not firing in particular scenarios
### Versions

```
* ngxs: 3.0.0-beta.0
* @angular/core: 5.0.0
```

### Repro steps

I have picked up some issues with the dispatch method observable. 
The observer does not ever get hit in these situations:
- when the action handler returns something other than a void/promise/observable
- when there aren't any action handlers for the Action

Do you agree that the observable should still publish completion in both of these scenarios?

markwhitfeld:
Added primary tests to test the observable returned by the dispatch method.
Reproduced both reported issues with tests.
Fixed the issues.


leon:
LGTM

amcdnl:
AWESOME WORK!

PR #236, opened 2018-04-06
added unit test for websocket plugin.

tsawan:


amcdnl:
@tsawan - Test failed :(

Issue #248, opened 2018-04-08
Bad RxJS imports creating a large bundle size


codediodeio:
Importing Rx classes with `import { something } from 'rxjs'` is causing the bundle size to explode. A picture is worth a thousand words. Sending over a PR with a fix shortly. 

![ngxs-import](https://user-images.githubusercontent.com/10172199/38462567-3f8a9da8-3a9e-11e8-958e-dec3e85ed41d.png)

amcdnl:
I think I did this in prep for rx6, will have to undo these when that happens

deebloo:
Yes when we upgrade we can change but for now it will cause enourmous bundles. Maybe we should create a ticket to track rxjs6 upgrade

PR #249, opened 2018-04-08
fix: removed rxjs imports that bring in full library

codediodeio:
Addressing #248

- Fixed bad rxjs imports
- Blacklisted this import style in tsconfig to prevent this in the future.


amcdnl:
This is going to be reverted back in rx6 release. Lol

PR #253, opened 2018-04-09
NG6/RX6 Upgrade

amcdnl:
DO NOT MERGE, WIP


amcdnl:
Not sure whats up w/ the websocket stuff.

package.json
```diff
@@ -94,9 +93,11 @@
     "karma-jasmine": "~1.1.0",
     "karma-jasmine-html-reporter": "^0.2.2",
     "lint-staged": "^7.0.0",
+    "mock-socket": "^7.1.0",
     "ng-packagr": "^2.3.0",
     "prettier": "^1.8.2",
-    "rxjs": "^5.5.2",
+    "rxjs": "^6.0.0-rc.1",
+    "rxjs-compat": "^6.0.0-beta.1",
```
deebloo: do we need this if we are updating to the new syntax?

amcdnl: I had to add this for websocket, not sure whats up w/ that.

yarrgh: From what I understand, and maybe I'm wrong, only applications should use the rxjs-compat library. Libraries should use rxjs 6.0.0 and then if the apps need the backwards compatibility, then they import rxjs-compat

PR #255, opened 2018-04-10
 feat(store): create action stream that shows the action lifecycle

deebloo:


markwhitfeld:
I assume that the `@Action` handlers will only respond to dispatched actions. We may need to look for a syntax for response to completion in future. Something like:
```typescript 
@Action(Completed(FeedAnimals))
```
or
```typescript 
@Action(onCompletion(FeedAnimals))
```
or
```typescript 
@ActionCompleted(FeedAnimals)
```

deebloo:
Hmmmm. Not sure if we wanna do that or not. Could potentially also be another config option.

deebloo:
Whatever we decide I think it will have to wait till after this.

docs/advanced/action-handlers.md
```diff
@@ -1,19 +1,28 @@
...
 The action handler is a observable that recieves all the actions dispatched before the state takes any action on it.
-Since its an observable, we can use pipes and we created a `ofAction` pipe to make filtering the actions easier.
+
+Actions in NGXS also have a lifecycle. Since any potential action can be async we tag actions showing whether they are "DISPATCHED" or "COMPLETED". This gives you the ability to react to actions at different points in their existence.
+
+Since its an observable, we can use pipes and created 3 of them.
+
+* `ofAction`: triggers both when an action has been dispatched and when it completes
+* `ofActionDispatched`: triggers when an action has been dispatched but NOT when it completes
+* `ofActionCompleted`: triggers when an action has been completed but NOT when it is dispatched
```
amcdnl: We might want a `ofActionErrored` too. Just a thought.

packages/store/src/state-factory.ts
```diff
@@ -117,7 +117,7 @@ export class StateFactory {
           if (result instanceof Observable) {
             result = result.pipe(
               (<ActionOptions>actionMeta.options).cancelUncompleted
-                ? takeUntil(actions$.pipe(ofAction(action.constructor)))
+                ? takeUntil(actions$.pipe(ofActionDispatched(action.constructor)))
```
amcdnl: is `action` here the class or instance? It looks like instance and if so, this will cause issues if the action dispatched is not a class (think devtools or websocket). We probably need to change the `action.constructor` to check and return.

packages/store/src/state-stream.ts
```diff
@@ -1,20 +1,12 @@
...
  * BehaviorSubject of the entire state.
  */
 @Injectable()
 export class StateStream extends BehaviorSubject<any> {
-  constructor(
-    @Optional()
-    @SkipSelf()
-    parent: StateStream
-  ) {
+  constructor() {
```
amcdnl: If we remove all that, do we event need the constructor now?

packages/store/src/store.ts
```diff
@@ -126,7 +127,14 @@ export class Store {
             this._actions,
             action
           )
-          .pipe(map(() => this._stateStream.getValue()));
+          .pipe(
+            tap(() => {
+              this._actions.next({ action, status: ActionStatus.Completed });
```
amcdnl: What do you think about making these class and just say: `new CompletedAction(action)`?

packages/store/src/of-action.ts
```diff
@@ -9,12 +13,63 @@ export function ofAction<T>(...allowedTypes);
...
+  return function(o: Observable<any>) {
+    return o.pipe(filterStatus(allowedMap, ActionStatus.Completed), mapAction());
+  };
+}
+
+function filterStatus(allowedTypes: { [key: string]: boolean }, status?: ActionStatus) {
+  return filter((ctx: ActionContext) => {
+    const actionType = getActionTypeFromInstance(ctx.action);
+
+    if (status) {
```
amcdnl: I'd prefer:

```
const type =  allowedTypes[actionType] ;
return status ? type && ctx.status === status || type;
```

docs/advanced/action-handlers.md
```diff
@@ -1,19 +1,28 @@
...
+
```
amcdnl: Need to add this to the changelog too.

deebloo: yes

packages/store/src/store.ts
deebloo: we could do that

packages/store/src/state-stream.ts
deebloo: I think something breaks if there is no immediate value

packages/store/src/state-factory.ts
deebloo: it is counting on an instance yes. we can make this more robust but we should also be safer in the operators as well

docs/advanced/action-handlers.md
deebloo: Ill look into it

amcdnl:
@markwhitfeld listening via decorators would be super cool but not possible due to AoT and needing to unbind on destroy. see this for more info: https://medium.com/thecodecampus-knowledge/the-easiest-way-to-unsubscribe-from-observables-in-angular-5abde80a5ae3

packages/store/src/state-factory.ts
deebloo: actually I should probably put all of that logic in the operators anyway

deebloo:
@amcdnl I don't think AOT has anything to do with this in particular. I think we could investigate but would def be a separate issue

PR #258, opened 2018-04-10
chore(docs): concepts/actions, rename file and fix typos

maartentibau:
Since in the whole docs, action it actually always mentioned in its plural form, I think it would be better to change the title and file name as well.


leon:
LGTM

PR #262, opened 2018-04-11
Video addendum via Angular Firebase a la YouTube.

Campos:
Added project video from Angular Firebase YouTube Channel.


amcdnl:
This is in the media section already.

Issue #265, opened 2018-04-11
@ngxs/form-plugin disabled controls data is removed from model


AlgoTrader:
Disabled controls are removed from model. After re-enabling controls all the data is lost

```
getRawValue(): any
The aggregate value of the FormGroup, including any disabled controls.
If you'd like to include all values regardless of disabled status, use this method. Otherwise, the value property is the best way to get the value of the group.
```

it seems ngxs use 'value' for FormGroup (and its friends).  We should use 'getRawValue()' instead of 'value'

This is one more killing model properties bug. Last time my own properties were killed, now disabled properties are killed. Can you please still consider merging value changes into model, not replace the whole model?

tsawan:
@amcdnl I will look into this.

AlgoTrader:
Exellent! Thanks a lot, wait for a new rc

arturovt:
@AlgoTrader 

Sorry for the very late answer, but I'm currently working on the form plugin issues/features. Could you help me by providing a stackblitz example if the issue still exists?

arturovt:
As I see it was fixed by [this PR](https://github.com/ngxs/store/pull/278).

PR #278, opened 2018-04-12
using formGroup.getRawValue() so disabled control's data is not thrown away. 

tsawan:
Fixed issue #265


deebloo:
I see that you updated the integration tests. should there be any unit tests updated as well?

tsawan:
@deebloo we don't need to add or update any unit test for these changes.

amcdnl:
LGTM

PR #283, opened 2018-04-13
feat(store): add life cycle interface and test

### LINKED ISSUE #272: Interfaces for Livecycle Events
I think it would be good to create a simple Interface that needs to be implemented on the state class for the livecycle events.

Implementing the interface would be more intuitive than guessing the signature of the livecycle method.

`implements NgxsOnInit`

deebloo:
closes #272

### DIFF
--- a/packages/store/src/symbols.ts
+++ b/packages/store/src/symbols.ts
@@ -67,3 +67,5 @@ export interface ActionOptions {
 export interface NgxsOnInit {
   ngxsOnInit(ctx?: StateContext<any>): void | any;
 }
+
+export type NgxsLifeCycle = Partial<NgxsOnInit>;
--- a/packages/store/src/symbols.ts
+++ b/packages/store/src/symbols.ts
@@ -63,3 +63,7 @@ export interface ActionOptions {
    */
   cancelUncompleted: boolean;
 }
+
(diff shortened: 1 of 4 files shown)


packages/store/src/symbols.ts
```diff
@@ -63,3 +63,9 @@ export interface ActionOptions {
    */
   cancelUncompleted: boolean;
 }
+
+export interface NgxsOnInit {
+  ngxsOnInit(ctx?: StateContext<any>): void | any;
+}
+
+export type NgxsLifeCycle = Partial<NgxsOnInit>;
```
deebloo: this makes sure that we know internally that lifecycle methods are optional. we can also add additional life cycle interfaces like this:

```TS
export type NgxsLifeCycle = Partial<NgxsOnInit, NgxsOnDestroy>;
```

amcdnl: when would destroy ever happen?

deebloo: oh i'm not saying it would. I am just showing an example of how we can compose the lifecycle in a typesafe way

yarrgh:
What about having a generic interface as well? That way the StateContext can be correctly typed.

PR #287, opened 2018-04-13
fix(router): fix imports

amcdnl:
Fixes imports


amcdnl:
Tracking build issue here: https://github.com/dherges/ng-packagr/issues/767

PR #289, opened 2018-04-14
Improve core testing and type safety

markwhitfeld:


packages/store/tests/state.spec.ts
```diff
@@ -1,5 +1,6 @@
 import { TestBed } from '@angular/core/testing';
 
+import { InitState, UpdateState } from '@ngxs/store';
```
deebloo: if we want to import for @ngxs/store we should move all of the imports to there. avoids confusion

packages/store/tests/state.spec.ts
```diff
@@ -10,7 +11,7 @@ describe('Store', () => {
     @State({
       name: 'moo'
     })
-    class BarState {}
+    class BarState { }
```
deebloo: are you using the exact same prettier config that is set in the project? for example in my project it actually removes the space between those brackets.

packages/store/tests/state.spec.ts
```diff
@@ -84,4 +85,107 @@ describe('Store', () => {
       expect(res).toBe(true);
     });
   });
+
+  describe('given the ngxsOnInit lifecycle method is present', () => {
+    it('should call the ngxsOnInit method on root module initialisation', () => {
```
deebloo: can remove this test. there is already a test for ngxsOnInit in state.spec

packages/store/tests/state.spec.ts
```diff
@@ -84,4 +85,107 @@ describe('Store', () => {
...
+      TestBed.configureTestingModule({
+        imports: [NgxsModule.forRoot([FooState])]
+      });
+
+      TestBed.get(FooState);
+
+      expect(listener).toEqual(['onInit']);
+    });
+
+    it('should call the ngxsOnInit method on feature module initialisation', () => {
```
deebloo: can you move these tests to state.spec?

deebloo: Only if it makes sense. Thoughts?

packages/store/src/action.ts
```diff
@@ -13,7 +13,7 @@ export function Action(actions: any | any[], options?: ActionOptions) {
     }
 
     for (const action of actions) {
-      const type = action.type;
+      const type: string = action.type;
```
amcdnl: Do we really need this? Can't it imply that?

packages/store/src/internals.ts
```diff
@@ -1,11 +1,32 @@
...
+export type DispatchFn = (actions: any | any[]) => Observable<any>;
 
 export interface MetaDataModel {
   name: string;
-  actions: any;
+  actions: ObjectKeyMap<ActionHandlerMetaData[]>;
   defaults: any;
   path: string;
-  children: any[];
+  children: StateKlass[];
```
amcdnl: Since its just just `Class`, you can problably make this `StateClass`

packages/store/src/internals.ts
```diff
@@ -67,9 +88,9 @@ export function fastPropGetter(paths: string[]): (x: any) => any {
  *  };
  *
  */
-export function buildGraph(states) {
-  const findName = klass => {
-    const meta = states.find(g => g === klass);
+export function buildGraph(stateKlasses: StateKlass[]): StateKeyGraph {
+  const findName = (klass: StateKlass): string => {
+    const meta: StateKlass = stateKlasses.find(g => g === klass);
```
amcdnl: Won't it imply this?

packages/store/src/internals.ts
```diff
@@ -132,11 +159,11 @@ export function nameToState(states) {
...
  *
  */
-export function findFullParentPath(obj: any, newObj: any = {}) {
-  const visit = (child: any, keyToFind: string) => {
+export function findFullParentPath(obj: StateKeyGraph, newObj: ObjectKeyMap<string> = {}): ObjectKeyMap<string> {
+  const visit = (child: StateKeyGraph, keyToFind: string): string => {
     for (const key in child) {
       if (child.hasOwnProperty(key) && child[key].indexOf(keyToFind) >= 0) {
-        const parent = visit(child, key);
+        const parent: string = visit(child, key);
```
amcdnl: Won't it imply `string` here?

packages/store/src/internals.ts
```diff
@@ -145,7 +172,7 @@ export function findFullParentPath(obj: any, newObj: any = {}) {
 
   for (const key in obj) {
     if (obj.hasOwnProperty(key)) {
-      const parent = visit(obj, key);
+      const parent: string = visit(obj, key);
```
amcdnl: Same re: implied state.

packages/store/src/symbols.ts
```diff
@@ -61,7 +61,7 @@ export interface ActionOptions {
   /**
    * Cancel the previous uncompleted observable(s).
    */
-  cancelUncompleted: boolean;
+  cancelUncompleted?: boolean;
```
amcdnl: Good catch!

packages/store/src/state-factory.ts
```diff
@@ -78,15 +103,24 @@ export class StateFactory {
     return mappedStores;
   }
 
-  addAndReturnDefaults(stateKlasses: any[]): { defaults: any; states: any[] } {
+  addAndReturnDefaults(stateKlasses: any[]): { defaults: any; states: MappedStore[] } {
     if (stateKlasses) {
       const states = this.add(stateKlasses);
-      const defaults = states.reduce((result, meta) => setValue(result, meta.depth, meta.defaults), {});
+      const defaults = states.reduce((result, meta: MappedStore) => setValue(result, meta.depth, meta.defaults), {});
```
amcdnl: If we are hardcore typing, we should make `result: any` too.

packages/store/src/state-factory.ts
```diff
@@ -64,7 +89,7 @@ export class StateFactory {
 
       const instance = this._injector.get(klass);
 
-      mappedStores.push({
+      mappedStores.push(<MappedStore>{
```
amcdnl: Wouldn't the type be implied based on the array?

packages/store/src/state-factory.ts
```diff
@@ -2,18 +2,40 @@ import { Injector, Injectable, SkipSelf, Optional } from '@angular/core';
...
+  GetStateFn,
+  SetStateFn,
+  DispatchFn
+} from './internals';
 import { getActionTypeFromInstance, setValue, getValue } from './utils';
 import { ofActionDispatched } from './of-action';
+import { Actions } from '@ngxs/store';
+import { InternalActions } from '@ngxs/store/src/actions-stream';
+
+interface MappedStore {
```
amcdnl: Typically I like to put these in a seperate file, i think symbols is what we been using.

packages/store/src/state-factory.ts
```diff
@@ -2,18 +2,40 @@ import { Injector, Injectable, SkipSelf, Optional } from '@angular/core';
...
+  ObjectKeyMap,
+  StateKlass,
+  GetStateFn,
+  SetStateFn,
+  DispatchFn
+} from './internals';
 import { getActionTypeFromInstance, setValue, getValue } from './utils';
 import { ofActionDispatched } from './of-action';
+import { Actions } from '@ngxs/store';
+import { InternalActions } from '@ngxs/store/src/actions-stream';
```
amcdnl: This import looks wrong.

packages/store/src/state-factory.ts
```diff
@@ -2,18 +2,40 @@ import { Injector, Injectable, SkipSelf, Optional } from '@angular/core';
...
+  ActionHandlerMetaData,
+  ObjectKeyMap,
+  StateKlass,
+  GetStateFn,
+  SetStateFn,
+  DispatchFn
+} from './internals';
 import { getActionTypeFromInstance, setValue, getValue } from './utils';
 import { ofActionDispatched } from './of-action';
+import { Actions } from '@ngxs/store';
```
amcdnl: This import looks wrong.

packages/store/src/state-factory.ts
deebloo: Both of those should import directly not use the alias

leon: I think this is a "feature" of autoimports in vscode and intellij.
the editor seems to sometimes want to do a absolute import.

Just do a last check before committing, making sure that there are no imports to /src/.

Maybe we can add a tslint rule that forbids importing from `@ngxs/store/src/*` ?

deebloo: that might not be a bad idea

deebloo: but even then we shouldn't import from @ngxs/store when we are inside of the store

packages/store/tests/state.spec.ts
markwhitfeld: Aren't they already in state.spec?

packages/store/tests/state.spec.ts
markwhitfeld: Yes, I saw that when I re-based. I would like to keep this test because it corresponds to other parts of the test describe section to show completeness. I will look at your test and see which to include.

packages/store/tests/state.spec.ts
markwhitfeld: I haven't made any choice regarding prettier. I was relying on the default setup. Is there anything specific I need to do? I think it runs as part of the commit.

packages/store/tests/state.spec.ts
markwhitfeld: Ah, those were auto imports. Will move to reference the local files. like the other imports

packages/store/src/action.ts
markwhitfeld: Yes, I think it became inferrable once I improved the major typings. Will review.

packages/store/src/internals.ts
markwhitfeld: Yipee! I was trying to keep with the naming of the Klass I found in the variable names. Would much prefer the name `StateClass`.

packages/store/src/internals.ts
markwhitfeld: Yes, I will move to implied where possible

packages/store/src/state-factory.ts
markwhitfeld: The explicit typing ensures that any deviation from the interface gets caught by the typing system. I'll double check, but I think that it is necessary.

packages/store/tests/state.spec.ts
deebloo: It should yes maybe something is different in my local setup then

packages/store/tests/state.spec.ts
deebloo: Ok. If this covers the same same remove the other so we don't just have a dupe test

packages/store/src/state-factory.ts
markwhitfeld: I kept this local to the file because it is not needed elsewhere and therefore not exported. Let me know what way you want me to go with this... keep it here or move it to another file `Symbols.ts` or `internal.ts`.

packages/store/tests/state.spec.ts
deebloo: Ah yes sorry I thought they were in dispatch

packages/store/src/state-factory.ts
markwhitfeld: Agreed

packages/store/src/state-factory.ts
deebloo: I'd say they are fine here. But Austin can make the final call

packages/store/src/state-factory.ts
deebloo: If mappedStore is type correctly I don't think you need it

markwhitfeld:
Hmmm... build failure seems unrelated to my changes... looks like a yarn failure. Will see when I push again.

markwhitfeld:
Changes made and build passing. Please review again @amcdnl @deebloo 
Thanks!

packages/store/src/state-factory.ts
amcdnl: It’s fine

PR #296, opened 2018-04-15
chore(docs): add info about using form-plugin in submodule

MartinNuc:
The `ngxsForm` used by form-plugin is a directive therefore must be imported in every module. This PR adds this information to the documentation.


amcdnl:
Thanks!

PR #299, opened 2018-04-16
refactor: update tests to use internal state management

deebloo:
This updates lifecycle tests to use the available state management mechanism. This ensures that state is also set in the correct order.


leon:
LGTM

deebloo:
@markwhitfeld that makes sense but I feel this is pretty safe to do here and demonstrates a more standard workflow so it verifies something a user is most likely to actually do.

amcdnl:
@deebloo - is this ready for merge?

PR #306, opened 2018-04-17
chore: update ng-packagr and rxjs

deebloo:
A new RC of ng-packagr was just released and conforms to the new Angular Package Format.
Also update rxjs to the latest rc and updated some of our code to make sure it is using the new rxjs style


deebloo:
@leon looks like that issue with yarn.lock again? how to you fix this issue?

amcdnl:
@deebloo - just run `yarn install` and it will fix it.

Issue #307, opened 2018-04-17
@ngxs/form-plugin model and status are empty until first field edit


AlgoTrader:
It is extremely unpleasant thing to have model and status undefined until first edit of any form field.

Angular form send valueChanges/statusChanges when something changed. For a newly initialized form, no valueChanges/statusChanges values sent yet, so all the model in store is completly uninitialized

It can be easyly fixed, in directive initialization there should be value/status form props read and used for state initialization

leon:
@AlgoTrader As you say it shouldn't be hard to fix.
Do you think you could have a go at implementing it in a pull request?

AlgoTrader:
The first state update cause two-way form and state synching. First, all data from state (if it exists) are transferred to form. The second, form model, form state and form dirty are transferred back to state.

We cannot do it in reverse order, then state data can be lost. We cannot make it one-way also.

AlgoTrader:
I fixed the integration test

tsawan:
not sure I am clear on this.
I am using this in my app. with form model initialized to an empty object. This is if the user opens the form in 'new' mode. If the user is editing an existing customer object, I initialize the state through a populate action and the form is shown in 'edit' mode.

tsawan:
sorry, just noticed it's already merged in.

LosD:
Shouldn't this have been closed with #309's merge?

(hint: If you use write e.g. `fixes #309` in the PR description or one of its commits, Github will auto-close the issue on PR merge. Also works for [other keywords](https://help.github.com/articles/closing-issues-using-keywords/))

arturovt:
@AlgoTrader could you close the issue?

arturovt:
As I understand the #309 PR is related to this issue. This issue should have been closed.

PR #309, opened 2018-04-18
fix: on fisrt state update, sync form model, status, dirty to state #307

AlgoTrader:
The fix is a little bit tricky. There are two situations

1) data from state model are transferred to form model
2) missing in state model fields are imported from form model

If state model is {first: '123'} and form is {first: '', second:''}, then we should first transfer state to form, then we need to do the reverse operation. As a result, state will be {first: '123', second: ''}.

### DIFF
--- a/packages/form-plugin/src/directive.ts
+++ b/packages/form-plugin/src/directive.ts
@@ -2,7 +2,7 @@ import { Directive, Input, OnInit, OnDestroy, ChangeDetectorRef } from '@angular
 import { FormGroupDirective } from '@angular/forms';
 import { Store, getValue } from '@ngxs/store';
 import { Subject } from 'rxjs';
-import { takeUntil, debounceTime } from 'rxjs/operators';
+import { takeUntil, debounceTime, first } from 'rxjs/operators';
 import { UpdateFormStatus, UpdateFormValue, UpdateFormDirty, UpdateFormErrors, UpdateForm } from './actions';
 
 @Directive({ selector: '[ngxsForm]' })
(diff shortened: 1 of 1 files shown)


packages/form-plugin/src/directive.ts
```diff
@@ -27,6 +27,27 @@ export class FormDirective implements OnInit, OnDestroy {
         }
       });
 
+    // On first state change, sync form model, status and dirty with state
+    this._store
+      .select(state => getValue(state, `${this.path}`))
+      .pipe(takeUntil(this._destroy$), first())
```
deebloo: if we use first() here we don't need takeUntil since it will cleanup after the first item it gets. you could also use _store.selectOnce instead.

deebloo: I don't think fire on the first state change since it will always return a value immediately.

AlgoTrader: You are right. In theory,  this._destroy$ may emit value before first value in store observable. 
It is safe to have a protection against possible value races.

deebloo: fair

AlgoTrader: There is a problem. If we do it in sync manner, then UpdateFormValue will kill all the current state. 

I don't like how it works, but first we push state to form, then get a form value and push it back to state.

The reason is state model can be null or undefined, it can just have part of fields or it may has all the values there.

deebloo:
looks like one of the integration tests is failing

AlgoTrader:
Integration test expects that the first dispatched value in the state will be the correct one. This is no longer correct. There are two UpdateFormValue dispatched, one is when form is initialized and the second is when you do `component.pizzaForm.patchValue({ toppings: 'oli' });`

The problem is  `component.pizza$.pipe(take(1))`

take(1) receives an empty initializing value. The easy way to fix the test is to skip(1).take(1). I just wonder to know if there are better ways? Any ideas?

AlgoTrader:
I finally make the test working, but there are several problems with current tests:

1) Test cases are not atomic. For example a single testcase makes two form updates and six asserts:
    `component.pizzaForm.patchValue({ toppings: 'oli' });`
    `component.pizzaForm.patchValue({ toppings: 'olives', crust: 'thick' });`

2) Test cases rely on observables to fire subscribe callback synchronously. The slightest implementation change case Observable callbacks to fire asyncronously. 

3) Test cases rely on index of observable value. For example `component.pizza$.pipe(take(1))`

I had to implement tests in async way to make them working. using done() jasmine function in observable callbacks

PR #321, opened 2018-04-21
feat: entity base class

leon:
A base class that you can implement to get methods on your state class to simplify handling crud operations.

It looks something like this.

```TS
class Project {
  id: string;
  name: string;
}

class AddOne {
  static readonly type = '[Project] AddProject';
  constructor(public project: Project) {}
}

class AddMany {
  static readonly type = '[Project] AddProjects';
  constructor(public projects: Project[]) {}
}

interface ProjectStateModel extends EntityState<Project> {}

@State<ProjectStateModel>({
  name: 'projects',
  defaults: {
    ...EntityBase.defaults
  }
})
class ProjectState extends EntityBase<Project, ProjectStateModel> {
  // pass injector to base class so that we can get and set state
  // @todo, can we work around this?
  constructor(injector: Injector) {
    super(injector);
  }

  // will automatically do a ctx.setState if needed
  @Action(AddOne)
  addProject(ctx: StateContext<ProjectStateModel>, action: AddOne) {
    this.addOne(action.project);
  }

  @Action(AddMany)
  addProjects(ctx: StateContext<ProjectStateModel>, action: AddMany) {
    this.addMany(action.projects);
  }
}
```

for now I've only implemented these two, but it shouldn't take me too long to implement the rest of the methods.

Though I would like some input on the api of it.

I would like to be able to get at the `StateContext` from within the class.
At the moment I'm hacking it by getting `this.constructor[META_KEY]` and then calling `.next` on the stateStream.

Please let me know what you think, and if you have any suggestions to improve it even further :)


amcdnl:
@leon -

Ya, I'm a NOT a fan of this!!!

```
// pass injector to base class so that we can get and set state
  // @todo, can we work around this?
  constructor(injector: Injector) {
    super(injector);
  }
```

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+  "typescript.tsdk": "node_modules/typescript/lib"
+}
```
amcdnl: Should end w/ new line

packages/store/src/entity-base.ts
```diff
@@ -0,0 +1,228 @@
...
+    return {
+      ids: [],
+      entities: {}
+    };
+  }
+
+  // @todo Don't like having to pass in an injector to be able to get at the
+  // state stream, how could we do this differently?
+  private stateStream: StateStream;
+  constructor(injector: Injector) {
```
amcdnl: State stream really shouldn't be exposed. Its only like that for plugins. I don't think I like the idea of using it here.

packages/store/src/entity-base.ts
```diff
@@ -0,0 +1,228 @@
...
+  static defaults() {
+    return {
+      ids: [],
+      entities: {}
+    };
+  }
+
+  // @todo Don't like having to pass in an injector to be able to get at the
+  // state stream, how could we do this differently?
+  private stateStream: StateStream;
```
amcdnl: Privates should be underscored.

packages/store/src/entity-base.ts
```diff
@@ -0,0 +1,228 @@
...
+
+    for (const entity of entities) {
+      didMutate = this._addOne(state, entity) !== EntityMutation.None || didMutate;
+    }
+
+    return didMutate ? EntityMutation.IdAndEntity : EntityMutation.None;
+  }
+
+  addAll(entities: T[]): S {
+    const state = this.getState();
```
amcdnl: Should `state` be a clone here rather than the instance?

deebloo:
I really don't like having to pass the injector :/ that is my only issue

packages/store/src/entity-base.ts
deebloo: boo

deebloo: :)

amcdnl: Its like that EVERYWHERE else in the code base ;P

packages/store/src/entity-base.ts
```diff
@@ -28,41 +29,68 @@ export enum EntityMutation {
   None
 }
 
-export abstract class EntityBase<T, S extends EntityState<T>> {
+export abstract class EntityBase<T, S extends EntityStateModel<T>> {
+  // @todo Don't like having to pass in an injector to be able to get at the
+  // state stream, how could we do this differently?
+  private _stateStream: StateStream;
```
amcdnl: What if we do what we did for `@Select` decorators for this? Make a static instance u can point to - https://github.com/ngxs/store/blob/master/packages/store/src/select.ts#L13

juracy:
I've just made a PR for this PR: https://github.com/ngxs/store/pull/387

juracy:
A new improvement in PR #399

filipjnc:
What's the status on this and can we help?

odahcam:
This looks interesting, is there any chance for this to continue?

splincode:
@markwhitfeld current PR is relevant?

splincode:
@leon https://github.com/ngxs-labs/entity-state/issues/6
development moved to separate repository

Issue #192, opened 2018-03-29
Extensibility for Action Handlers... (Action Pipes?)


markwhitfeld:
### Challenge
In order to preserve the simplicity of ngxs I think that it would be great to have an extensibility point for action handlers so that some of the more complex things can be abstracted into helpers.
Having used ngrx on a large project I have seen how developers struggle with ngrx a great deal due to having to learn rxjs. It would be great if it would be possible to implement a relatively complex solution using ngxs without having to resort rxjs to handle certain scenarios.
The cancellation of the previous request to prevent race conditions is just one example.
There is currently an open issue (#191) with a proposal to simplify things for cancellation using an option parameter for the @Action decorator. This is the same as what I proposed as an alternative in the description of issue #143, so I think it is on the right track.
What I am proposing here is an extension of that idea that allows for extensibility.

### Proposal
This proposal is essentially a copy of my comments ([here](https://github.com/ngxs/store/issues/143#issuecomment-375963611) and [here](https://github.com/ngxs/store/issues/143#issuecomment-376114463)) on issue #143.

What I am thinking is that we need to be able intercept and control the execution, inputs and outputs of the action handler. This sounds like a [Chain of Responsibility](https://en.wikipedia.org/wiki/Chain-of-responsibility_pattern) design pattern to me. Although, the actual implementation detail should have no bearing on its usage.

We could have the ability to add classes into the pipeline that processes the request.
These could be called one of the following: ActionFilter, ActionInterceptor, ActionHandler, ActionPipe
I think that I prefer ActionPipe because it fits that paradigm so I will use that word going forward.
I propose that the code could look like this:
```typescript
@Action(FeedAnimals, [ LatestCallWins ])
feedAnimals({ getState, setState }: StateContext<ZooStateModel>, { payload }: FeedAnimals) { 
  ...
  }
```
The Action Pipes could be expressed as an array ```@Action(FeedAnimals, [LatestCallWins, ...])``` or as extra parameters ```@Action(FeedAnimals, LatestCallWins, ...)```.
I deliberately tried to keep the `LatestCallWins` name expressive but it could also be `Cancellable`, `SwitchMap`, `CancelsPrevious`, `LatestCallCancelsPrevious` or similar. You would be able to list multiple Action Pipes in the same fashion as providers on a module (by type, by factory or by value).

Regarding the proposed solution for cancellation in issue #191, this could fit into the action pipe pattern in the following way (although the Options pipe could be inferred too):
```typescript
[...]

Issue #313, opened 2018-04-19
this.store.dipatch.subscribe(value => {}) in case of action success value is state, otherwise it is error


AlgoTrader:
this.store.dipatch.subscribe(value => {})

When action was success value is state
When action errored value is error

As a result, I always have to inspect value type in subscribe callback of any dispatch
Also 'DISPATHCED' and 'COMPLETED' are capitalized. but 'Errored' is first capitalized

deebloo:
I think that is a bug in that we don't want any value coming out of the subscribe. @amcdnl is that correct?

AlgoTrader:
I would prefer this.dispatch(new Action) to push value on success and throw on error

```
this.dispatch(new DoComplexStuff)
   .subscribe(successCb. errorCb)
```

Now if action throws, I get successCb with error value. I don't like values that I have to deep inspect to decide dispatch was success or failure

deebloo:
ah I see. Let me take a look and see how we are handling these errors

juristr:
> When action was success value is state

Also, the question is whether this is actually desired. Because right now, since a successful dispatch returns the entire state obj (which IMHO is useless), the following code from the docs also wouldn't work any more:

```
export class ZooComponent {
  @Select(state => state.animals) animals$: Observable<any>;
  constructor(private store: Store) {}
​
  addAnimal(name) {
    this.store.dispatch(new AddAnimal(name))
      .pipe(withLatestFrom(this.animals$))
      .subscribe(([animals]) => {
        // do something with animals
        this.form.reset();
      });
  }
}
```

Because now the `value` of the `.subscribe(..)` would be an array having as first value the entire state tree and as 2nd value (in this case) the last value from `animals$`.

AlgoTrader:
I also think the entire state in callback is useless. First, I can select any state slice if I need it. Second, I often want to know the result of the last dispatch. I would prefer to see the result of action but not the whole state

deebloo:
I agree with both of you. So there are two parts.

1. no state should back in the callback.
2. if there is an error it should fire the subscribers errors callback

AlgoTrader:
Exactly. Errors should be explicit

amcdnl:
I agree w/ @deebloo

amcdnl:
I believe this is resolved in 3.0

juristr:
@amcdnl "believing" is not enough Austin. We want proof 😉. As soon as I can I'm gonna test it and let you know in case

Issue #311, opened 2018-04-18
patchState seems to mutate the state


charlatan154:
### Versions

```
* ngxs: 2.0
* @angular/core: 5.2.8
```

I've build an `ngxs-freeze-plugin` for ngxs, mostly inspired from `@ngrx-store-freeze` meta reducer and it seems to work fine, except when using `pactState` from `StateContext`. The way you patch the state seems to mutate the current state slice, as the internal `getValue` function seems to return a value by reference, while passing `obj` as the default reduce value.
```
export const getValue = (obj: any, prop: string) =>
  prop.split('.').reduce((acc: any, part: string) => acc && acc[part], obj);
```

and then below, inside `for` loop, happens the mutation
```
let state = getState();
const local = getValue(state, metadata.depth);
for (const k in val) {
   local[k] = val[k];
}
```

One possible fix could be use a new copy while returning a value from `getValue` function like:
```
export const getValue = (obj: any, prop: string) =>
  prop.split('.').reduce((acc: any, part: string) => acc && acc[part], {...obj});
```
or even better pass a new copy while setting up the `local` slice inside `for` loop above.

[ngxs-freeze-plugin issue demo](https://stackblitz.com/edit/ngxs-freeze-plugin)

`setState` from `StateContext` seems to work as expected.

AlgoTrader:
Both setState and patchState do not like deep changes. Selectors use distinctUntilChanged and to make everyone happy, mutate state so that refererences for updated objects is also updated.

`The distinctUntilChanged operator will only emit when the next value is unique, based on the previously emitted value. In cases of numbers and strings, this means equal numbers and strings, in the case of objects, if the object reference is the same a new object will not be emitted.`

plattdl:
@amcdnl I think this would be nice. I am encountering this as well and would like some guidelines on how to approach large nested objects.

  My first guess would be to separate that large object into many separate sub states but that would be more work than I would like if there is a way to call setState in a way that these changes are picked up easier. Right now I just have an action that updates the object as a whole from an HTTP call with a few small changes here and there client side during the life of the session.

deebloo:
Ill add tests around this and update

Eikc:
heh, I was about to ask if you needed help and if so could point me in the right direction to fix the issue, anyway if you need any help, ping me and i will give it a try :) 

Thanks for a nice library ❤️ ❤️

AlgoTrader:
Mutating state is a very common problem

```
const state = getState()
[...]

PR #324, opened 2018-04-23
Action stream consolidation

markwhitfeld:
As discussed by the team, we need to consolidate the dispatch and action handler invocation to use the action stream as the mechanism for execution.
This is required for the implementation of Action Pipes (#192 ).

**Note:** Due to the large changes in this PR, please could you merge into master without squashing the commits. I have taken care to document the decisions along the way in each commit and to keep them relatively small. This will help with potential merge conflicts with other branches too.


markwhitfeld:
@amcdnl @deebloo @leon Please review. I think that this is ready for merge.

There are some particular points of importance to note:
- I added a canceled status to the action stream for canceled actions. I went for the US spelling of this (`canceled`) as opposed to the UK spelling (`cancelled`). I prefer the UK spelling but most frameworks seem to follow US spelling.
- The completed/errored/canceled updates to the action stream always happen asynchronously (delayed by a 0ms timer in the observable). This was done to avoid the misordering of events in the stream for subsequent observers. I can go into more detail with you directly if needed.
- The observable returned by the dispatch method will fire before the completed/errored/canceled updates on the action stream (the reasoning is related to the point above). You will find explicit tests related to this in the `action.spec.ts` class.
- When there is an error the value of the error is no longer sent to the dispatch observer. We will need to include the in the ActionContext if this is something we want back. It is actually discussed in issue #313 that there should be no returned value. I think that an error should hit the subscriber's error callback (with the actual error) and not the next callback. See Danny's comment [here](https://github.com/ngxs/store/issues/313#issuecomment-383072063).
- I extracted the getStateFn, setStateFn and dispatchFn into an InternalStateOperations interface and pushed its creation into `state-factory.ts` to remove some duplication and to try to satisfy the failing Code Climate error. Code Climate is still unhappy but I'm not sure how much value there would be in pursuing this removal of duplication further.

PS. I was tempted to do some further refactoring around the creation of the StateContext but given that there is an outstanding issue (#311 ) regarding the fact that updateState seem to mutate state I rather left that code so as to avoid merge conflicts.

markwhitfeld:
@deebloo I still need to wire up the stream directly to the handler. At the moment the stream is triggering the invokeActions call that the dispatcher used to call.
Essentially I have decoupled the triggering and return value of the action handlers from direct code calls and made it part of the stream processing albeit still a bit procedural inside at the moment.
I am doing this in gradual steps so that you we arrive at a totally different codebase overnight.

I also see pushing the stream further into the invoke actions code as part of the Action Pipes work. If you see this as something that should be part of this PR then I am happy to continue along that path. My concern would just be around merge hell if this branch sticks around for too long.

amcdnl:
> The observable returned by the dispatch method will fire before the completed/errored/canceled updates on the action stream (the reasoning is related to the point above). You will find explicit tests related to this in the action.spec.ts class.

Its very important this happens after completion.

> When there is an error the value of the error is no longer sent to the dispatch observer.

This is fine in my opinion as long as it does error.

packages/store/src/dispatcher.ts
```diff
@@ -0,0 +1,119 @@
...
+  }
+
+  private _dispatch(action: any): Observable<any> {
+    const prevState = this._stateStream.getValue();
+    const plugins = this._pluginManager.plugins;
+
+    return compose([
+      ...plugins,
+      (nextState, nextAction) => {
+        if (nextState !== prevState) {
```
amcdnl: Given the complexity of this function now, we should break it out into its own method.

packages/store/src/dispatcher.ts
```diff
@@ -0,0 +1,119 @@
...
+          filter((ctx: ActionContext) => {
+            return ctx.action === nextAction && ctx.status !== ActionStatus.Dispatched;
+          }),
+          take(1),
+          shareReplay()
+        );
+
+        actionResult$
+          .pipe(
+            delay(0) // Force completion onto new task to prevent completion from firing before dispatch event on subsequent observers
```
amcdnl: Very clever! 👏

packages/store/src/dispatcher.ts
```diff
@@ -0,0 +1,119 @@
...
+          )
+          .subscribe(ctx => {
```
amcdnl: Make this is lambda aka remove `{`

packages/store/src/dispatcher.ts
```diff
@@ -0,0 +1,119 @@
...
+            this._actions.next(ctx);
+          });
+        /*
```
amcdnl: Remove dead code.

packages/store/src/dispatcher.ts
```diff
@@ -0,0 +1,119 @@
...
+          */
+
+        this._actions.next({ action: nextAction, status: ActionStatus.Dispatched });
+
+        return actionResult$
+          .pipe(
+            exhaustMap((ctx: ActionContext) => {
+              switch (ctx.status) {
+                case ActionStatus.Completed:
+                  return of(this._stateStream.getValue());
```
amcdnl: So the subscribe result should not return a value. The reason is was before was the value is needed for the plugins `next().subscribe(state)`. Not sure how we can make the `dispatch().subscribe()` not return anything but make the `next().subscribe(state=>` work too. cc/ @deebloo

packages/store/src/state-factory.ts
```diff
@@ -114,20 +119,45 @@ export class StateFactory {
...
+              return <ActionContext>{ action, status: ActionStatus.Completed };
+            }),
+            defaultIfEmpty(<ActionContext>{ action, status: ActionStatus.Canceled }),
+            catchError(err => {
+              return of(<ActionContext>{ action, status: ActionStatus.Errored });
+            })
+          );
+        })
+      )
+      .subscribe(ctx => {
```
amcdnl: Make this a lambda.

packages/store/src/state-factory.ts
```diff
@@ -114,20 +119,45 @@ export class StateFactory {
     }
   }
 
+  connectActionHandlers(mappedStores: MappedStore[]): any {
+    if (this._connected) return;
+    this._actions
+      .pipe(
+        filter((ctx: ActionContext) => ctx.status === ActionStatus.Dispatched),
+        mergeMap(({ action }) => {
```
amcdnl: Are we sure we want `mergeMap` here and not `concatMap`?

packages/store/src/state-factory.ts
```diff
@@ -114,20 +119,45 @@ export class StateFactory {
...
+          );
+        })
+      )
+      .subscribe(ctx => {
+        this._actionResults.next(ctx);
+      });
+    this._connected = true;
+  }
+
+  private get rootStateOperations(): InternalStateOperations<any> {
```
amcdnl: Getters/Setters should be before methods.

deebloo:
@amcdnl 
"Also, just to double check if i do dispatch() of the same action twice really quickly, it will only complete when that EXACT action is completed not one of that type?"

this will only happen if you dispatch the exact same object.

```TS
const action = new Action();

store.dispatch(action);
store.dispatch(action)
```

amcdnl:
@deebloo - Makes sense. I think I'm ok with that. We could fingerprint each one but that might be overkill.

packages/store/src/dispatcher.ts
deebloo: @amcdnl yeah we would have to change plugins. Which I am certainly open to

packages/store/src/state-factory.ts
deebloo: for this case I don't think it really matters. as long as it isn't switchMap

packages/store/src/actions-stream.ts
```diff
@@ -7,6 +7,7 @@ import { Observable, Subject } from 'rxjs';
 export enum ActionStatus {
   Dispatched = 'DISPATCHED',
   Completed = 'COMPLETED',
+  Canceled = 'CANCELED',
   Errored = 'Errored'
```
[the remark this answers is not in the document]
markwhitfeld: Done. This is actually related to part of issue #313

packages/store/src/dispatcher.ts
markwhitfeld: Done

packages/store/src/dispatcher.ts
markwhitfeld: Thanks

packages/store/src/dispatcher.ts
markwhitfeld: Done, although I prefer having them in a statement block because you can then put a breakpoint there if necessary.

packages/store/src/dispatcher.ts
markwhitfeld: Done

packages/store/src/dispatcher.ts
markwhitfeld: I think that I will have to leave it like this for now so that I don't break the plugins.

packages/store/src/state-factory.ts
markwhitfeld: Done, although I prefer having them in a statement block because you can then put a breakpoint there if necessary.

packages/store/src/state-factory.ts
markwhitfeld: Done. I moved it to the top.

markwhitfeld:
@amcdnl I think that the PR is ready now, please review again.
If it is possible please don't squash the commits as I made sure they all took logical steps and documented the progress as I solved various aspects of the PR.

packages/store/src/actions-stream.ts
```diff
@@ -7,19 +7,39 @@ import { Observable, Subject } from 'rxjs';
...
+  Canceled = 'CANCELED',
+  Errored = 'ERRORED'
 }
 
 export interface ActionContext {
   status: ActionStatus;
   action: any;
 }
 
+export class OrderedSubject<T> extends Subject<T> {
```
markwhitfeld: @deebloo I think you might like this.

packages/store/src/actions-stream.ts
```diff
@@ -16,11 +16,30 @@ export interface ActionContext {
...
```
markwhitfeld: @deebloo I think you might like this.

deebloo:
idk about not squashing. this is one feature and if it ever needs to be reverted or picked it will be difficult. All of the commits will be listed in the description

amcdnl:
Ya, I agree w/ @deebloo - I think squash is the right thing given all the commits.

PR #333, opened 2018-04-26
 chore(store): add lazy loading unit tests

deebloo:
I want to simplify store init and modules. this is the first step to ensuring nothing breaks when that happens


deebloo:
try to find why it is failing. works locally and does not work with "websocket" instead of "webSocket"

deebloo:
@amcdnl do you now what is going on here?

amcdnl:
@deebloo - can u rebase?

PR #339, opened 2018-04-28
chore: update docs

dtslvr:
Harmonize bullet list


deebloo:
@dtslvr can you change your commit so adhere to the docs? after that I will merge

PR #342, opened 2018-04-30
fix(store): only wire action onces

amcdnl:
- Fix actions being dispatched multiple times
- Remove unused variables
- enhance tests

### DIFF
--- a/packages/store/src/module.ts
+++ b/packages/store/src/module.ts
@@ -36,8 +36,10 @@ export class NgxsRootModule {
       stateStream.next({ ...cur, ...results.defaults });
     }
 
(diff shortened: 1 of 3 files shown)


packages/store/src/module.ts
```diff
@@ -36,8 +36,10 @@ export class NgxsRootModule {
       stateStream.next({ ...cur, ...results.defaults });
     }
 
-    factory.connectActionHandlers(results.states);
```
markwhitfeld: This parameter was in place because it will be required when we do Action Pipes. We can bring it back when needed ;-)

PR #345, opened 2018-05-01
fix(store): fix enum compile out

amcdnl:
- Fixes enums getting compiled out at prod build time
- Fixes race case when loading router module in production
- Properly imports the router module using `@dynamic`
- Minor code cleanup


leon:
LGTM

PR #360, opened 2018-05-06
Added El Terminador, an Electron-based Terminal Manager

mflorence99:
This project may be of interest to the ngxs community. It has just been updated to use ngxs 3, and consequently Angular 6 and RxJS 6. I use it every day for all CLI activity.


amcdnl:
Thats Awesome! Thanks for sharing!

PR #363, opened 2018-05-07
Renaming ArchiveProject to Project

civicwar:
By example, there was no verb added to the ArchiveProject, 
so I think the entity to act upon is Project and the verb is Archive.


amcdnl:
Thank you!

Issue #51, opened 2018-03-07
Select Combinations


amcdnl:
Add ability to combine selects via decorators:

```
@Select(
   {  foo: 'foo.bar', bar: (state) => state.bar },
   (({ foo, bar }) => foo.filter(f => f !== bar)))
bars: Observable<Bar>;
```

amcdnl:
@deebloo - I have an idea for this, will be pretty easy to implement too.

```TS
@State({ ... })
export class ZooState {
   @Selector({
       combine: [CowState.cows, ...n*]
    })
    static animals(state: ZooStateModel, cows: Cows, ...n*) { ... }
}
```

leon:
I would prefer

    @Selector([CowState.cows, FarmState.milktime])

amcdnl:
Ya, I was trying to keep it where if we wanted to add other options down the road it would be easy but ya I agree

markwhitfeld:
I have spent some time thinking about this feature. 
I have tried to define clear rules on what would be expected from this.
Here are my thoughts on this:
- (**This is the existing usage**) A `@Selector` decorator with zero parameters is only valid in a state class and provides the container class' state model as the only parameter :
```typescript
@State<UserStateModel>( ... )
export class UserState {
  @Selector()  
  getUsers(userStateModel: UserStateModel) { ... }
}
```
- Allow state classes to be passed in as parameters. This can be used from within any class, be it a State class or another service class. It only provides the models as requested (ie. it does not add the model for the containing class, no matter if it is defined in a state class or not) :
```typescript
export class SomeStateClassOrRegularService {
  @Selector([UserState, TodoState])  
  getActiveUserTodos(userState: UserStateModel, todoState: TodoStateModel) { ... }
}
```
- Allow selectors to be passed in. The rules would be the same as passing State Classes as parameters:
```typescript
export class SomeStateClassOrRegularService {
  @Selector([UserState.getName, TodoState.getCount])  
  getUserTodoCount(userName: string, todoCount: number) { ... }
}
```
- Or a mix of the two:
```typescript
export class SomeStateClassOrRegularService {
  @Selector([UserState.getName, TodoState])  
  getUserTodos(userName: string, todoState: TodoStateModel) { ... }
}
```
- A `@Selector` with no parameters would not be valid on a class that is not a state. There was an option that the global state could then be passed as the parameter, but this would break encapsulation because it would require the dev to have knowledge of the global state structure. 
To be clear, this would **not be valid**:
[...]

PR #364, opened 2018-05-07
feat(selector): add multi select support

amcdnl:
Adds support for multiple selects #51

### DIFF
--- a/packages/store/src/store.ts
+++ b/packages/store/src/store.ts
@@ -21,7 +21,7 @@ export class Store {
   /**
    * Selects a slice of data from the store.
    */
-  select<T>(selector: (state: any) => T): Observable<T>;
+  select<T>(selector: (state: any, states?: any[]) => T): Observable<T>;
   select(selector: string | any): Observable<any>;
   select(selector: any): Observable<any> {
     if (selector[META_KEY] && selector[META_KEY].path) {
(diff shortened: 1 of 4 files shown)


packages/store/src/store.ts
```diff
@@ -21,7 +21,7 @@ export class Store {
   /**
    * Selects a slice of data from the store.
    */
-  select<T>(selector: (state: any) => T): Observable<T>;
+  select<T>(selector: (state: any, states?: any[]) => T): Observable<T>;
```
markwhitfeld: This does not seem to match the signature  executed by the selector above.
Shouldn't it rather be: `selector: (state: any, ...states?: any[]) => T)`

packages/store/src/store.ts
```diff
@@ -47,7 +47,7 @@ export class Store {
   /**
    * Select one slice of data from the store.
    */
-  selectOnce<T>(selector: (state: any) => T): Observable<T>;
+  selectOnce<T>(selector: (state: any, states?: any[]) => T): Observable<T>;
```
markwhitfeld: Same comment as above.

packages/store/src/store.ts
```diff
@@ -56,7 +56,7 @@ export class Store {
   /**
    * Select a snapshot from the state.
    */
-  selectSnapshot<T>(selector: (state: any) => T): T {
+  selectSnapshot<T>(selector: (state: any, states?: any[]) => T): T {
```
markwhitfeld: Same comment as above.

packages/store/src/store.ts
amcdnl: #facepalm

PR #383, opened 2018-05-16
fix(perf): run actions outside zones for perf

amcdnl:
Runs actions outside of zones for performance reasons.


deebloo:
i mean it seems fine to me. do we have anything to show what the performance improvement would actually be?

markwhitfeld:
Yes, I think it would be great if we could have some sort of comparison on performance improvement to go with this... Possibly in the docs as a motivation why.

Do you know what the potential impact of running angular services outside the angular zone would be? Http calls, etc.

PR #385, opened 2018-05-17
Error in AddAnimal action in first example

juracy:
There was an error, importing an action and declaring in same module.


juracy:
@amcdnl in original docs there was:
```
import { AddAnimal } from './animal.actions';
 	 
export class AddAnimal {	
  static readonly type = '[Zoo] Add Animal';	
  constructor(public name: string) {}	
}
```

Import and export the same symbol (`AddAnimal`) raises an exception: `error TS2440: Import declaration conflicts with local declaration`

amcdnl:
Haha gotcha, thanks!

PR #387, opened 2018-05-18
fix: mutations were overwriting the entire store

juracy:
Mutations were changing the entire store instead of only the correct state


leon:
Thanks!

PR #394, opened 2018-05-24
fix(store): move the subscribe into a zone

amcdnl:
This moves the `subscribe` function of the dispatch back into a angular zone.

### DIFF
--- a/packages/store/src/dispatcher.ts
+++ b/packages/store/src/dispatcher.ts
@@ -41,7 +41,7 @@ export class InternalDispatcher {
     });
 
     result.subscribe({
-      error: error => this._errorHandler.handleError(error)
+      error: error => this._ngZone.run(() => this._errorHandler.handleError(error))
     });
 
     return result.pipe(enterZone(this._ngZone));
--- a/packages/store/src/dispatcher.ts
+++ b/packages/store/src/dispatcher.ts
@@ -6,6 +6,7 @@ import { compose } from './compose';
 import { InternalActions, ActionStatus, ActionContext } from './actions-stream';
 import { StateStream } from './state-stream';
 import { PluginManager } from './plugin-manager';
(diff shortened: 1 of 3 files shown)


packages/store/src/zone.ts
```diff
@@ -0,0 +1,19 @@
...
+ */
+export function enterZone<T>(zone: NgZone) {
+  return (source: Observable<T>) => {
+    return new Observable((sink: Observer<T>) => {
+      return source.subscribe({
+        next(x) {
+          zone.run(() => sink.next(x));
+        },
+        error: sink.error.bind(sink),
+        complete: sink.complete.bind(sink)
```
markwhitfeld: Shouldn't error and complete also be run in the zone?

packages/store/tests/dispatch.spec.ts
```diff
@@ -57,6 +57,28 @@ describe('Dispatch', () => {
...
+      increment() {
+        expect(NgZone.isInAngularZone()).toBe(false);
+      }
+    }
+
+    TestBed.configureTestingModule({
+      imports: [NgxsModule.forRoot([MyState]), NgxsModule.forFeature([])]
+    });
+
+    const store: Store = TestBed.get(Store);
```
markwhitfeld: I think that we should run the whole test in the Angular zone and test that we are in the zone before we call `dispatch` just to be explicit. The test will then demonstrate that we go from the angular zone, out of it, and back into the zone.

packages/store/src/dispatcher.ts
```diff
@@ -43,7 +44,7 @@ export class InternalDispatcher {
       error: error => this._errorHandler.handleError(error)
```
markwhitfeld: Should the error handler also be run in the angular zone? What is a reasonable expectation here?

PR #398, opened 2018-05-25
fix: selector function should still be usable as a function

markwhitfeld:


leon:
LGTM

Great that you fixed this. Needed it the other day :)

amcdnl:
img

deebloo:
This means using parent state classes is valid when you need selectors across sibling classes

PR #399, opened 2018-05-26
feat: possibility to serialize into a instance of model (entity)

juracy:
A new method serialize to do simple copy of an instance, but can be override to serialize back to model


juracy:
Maybe the Object.assign in _mutateStateIfNeeded method should be replaced with serialize call too...

Anyway, "serialize" is a bad name, but "deserialize" give the idea of retrieving data... Any thoughts?

juracy:
Another solution is save in state only simple objects (even in AddOne) and serialize back with selectors

eranshmil:
I guess it won't work if you have a more complex model, for example:
```
class Project {
  id: string;
  name: string;
  user: User;
}
```

juracy:
@eranshmil It would be necessary override the `serialize` method in your class and made your own serializer:

```typescript
serialize(original, changes = {}): Project {
  const result = Object.assign(new Project(), original, changes);
  Object.assign(result.user, changes.user);
  return result;
}

```

Anyway, I'm evaluating working without this PR and make serializations on @Selector

splincode:
@juracy it's relevant?

juracy:
> @juracy it's relevant?

@splincode I don't believe so...

PR #401, opened 2018-05-30
feat(storage): migrations

amcdnl:
Adds the ability to migrate data.

### DIFF
--- a/packages/storage-plugin/src/storage.plugin.ts
+++ b/packages/storage-plugin/src/storage.plugin.ts
@@ -1,13 +1,7 @@
 import { Inject, Injectable } from '@angular/core';
 import { NgxsPlugin, setValue, getValue, InitState, UpdateState, actionMatcher } from '@ngxs/store';
 
-import {
-  NgxsStoragePluginOptions,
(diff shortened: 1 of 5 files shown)


packages/storage-plugin/src/storage.plugin.ts
```diff
@@ -30,7 +37,18 @@ export class NgxsStoragePlugin implements NgxsPlugin {
...
           }
 
-          if (key !== '@@STATE') {
+          if (options.migrations) {
+            options.migrations.forEach((strategy: MigrationStrategy) => {
+              const versionMatch = strategy.version === getValue(val, strategy.versionKey || 'version');
+              const keyMatch = (!strategy.key && isMaster) || strategy.key === key;
+              if (versionMatch && keyMatch) {
+                val = strategy.migrate(val);
+                hasMigration = true;
```
deebloo: can't we just assume that hasMigration is true if there are migrations in the list?

amcdnl: What if no migrations matched? No reason to set it then.

PR #405, opened 2018-06-01
feat: payload and dispatch decorators

maktarsis:
Hey. I create a new way dispatching actions.
That wasn't smooth. But pretty good alternative.

## General notes:
1. Documentation must be updated after merge
2. Decorators **dispatch** and **payload** are **optional**
3. **Dispatch** decorator is interlayer that just dispatching actions 
4. **Payload** decorator is runtime typing function. You can use just dispatch decorator (isn't safe).

### Pros:
1. ngxs way
2. simpler
3. no store injection

### Cons:
1. new moves in .actions.ts 
    - payload decorator
    - action must be class and props in constructor
2. runtime typing

## Proposals
1. **Declare action model in .actions.ts instead .state.ts**
  image
  image
2. **set defaults as instance** 
  image
  image

Also, created change-value module as use case of new decorators. If you wish, I'll provide more simpler & difficult examples in this module.

## Use case:
1. image
2. image

**Please ask questions and leave your opinion.
I will be glad to answer**

### DIFF
--- a/integration/app/app.component.ts
+++ b/integration/app/app.component.ts
@@ -40,9 +40,7 @@ import { AddTodo, RemoveTodo, TodoState, SetPrefix, TodosState, LoadData } from 
           🐼
         </li>
       </ul>
-      
       <app-change-value></app-change-value>
-      
       <router-outlet></router-outlet>
     </div>
   `,
--- a/integration/app/app.component.ts
+++ b/integration/app/app.component.ts
@@ -40,6 +40,9 @@ import { AddTodo, RemoveTodo, TodoState, SetPrefix, TodosState, LoadData } from 
           🐼
         </li>
       </ul>
+      
+      <app-change-value></app-change-value>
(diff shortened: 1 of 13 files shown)


amcdnl:
Thanks for contributing, this is interesting. A few notes:

1) We can't use classes for options. We are dealing with immutable objects rather than instances.
2) I noticed your selector where you just returned the state. You actually don't have to do that, you can just pass the class and it will return the state for you.
3) Runtime checking is notoriously slow and in typescript we rely on design time checking.
4) It feels like this adds more code rather less?

I'm also not entirely sure I understand what `Payload` and `Dispatch` decorators give us. I *think* you are trying to get rid of store dispatch? If thats the case then it would not be a good candidate for this library since that would be considered domain driven pattern.

I'm going to close this for now given all those comments. If you want to publish independently I'll gladly link to it though.

PR #406, opened 2018-06-01
Plugin state operation extension point

markwhitfeld:


amcdnl:
I think this might be over board for what we are trying to accomplish with this and I'm also not sure I want to expose this api.

PR #407, opened 2018-06-02
Consolidate internal state operations

markwhitfeld:
This consolidation is required for the freeze functionality of the development mode.
This is so that all internal `dispatch` and `setState` operations go through a single point so that they can be intercepted and frozen. This could also be a useful point to gather debug information.
 
PS. Kept the `StateOperations` interface internal and reverted the plugin interface (as opposed to the last PR).

### DIFF
--- a/packages/store/src/internals.ts
+++ b/packages/store/src/internals.ts
@@ -17,7 +17,7 @@ export interface ActionHandlerMetaData {
   type: string;
 }
 
-export interface InternalStateOperations<T> {
+export interface StateOperations<T> {
   getState(): T;
   setState(val: T);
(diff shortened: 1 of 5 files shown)


amcdnl:
Just realized devtools uses the state stream, probably need to update that too!

markwhitfeld:
@amcdnl hmmm... the class is called `InternalStateOperations` because it is intended for internal use only. Would you be ok with it being called from the dev tools plugin? (it is not part of the items exported from `public_api`)

amcdnl:
Well we need to ensure all operations go thru the same avenue so it feels like we do need that to be exposed.

PR #410, opened 2018-06-02
Memoize inner selector function

markwhitfeld:


packages/store/src/memoize.ts
```diff
@@ -1,23 +1,40 @@
+function defaultEqualityCheck(a, b) {
+  return a === b;
+}
+
+function areArgumentsShallowlyEqual(equalityCheck, prev, next) {
+  if (prev === null || next === null || prev.length !== next.length) {
+    return false;
+  }
+
+  // Do this in a for loop (and not a `forEach` or an `every`) so we can determine equality as fast as possible.
```
amcdnl: I thought you could break out of a forEach by returning false.  Just a side note.

markwhitfeld: This is the code from the reselect redux library, so I assume it was done this way for fine tuned performance

amcdnl: Ya, `while` used to be fastest but I think thats changed recently. https://jsperf.com/for-vs-foreach/333

PR #414, opened 2018-06-04
fix(store): add deep-freeze-strict to peerDeps

### LINKED ISSUE #412: Module not found: Error: Can't resolve 'deep-freeze-strict' in '...../node_modules/@ngxs/store/fesm5'
Angular version: 6.0.3
ngxs: 3.1.0
rxjs: 6.1
rxjs-compat: 6.1
typescript: 2.7.1

jotatoledo:
Add missing (peer)dependecy of the store module.  Although it seems that
defining lib. deps through peerDeps is the preferred way (see https://github.com/dherges/ng-packagr/pull/687) it feels weird to force consumers to explicitly add this to their
application deps.

A better approach would be to bundle the deep-freeze module into the store package.

Closes #412

### DIFF
--- a/packages/store/package.json
+++ b/packages/store/package.json
@@ -4,6 +4,7 @@
   "license": "MIT",
   "sideEffects": true,
   "peerDependencies": {
+    "deep-freeze-strict": "^1.1.1",
     "@angular/core": ">=5.0.0 <7.0.0",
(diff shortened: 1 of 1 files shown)


amcdnl:
Fixed in 3.1.1

Issue #373, opened 2018-05-10
Logger collapsed option is not working as expected


un33k:
```
[x] Bug report
```

Neither `NgxsLoggerPluginModule.forRoot()` nor ` NgxsLoggerPluginModule.forRoot({ collapsed: true })` work as expected. (image included)

### Versions

```
* ngxs: 3.0.1
* @angular/core: 6.0.0
```

### Repro steps

* Enable the logger plugin
* Ensure some actions are triggered
* Open Console and look at the logs

Here is a screenshot ... 

![screen shot 2018-05-10 at 3 19 43 pm](https://user-images.githubusercontent.com/407537/39889344-9b173844-5465-11e8-94e7-f7ba6168bdcd.png)

### Desired behavior

![screen shot 2018-05-10 at 3 23 14 pm](https://user-images.githubusercontent.com/407537/39889512-1bd2d5ba-5466-11e8-9756-fdd67006c1bd.png)

acostaf:
Hi All

Also disabled is not working 

e.g NgxsLoggerPluginModule.forRoot({ logger: console, collapsed: false, disabled: true }),

mehrad-rafigh:
Hi All. Having the same issue while using the environment file from angular. 

`NgxsLoggerPluginModule.forRoot({collapsed: true, disabled: environment.production}),`

![image](https://user-images.githubusercontent.com/4339673/40575484-07e740d6-60e6-11e8-85aa-94d3648f9072.png)

kattsushi:
i have the same issue :(

martijn1967:
Me too

PR #420, opened 2018-06-05
fix(logger): fix logger not using custom options

Thenis:
made loggerOptionFactory to use custom options

no breaking changes. Fix for issue #373


amcdnl:
Thanks!

Issue #413, opened 2018-06-04
Usage of arguments in Selectors


piernik:
In `3.1` You introduced arguments in Selectors. How to use them? Here is my code which is not working.

	@State<ConfigStateModel>({
	  name: 'config',
	  defaults: {
		clients: {
			schema: {}
		},
		incomes: {
			schema: {}
		},
	  }
	})
	export class ConfigState {
	  @Selector()
	  static getSchema(type) {
		return (state) => {
		  return state[type].schema;
		};
	  }
	}
	
Usage in component

	export class ClientsListComponent {
		@Select(ConfigState.getSchema('clients'))
		schema$: Observable<any>;
	}

leon:
You spelled `ConfigState.getSchemat` wrong?

also `state[type]` could be null if type doesn't exist.

But we need more info on what is happening.
Could you please put together a stackblitz with your problem. It will make it much easier to debug and propose a solution.

piernik:
I did misspelled but it was fast example and it is not an issue.

In standard `@Select` You pass reference to static method `@Select(ConfigState.getClientsSchema)` You do not execute it like `@Select(ConfigState.getClientsSchema())`.

How to pass argument to selector?
`@Select(ConfigState.getSchema('clients'))` is not working.

evertonrobertoauler:
Same problem here, see my stackblitz to reproduce the issue.

https://stackblitz.com/edit/angular-nyzr88

markwhitfeld:
There was a team misunderstanding over the signature of the dynamic selectors.
See this for original proposal: 
https://github.com/ngxs/store/issues/386#issuecomment-390181710
It has now been corrected in the docs. See here: 
https://ngxs.gitbook.io/ngxs/concepts/select#memoized-selectors
@amcdnl has also fixed the example in his announcement post.

Your example would be:
```typescript
@State<ConfigStateModel>({
  name: 'config',
  defaults: {
	clients: {
		schema: {}
	},
	incomes: {
		schema: {}
	},
  }
})
export class ConfigState {
  @Selector()
  static getSchema(state) {
	return (type) => {
	  return state[type].schema;
	};
  }
}
```
Usage in component
```typescript
export class ClientsListComponent {
	@Select(ConfigState.getSchema)
	schemaFn$: Observable<(type: string) => any>;

	get clientsSchema$() {
	    return this.schemaFn$.pipe(map(fn => fn('clients')));
	}        
}
```

zygimantas:
[...]

PR #426, opened 2018-06-07
feat(selector): enable the creation of selectors with arguments (#413)

evertonrobertoauler:
Hi team, I believe this new feature will bring great value to this awesome library, this feature enables what can be easily done in `@ngrx/store` using the `createSelector` function:

```ts
export function getSchemaSelector(type: keyof ConfigStateModel) {
   return createSelector(state => state[type].schema);
}
```

This new decorator will work as a factory to create selectors, with preconfigured parameters, here is an example of the syntax:

```ts
@State<ConfigStateModel>({
  name: 'config',
  defaults: { clients: { schema: {} }, incomes: { schema: {} } }
})
export class ConfigState {
  @SelectorCreator()
  static getSchema(type: keyof ConfigStateModel) {
    return (state: ConfigStateModel) => state[type].schema;
  }
}

@Component(...)
export class ClientsListComponent {

  @Select(ConfigState.getSchema('clients'))
  public clientSchema$: Observable<any>;

}
```

Here I have a working demo, using this new feature:
https://stackblitz.com/edit/angular-v89xad?embed=1&file=src/app/app.module.ts

### DIFF
--- a/packages/store/src/public_api.ts
+++ b/packages/store/src/public_api.ts
@@ -7,4 +7,5 @@ export { Actions } from './actions-stream';
 export { ofAction, ofActionSuccessful, ofActionDispatched, ofActionErrored } from './of-action';
 export { NgxsPlugin, NgxsPluginFn, StateContext, NgxsOnInit } from './symbols';
 export { Selector } from './selector';
+export { SelectorCreator } from './selector-creator';
 export { getActionTypeFromInstance, actionMatcher } from './utils';
--- a/packages/store/src/selector-creator.ts
+++ b/packages/store/src/selector-creator.ts
@@ -1,0 +1,19 @@
+import { Selector } from './selector';
+
+export function SelectorCreator(selectors?: any[]) {
+  return (target: any, key: string, descriptor: PropertyDescriptor) => {
+    if (descriptor.value !== null) {
+      return {
+        configurable: true,
+        get() {
+          return function factory(...args) {
+            const value = descriptor.value(...args);
+            return Selector(selectors)(target, key, { ...descriptor, value }).get();
+          };
+        }
+      };
+    } else {
+      throw new Error('SelectorCreators only work on methods');
+    }
+  };
+}
--- a/packages/store/tests/selector-creator.spec.ts
+++ b/packages/store/tests/selector-creator.spec.ts
@@ -1,0 +1,102 @@
+import { async, TestBed } from '@angular/core/testing';
+import { State, SelectorCreator } from '@ngxs/store';
+import { Store } from '../src/store';
+import { NgxsModule } from '../src/module';
+import { Selector } from '../src/selector';
+
+describe('SelectorCreator', () => {
+  interface MyStateModel {
+    foo: string;
+    bar: string;
+  }
+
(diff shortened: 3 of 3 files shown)


amcdnl:
You should be able to do this now.

```
@Selector()
static foo(state) {
   return (arg) => {};
}
```

evertonrobertoauler:
Thanks, but that means that I need to add this function call boilerplate `.pipe(map(fn => fn(...args)))`  all over my code when I need to pass argument to a selector?

```ts
 this.product$ = this.store.select(ProductState.getProductById).pipe(map(fn => fn('123')));
```

Wouldn't it be nicer if I could do something like this instead? 

```ts
this.product$ = this.store.select(ProductState.getProductById('123'));
```

Why can't we have a way to create selectors on the fly, like a factory that creates selectors? 

Does it have some performance penalties?

Does it go against the redux pattern  or the @ngxs/store mental model?

markwhitfeld:
Thanks for the PR.
@evertonrobertoauler maybe adjust the description to show what new syntax this would introduce. 
@amcdnl please see issue #413 for context of this PR.
I do believe that it has its own specific use cases. This is not the same as the dynamic selector parameters pattern I proposed. It is more of a selector factory function to create a new selector that has a preconfigured parameter.

piernik:
@amcdnl This is great addition. It reduces boilerplate and takes all logic to selectors. Is it not the purpose of Ngxs to reduce boilerplate?

amcdnl:
Ok, I see. I'd like to try to make this work w/ the current API so we don't have to introduce a new decorator. Do you think this is possible @markwhitfeld ?

markwhitfeld:
@amcdnl It is possible but would require some option of sorts because you would not be able to detect this at runtime.
Something like :
`@Selector([...], { factory: true}) ... `
I really feel that this would be abusing the existing `@Selector` decorator though and would be making things quite cryptic.
I prefer the additional `@CreateSelector` decorator.

amcdnl:
I'm good with making it an option to the selector, though with the current API it might be tricky to infer the options over a join function but possible.

markwhitfeld:
@amcdnl If we did make it an option of the `@Selector` decorator it would make type checking virtually impossible. As a separate decorator the type checking would be pretty straight forward.

You may see that my opinion here is very different to the `@Dispatch` decorator proposed in another PR. This is because I believe that the benefit here would certainly be worth the cost of the new decorator.

As an alternative we could expose a `createSelector` method that could be used directly to enable the syntax used by ngrx. For example:
```typescript
@State<ConfigStateModel>({
  name: 'config',
  defaults: { clients: { schema: {} }, incomes: { schema: {} } }
})
export class ConfigState {
  static getSchema(type: keyof ConfigStateModel) {
    return createSelector((state: ConfigStateModel) => state[type].schema);
  }
}
```
What do you think?

evertonrobertoauler:
@markwhitfeld This createSelector function looks nice and simple, I believe it solves well the issue, Thanks

markwhitfeld:
@evertonrobertoauler I have exposed the `createSelector` function in PR #484

PR #431, opened 2018-06-08
Fix inconsistencies in state documentation

dominicm00:


docs/concepts/state.md
```diff
@@ -65,7 +65,7 @@ export interface ZooStateModel {
 })
 export class ZooState {
   @Action(FeedAnimals)
-  feedAnimals(ctx: StateContext<ZooStateModel>, action: FeedAnimals) {
+  feedAnimals(ctx: StateContext<ZooStateModel>) {
```
markwhitfeld: I don't think we gain anything through removing the parameter. It could cause confusion because of its omission. We are demonstrating the API, so this should remain clear (even if the new version of VS Code wants to remove the unused parameter ;-) ) 
Could you revert this?

amcdnl: On the other hand, TS might complain unused variables?

dominicm00: The `action` parameter isn't mentioned until the next example, so removing it might help narrow discussion to the `StateContext` parameter and show the "best practice" for a simple action.

If the parameter is added back in, the explanation of the example needs to be updated since it refers to the simple action function as having only one parameter.

PR #445, opened 2018-06-14
feat: expose RouterStateModel in @ngxs/router-plugin

juracy:
I need these imports:

```
import { RouterState } from '@ngxs/router-plugin';
import { RouterStateModel } from '@ngxs/router-plugin/src/router.state';
```

to use this:
```
this.store.select(RouterState)
  .subscribe((route: RouterStateModel) => console.log(route.state));
```

And I think it should be better:

```
import { RouterState, RouterStateModel } from '@ngxs/router-plugin';
```

### DIFF
--- a/packages/router-plugin/index.ts
+++ b/packages/router-plugin/index.ts
@@ -1,3 +1,3 @@
 export { NgxsRouterPluginModule } from './src/router.module';
-export { RouterState } from './src/router.state';
+export { RouterState, RouterStateModel } from './src/router.state';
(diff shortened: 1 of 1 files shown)


juliusstoerrle:
I don't mind, but shouldn't typescript know/infer the type in the next function by itself? It does it for my normal subscriptions!

juracy:
@juliusstoerrle Using the syntax of select: `this.store.select(RouterState)` there isn't a type inference...

PR #447, opened 2018-06-16
feat: dispatch decorator, implement decorator factory

maktarsis:
Re-solution of #405 .

- It's easier to understand (@Dispatch)
- NGXS way (decorator)
- No store injection
- check at compile time instead of runtime as in #405 

Also, I created a decorator factory instead of selector factory, that provides store to selector.ts and dispatch.ts.

Please ask questions and leave your opinion.
I will be glad to answer

### DIFF
--- a/integration/app/app.component.ts
+++ b/integration/app/app.component.ts
@@ -4,6 +4,7 @@ import { Observable } from 'rxjs';
 import { FormBuilder, FormArray } from '@angular/forms';
 
 import { AddTodo, RemoveTodo, TodoState, SetPrefix, TodosState, LoadData } from './todo.state';
+import { Dispatch } from '@ngxs/store';
 
 @Component({
   selector: 'app-root',
@@ -65,8 +66,9 @@ export class AppComponent {
 
   constructor(private store: Store, private formBuilder: FormBuilder) {}
 
+  @Dispatch()
   addTodo(todo: string) {
-    this.store.dispatch(new AddTodo(todo));
+    return new AddTodo(todo);
   }
 
   removeTodo(index: number) {
--- a/packages/store/src/decorators/decorator-factory.ts
+++ b/packages/store/src/decorators/decorator-factory.ts
@@ -1,0 +1,14 @@
+import { Injectable } from '@angular/core';
+import { Store } from '../store';
+
+/**
+ * Allows decorator to get access to the DI store.
+ * @ignore
+ */
+@Injectable()
+export class DecoratorFactory {
+  static store: Store | undefined = undefined;
+  constructor(store: Store) {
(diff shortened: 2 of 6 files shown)


packages/store/src/decorator-factory.ts
```diff
@@ -0,0 +1,14 @@
+import { Injectable } from '@angular/core';
+import { Store } from './store';
+
+/**
+ * Allows decorator to get access to the DI store.
+ * @ignore
+ */
+@Injectable()
+export class DecoratorFactory {
```
amcdnl: We already have a select factory we can just rename to be more generic use that.

amcdnl:
Ok, this is interesting. I'm not sure what I think about it yet but its interesting for sure.

So if I'm following correctly, you would do this and it would call dispatch for you.

```
@Dispatch((todo: string) => new AddTodo(todo))
addTodo(todo: string) {}
```

It feels odd to have the empty addTodo method is my first thought. I mean this could be done just the same and it doesn't require any special code.

```
class Foo {
   addTodo (todo: string) => this.store.dispatch(new AddTodo(todo));
  onClick() { this.addTodo('foo'); }
}
```

I do see where you are going with this though. Its kinda like the `HostListener` or something in Angular.

@markwhitfeld - curious your thoughts.

packages/store/src/decorator-factory.ts
maktarsis: Okay, now I should to explain what is decorator factory. 

In short, it's just refactoring.

Decorator factory is refactored selector factory which was moved to own file service file. Why? Because right now, decorator factory provides service for selector and dispatch decorators.

markwhitfeld:
@maktarsis Thanks for making those changes, although I still don't think it provides enough value to merit the increase in the footprint of the API. 
Depending on if @amcdnl accepts this or not you could also provide the decorator as a gist which we could link from the docs for anyone that wants to use this syntax.

maktarsis:
@markwhitfeld 

Nonsense. I totally disagree.

For now, user can use this signature, without store injection. Old way is still available for users. For example, you can use old signature, but ngxs users will use new way to dispatch actions with pleasure.

```
@Dispatch() 
addTodo(todo: string) {
  return new AddTodo(todo);
}
```

juliusstoerrle:
I can see hardly any use cases for this and can only agree @markwhitfeld on the increase of ambiguity.
I regularly have new developers on my team and would prefer a slim, unambigous and simple API for them to learn.

amcdnl:
I agree w/ the consensus. I would encourage you to create a library and we can link it in the docs :).

juracy:
I really liked the possibility to dispatch an action without injecting the store, similar to @Select decorator.

PR #456, opened 2018-06-21
fix: action stream should return in Angular zone

### LINKED ISSUE #415: Change detection stopped working in production
### Versions
```
* ngxs: 3.1.1/3.1.2
* @angular/core: 6.0.3
```

### Repro steps

* In smart component, add ```@Select(ZooState.panda) public panda$!: Observable<Panda>;```
* In the same component, create subscription:
```this.panda$.subscribe(a => this.panda = a);```

### Observed behavior

When ```this.panda``` is changed, change detection does not happen in production environment and UI is not updated (*ngIf="panda").

### Desired behavior

Change detection should happen.

It works as expected in 3.0.1 - all environments, 3.1.1/3.1.2 - development environment.
There is NO ChangeDetectionStrategy.OnPush set on component.

### LINKED ISSUE #455: Action gets logged after ActionSuccess in Redux DevTools
### Versions

```
* ngxs: 3.1.4
* @angular/core: 6.0.0
```

### Repro steps

* Dispatch an async action (e.g. named Action) in the component.
```
@Component({
  templateUrl: 'dashboard.component.html'
})
export class DashboardComponent {

  constructor(private store: Store) {
    this.store.dispatch(new FetchAllSkills);
  }

}
```
* Dispatch ActionSuccess from `@Action`.
```
    @Action(FetchAllSkills, { cancelUncompleted: true })
    fetchAll(ctx: StateContext<ISkillState>) {
        return this.backendService.getSkills().pipe(
            map(skills => ctx.dispatch(new FetchAllSkillsSuccess(skills))),
            catchError(error => ctx.dispatch(new FetchAllSkillsError(error))),
            takeUntil(this.actions$.pipe(ofAction(CancelFetchSkills)))
        );
    }
```
### Observed behavior

In Redux DevTools The ActionSuccess appears before Action. It seems that Action waits for ActionSuccess to be done (confirmed by using the delay operator) and then gets outputted:
screen shot 2018-06-20 at 21 08 15

### Desired behavior

The expectation is that Action logs before ActionSuccess, at the actual time when it is dispatched. I went through the documentation and I don't think I misused something. What could've gone wrong here?

### PR DESCRIPTION
**Pull request by markwhitfeld**
fixes #415, fixes #455

### DIFF
--- a/packages/websocket-plugin/src/websocket-handler.ts
+++ b/packages/websocket-plugin/src/websocket-handler.ts
@@ -1,7 +1,6 @@
 import { Injectable, Inject } from '@angular/core';
 import { WebSocketSubject } from './websocket-subject';
-import { Store, getValue, ofActionDispatched } from '@ngxs/store';
-import { InternalActions } from '@ngxs/store/src/actions-stream';
+import { Actions, Store, getValue, ofActionDispatched } from '@ngxs/store';
 import {
   ConnectWebSocket,
   DisconnectWebSocket,
@@ -15,7 +14,7 @@ import {
 export class WebSocketHandler {
   constructor(
     store: Store,
-    actions: InternalActions,
+    actions: Actions,
     socket: WebSocketSubject,
     @Inject(NGXS_WEBSOCKET_OPTIONS) config: NgxsWebsocketPluginOptions
   ) {
--- a/packages/websocket-plugin/src/websocket-handler.ts
+++ b/packages/websocket-plugin/src/websocket-handler.ts
@@ -1,6 +1,7 @@
 import { Injectable, Inject } from '@angular/core';
 import { WebSocketSubject } from './websocket-subject';
-import { Actions, Store, getValue, ofActionDispatched } from '@ngxs/store';
+import { Store, getValue, ofActionDispatched } from '@ngxs/store';
+import { InternalActions } from '@ngxs/store/src/actions-stream';
 import {
   ConnectWebSocket,
   DisconnectWebSocket,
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: packages/websocket-plugin/src/websocket-handler.ts [hunk 1]
```diff
@@ -1,6 +1,7 @@
 import { Injectable, Inject } from '@angular/core';
 import { WebSocketSubject } from './websocket-subject';
-import { Actions, Store, getValue, ofActionDispatched } from '@ngxs/store';
+import { Store, getValue, ofActionDispatched } from '@ngxs/store';
+import { InternalActions } from '@ngxs/store/src/actions-stream';
```
markwhitfeld: @amcdnl This import is a bit strange because it is accessing an internal class. 
I did not want to expose the InternalActions as part of the public api.
As far as I can see, this seems to work. the packager ran fine and the tests all passed.
I don't have a sample websockets project to test with. Could you test one once merged?

amcdnl: Why do we need to use internal actions here? Seems like the normal actions would be fine.

markwhitfeld: Because the other changes in this PR makes normal actions run back in the angular zone and I wouldn't want to assume that the websocket stuff should be run in the angular zone again.

amcdnl: I think for these small things it would be fine.

markwhitfeld: Ok. I'll revert the changes to the websockets part

markwhitfeld:
Should be good to go now.

PR #484, opened 2018-07-06
feat: extract createSelector function for public api

markwhitfeld:
Implements solution proposed in #426 to expose the createSelector function.
Provides a solution for #413.


deebloo:
I think  this is a good idea

markwhitfeld:
The integration tests failed in a strange way (looks like a server error):
https://circleci.com/gh/ngxs/store/3541

Rewrote commit and pushed again to get a rebuild going and got the same error:
https://circleci.com/gh/ngxs/store/3544

It works locally so I'm not sure what the issue is here. Looks like missing pollyfills or something

markwhitfeld:
Added more tests as part of this and now we have broken the 100 tests mark!

deebloo:
@markwhitfeld this is actually a bit different then I was thinking. still poking at it

markwhitfeld:
@deebloo what were you thinking? I'm interested in what other use cases for a createSelector that there may be

deebloo:
I still don't like pulling one state class into another. but I also don't see a way around it if we want to do something like this

amcdnl:
@markwhitfeld - Can u PR docs for this?

juliusstoerrle:
Hi! When would I choose this over the earlier introduced dynmic selector pattern (`@Selector()
static foo(state) {
   return (arg) => {};
}`)?

Main benefit I see is the use in `@Select()` and the reduced boilerplate which was introduced by the `.pipe(map(...))`? Art there any downsides to using this new method?

markwhitfeld:
@juliusstoerrle See the docs added in PR #511 
It clarifies the difference between lazy and dynamic selectors (the existing type being called 'lazy' and this new type 'dynamic').

PR #485, opened 2018-07-07
feat: entity base with NgxsOnInit

juracy:
Use NgxsOnInit to avoid inherit from StateBase

I believe with this approach, there is some advantages:
- Simple classes
- Maybe StateBase isn't neccessary anymore, so it's possible to use it without Store modifications


markwhitfeld:
@leon I think that this is a better approach than attaching the state context to the metadata. This looks like it is along the lines of what I suggested here: https://github.com/ngxs/store/pull/321#r29447615

PR #489, opened 2018-07-11
fix(entity-base): cleanup changes in the store

juracy:
- After #485 those changes became unnecessary


markwhitfeld:
LGTM. It is important that the state context is unique for each call so that the debug information can be collected reliably for our dev tool.

Issue #416, opened 2018-06-04
Unable to import state.json with NgXS devtools


LmKupke:
### Versions

```
* ngxs: 3.1.2
* @angular/core: 6.0.3
```

### Repro steps
Currently using the demo provided on the NgXS site. 
[ToDo NgXS Repo](https://github.com/mailok/todo-ngxs)
[ToDo NgXS Demo](https://mailok.github.io/)

Run a set of actions on the demo and open Redux Dev Tools.
Download all the actions as a JSON file. Then refresh the page and clear the state. 
Importing the state.json file to the redux dev tools.

### Observed behavior

 Unable to see the app fire the same actions recorded earlier and imported by the state.json file when using the slider.

### Desired behavior
When uploading the state.json to the Redux Devtools, you should be able to time travel the app.

I would like to be able to see importing JSON state as an option in the NgXSDevTools options.

PR #507, opened 2018-07-18
fix(devtools-plugin): implement state.json import

sroettering:
Importing an exported state json file was not implemented yet (see #416).

When importing a file in the redux devtools an action of type 'IMPORT_STATE' is dispatched which is now handled by the devtools-plugin with these steps:
* reset the devtools extension to the first state found in the imported data
* replay every action but the first (this is done during reset)
* reset the ngxs store to the current state with regards to the exported jump position


amcdnl:
Thanks!

LmKupke:
Is this in version 3.2?

markwhitfeld:
@LmKupke It is in the current `@dev` npm release. To be released with v3.3.0

PR #510, opened 2018-07-19
feat(select): integrated memorized select decorator

NarHakobyan:
to be able to add params to Select for memorized selection

### DIFF
--- a/packages/store/src/decorators/select.ts
+++ b/packages/store/src/decorators/select.ts
@@ -20,60 +20,7 @@ export class SelectFactory {
 /**
  * Decorator for selecting a slice of state from the store.
  */
-export function Select(selectorOrFeature?, ...paths: string[]) {
-  return function(target: any, name: string) {
-    const selectorFnName = '__' + name + '__selector';
-
-    if (!selectorOrFeature) {
-      // if foo$ => make it just foo
-      selectorOrFeature = name.lastIndexOf('$') === name.length - 1 ? name.substring(0, name.length - 1) : name;
-    }
-
-    const createSelect = fn => {
-      const store = SelectFactory.store;
-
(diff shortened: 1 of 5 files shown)


markwhitfeld:
This is very similar to what I suggested here:
https://github.com/ngxs/store/issues/413#issuecomment-394468056
I would say that the parameters would need to be an array to allow for more than one argument.

@amcdnl What do you think of this?

NarHakobyan:
@markwhitfeld here you can do something like this 
`@Select(UsersState.list, 'clients', 'developers')`

markwhitfeld:
@NarHakobyan That would limit the opportunity for subsequent @Select argument types.

NarHakobyan:
@markwhitfeld can you please send what do you mean?

markwhitfeld:
@NarHakobyan This is what I mean:
```@Select(UsersState.list, ['clients', 'developers'] )```
This would mean that the second parameter would always be treated as an argument source. If we wanted to add any other options in future then they would be from the 3rd parameter onwards. 

Ps. Have you seen the new `createSelector` feature? Read the page in the digs on select for an explanation of the difference.

**Comment by amcdnl** [comment]
Bump.

File: packages/store/src/decorators/select.ts [hunk 1]
```diff
@@ -19,7 +20,7 @@ export class SelectFactory {
 /**
  * Decorator for selecting a slice of state from the store.
  */
-export function Select(selectorOrFeature?, ...paths: string[]) {
+export function Select(selectorOrFeature?: any | string[], ...args: any[]) {
```
markwhitfeld: This change would unfortunately be a breaking change for the framework. This is not an option.

PR #514, opened 2018-07-20
fix(devtools-plugin): support "relaxed" json format

eranshmil:
When copying a raw state from the devtools plugin, the JSON is in "relaxed" format and JSON.parse isn't supporting that, for example:
```
{
  payload: 'Make pizza',
  type: 'AddTodo'
}
```

So in this PR I did the following:
1. Added [JSON5](https://github.com/json5/json5) package dependency and changed JSON.parse to json5.parse. I also changed the JSON.parse in importing state for consistency, even though it's not required because the state.json is stringified.
2. After updating the package-lock file, the ngxs version has also been updated, the dependencies automatically reordered and I found out there is an unnecessary package (@types/googlemaps).
3. I also refactored a little bit the dispatched method in the plugin to make it more readable.

All the features have been tested using Chrome.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "ngxs",
-  "version": "3.1.4",
+  "version": "3.2.0",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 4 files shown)


Thenis:
I think that the json5 library should be wrapped in a service provider and should be injected into the devtools plugin using Angular's DI because at the moment json5 is coupling the devtools plugin

amcdnl:
I don't want to add anymore dependencies to the project. I think there is a way to make it where you can parse it yourself. I don't remember off the top of my head. If not, then we should allow that rather than add this.

eranshmil:
The other way is to use eval, just like implemented [here](https://github.com/ngrx/platform/blob/55a0488c7e456d74affa2354a12cc20db12e5f38/modules/store-devtools/src/extension.ts#L185).
I also tested that with eval wand it's working, but I wanted to avoid that.

PR #527, opened 2018-07-27
docs(install): add yarn to instructions

eranshmil:
Add yarn installation commands to main installation page and all the plugins.


eranshmil:
Added the save flag back, also to commands that missed that flag.

Issue #543, opened 2018-08-14
[Feature Request] Remove action class concept


kyusupov33:
## Feature Request?
Is it possible to get rid of the concept with the action in a separate class and set the type and the payload directly in the decorator?

Something like this

```
@Action(type: string, (payload: T) => payload)
```

splincode:
```ts
@Action({ type: '[Action type] description', action: (payload) => payload })
```

But if you set 

```ts
@Action({ action: (payload) => payload })
```

type will be auto generated by current state class

markwhitfeld:
Hmmm... This is a very interesting idea. Did you get this idea from anywhere?
@amcdnl I would be interested to get your take on this idea (from the comment, not the issue description) with the @Dispatcher and @Dispatch

splincode:
https://github.com/ngxs-labs/emitter

PR #575, opened 2018-09-22
feat(store): add dispatcher decorators

splincode:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: https://github.com/ngxs/store/issues/543

counter.actions.ts

```ts
export class Set {
  static readonly type = 'Set';
  constuctor(payload: number) {}
}
```

counter.state.ts

```ts
import { State, Action } from '@ngxs/store';

export interface CounterStateModel {
  value: number;
}

@State<CounterStateModel>({
  name: 'counterState',
  defaults: { value: 0 }
})
export class CounterState {

  @Action(Set)
  public updateCounterState({ setState }, { payload }: Set<number>) {
    setState({ value: payload });
  }

}
```

app.component.ts

```ts
... 

@Component({ 
 selector: 'app-root'
 template: `
    <ng-container *ngIf="count$ | async as count">
        <h1>Count is {{count.value}}</h1>
        <button (click)="set(count.value + 1)">increment</button>
        <button (click)="set(count.value - 1)">decrement</button>
     </ng-container>
 ` 
})
export class AppComponent {
 
 @Select(CountState) count$: Observable<CounterStateModel>;

 public set(value: number) {
    this.store.dispatch(new Set(value));
 }

}
```

Potential problems:
1. The name of the method is not used anywhere (updateCounterState) and can be named as you like
2. If I click on the Set class in the IDE, I'll go to the file I do not need, which does not give me any useful information
3. If we have a lot of business logic, then to interact with the store, I have to create a new action every time, which does not increase the development speed. In simple cases, this is not convenient.
4. If you create 2 actions and assign them the same type equally (accidentally or copied), it is impossible to understand why the data store is not updated:

counter.actions.ts

```ts
export class Add {
  static readonly type = 'DUPLICATE';
}

export class Delete {
  static readonly type = 'DUPLICATE';
}
```

## What is the new behavior?

Now no longer need a boilerplate to create action. Now you can just use the dispatcher directly.

+ Fixed bugs for methods types
+ Added the ability to freeze only on reads for objects (action)
+ Added tests

## Dispatcher Decorators

counter.state.ts

```TS
import { State, Dispatcher, DispatchAction } from '@ngxs/store';

export interface CounterStateModel {
  value: number;
}

@State<CounterStateModel>({
  name: 'counterState',
  defaults: { value: 0 }
})
export class CounterState {

  @Dispatcher()
  public static setValue({ setState }, { payload }: DispatchAction<number>) {
    setState({ value: payload });
  }

}
```

app.component.ts

```ts
import { Select, Dispatch, DispatchEmitter } from '@ngxs/store';
...

@Component({ 
 selector: 'app-root'
 template: `
    <ng-container *ngIf="count$ | async as count">
        <h1>Count is {{count.value}}</h1>
        <button (click)="counterValue.emit(count.value + 1)">increment</button>
        <button (click)="counterValue.emit(count.value - 1)">decrement</button>
     </ng-container>
 ` 
})
export class AppComponent {
 
  // Reads the name of the state from the state class
  @Select(CounterState) count$: Observable<CounterStateModel>;
  
  // Use in components to emit asynchronously payload
  @Dispatch(CounterState.setValue) counterValue: DispatchEmitter<number>;
 
}
```

Potential improvements:
1. It is no longer necessary to create an action for simple cases and call updates directly from the template.
2. Improved logging, now you are clearly oriented and you can understand what action (anonymous) is performing the update of the data warehouse:

image

3. Improved typing and simplification of testing

```ts
export interface AppState {
   counterState: CounterStateModel;
}

TestBed.configureTestingModule({
      imports: [ NgxsModule.forRoot([ CounterState ]) ]
});

const store: Store = TestBed.get(Store);

store.action<number, AppState>(CounterState.setValue)
   .emit(10)
   .subscribe(({ counterState }) => {
      expect(counterState.value).toEqual(10);
   });
```

4. Backward compatibility

```ts
@State<CounterStateModel>({
  name: 'counterState',
  defaults: { value: 0 }
})
export class CounterState {

  @Dispatcher({ type: '[Counter API] ChangeCount' }) // <--- custom action type (name)
  public static setValue({ setState }, { payload }: DispatchAction<number>) {
    setState({ value: payload });
  }

}
```

```
[ ] Yes
[x] No
```

## Other information

Co-authored-by: Max Ivanov (splincode) <splincodewd@yandex.ru>
Co-authored-by: Artur Androsovych (arturovt) <arthurandrosovich@gmail.com>


docs/advanced/immutable-actions.md
```diff
@@ -0,0 +1,143 @@
...
+Suppose that you had a coworker modify the announce function above to highlight Elvis rather than Cash, but they mistyped the comparison.
+
+```TS
+@State<Artist>({ .. })
+export class ApplicationState {
+
+  @Action(ApplicationUpdate)
+  public set({ getState, setState }, { artist }: ArtistAction) {
+
+    if (artist.name = "Elvis Presley") {
```
thekiba: `===` instead of `=`

splincode: image

markwhitfeld:
@splincode @arturovt Are you planning to modify this PR?

splincode:
@markwhithfeld yes, today

thekiba:
I think the Dispatcher needs to be removed too. And to remake the signature of Dispatch, this should take Action. All this will make it possible to reuse Action and make the code cleaner.

Otherwise, there is a confusion between Dispatcher and Action.
Dispatch is also an incorrect thing - now it's not a dispatch Action.

markwhitfeld:
@splincode @arturovt 
PS. There were quite a few changes in this PR that are not entirely necessary like formatting or moving the component template into a separate file (for example). While this is useful and I see the merits in this it does add quite a bit of noise to the code being reviewed and detracts from the main purpose of the PR. I would recommend doing these sorts of tweaks in another `cleanup` sort of lightweight PR. Just a thought... Although I would never discount a PR for these reasons.

karma.conf.js
```diff
@@ -7,6 +7,7 @@ module.exports = function(config) {
     frameworks: ['jasmine', '@angular-devkit/build-angular'],
     plugins: [
       require('karma-jasmine'),
+      require('karma-spec-reporter'),
```
markwhitfeld: These karma changes are all unrelated to the PR

yarn.lock
```diff
@@ -169,42 +169,28 @@
...
     get-stdin "5.0.1"
     lodash.merge "4.6.1"
     lodash.pick "4.4.0"
-    meow "3.7.0"
+    meow "4.0.0"
 
 "@commitlint/config-conventional@^6.1.3":
   version "6.1.3"
-  resolved "https://registry.yarnpkg.com/@commitlint/config-conventional/-/config-conventional-6.1.3.tgz#6c06eeae04c5ac789c3618df4d52aeda89ffb810"
+  resolved "http://registry.npmjs.org/@commitlint/config-conventional/-/config-conventional-6.1.3.tgz#6c06eeae04c5ac789c3618df4d52aeda89ffb810"
```
markwhitfeld: https changed to http!?

thekiba:
@maartentibau 
I think the proposed solution is still complicated. I'm talking about this:
```typescript
@Dispatch() 
addTodo(todo: string) {
  return new AddTodo(todo);
}
```

What do you think about something like that? `DispatchType` can be either a function or an emitter.
```typescript
@Dispatch(AddTodo) addTodo: DispatchType<AddTodo>;
```

I also have a couple of words about ngxs. At the moment it is not usable for projects with a large code base. The main problem is that you can't create a new Store/State dynamically. I'm not talking about the state of the application, I'm talking about the state of the component.

This can be solved, if take out all the decorators from the components back to the State. And if you do get the Store/State from DI. And also remove Injector in the decorators.

I understand that these are very critical changes, but if you develop a solution for large projects too, it will be cool.

juliusstoerrle:
The problems @splincode mentioned when creating this PR definitely occured to me aswell and it would be great to solve them! One additional problem I have with action (types) is that you are supposed to name them by from where they come from, but define them where they go to (in the state library).  (https://ngxs.gitbook.io/ngxs/concepts/actions#how-should-you-name-your-actions)

The @Dispatch() Operator from #447 still needs an Action, therefore I (still) agree with the comments made there, that does not add value (for me). Therefore I also dont agree with @thekiba in https://github.com/ngxs/store/pull/575#issuecomment-425655721 as I understand this PR targets a way to "remove" the action class concept.

I do think that this PR can help with reducing boilerplate in client code (which might be worth the "bloating" of ngxs) and improve DX by reducing the code nessesary and improving IDE support.


@splincode you mentioned:
> It is no longer necessary to create an action for simple cases and call updates directly from the template.

I would like to know what cases are too complicated for this concept? I think we should try to eliminate any limitations, so that users can opt to use only one concept in their project.

**Error Handling**
While I'm not sure whats the best way to do error handling yet, right now I subscribe to the return of store.dispatch(). This would be something that slow down my adaption of such a concept, but probably there is a better way I should do error handling..? (Maybe we can collect some tipps/best practices for error handling, somewhere external to this PR)

**Naming**
I feel like the names `@Dispatcher` and `DispatchAction` are confusing. (`@Dispatch()` is fine).
In my definition a dispatcher gets an event/action and distributes it to some handler. So mabye something along the lines of `@Handler()`.
Haven't come up with an alternative for `DispatchAction` yet.

**External Package?**
If we decide that this is usefull, but maybe not for most people, before opting to move this to a different package, we should check if there is an increase in footprint after tree shaking, for people not using any of the new functions. (Keeps down maintenance if you have to install/update less packages). 

**DevTools**
I did not view the code yet, but how (under what name?) would actions of type `DispatchAction` be logged in the ReduxDevTools? "CounterState.setValue" like in the console?


I think we should get some more feedback from different users and discuss a little on the above mentioned topics, but If we can get the documentation to clearly seperate both concepts and also list some advantages/disadvantages for each, this could be a nice addition to this package.

juliusstoerrle:
Another feature I'm waiting for is some entity state (#321, #580,..) concept to reduce boilerplate for CRUD operations, it might be worth waiting for that or creating a solution for entity state, that works for this PRs concept aswell.

splincode:
@juliusstoerrle @markwhitfeld What do you think about this approach, we will transfer the functionality to a separate plugin?

https://github.com/ngxs-contrib/dispatcher/pull/1/files

```ts
@NgModule({
  imports: [
       NgxsModule.forRoot([TodosState]),
       NgxsEmitPluginModule.forRoot()
  ],
  provide: [
       ApiService
  ]
})
export class AppModule {}
```

todo.state.ts

```ts
type Todo = string;

@State<Todo[]>({
      name: 'todos',
      defaults: []
 })
 class TodosState {
     
    public static api: ApiService;
        
    constructor(injector: Injector) {
        TodosState.api = injector.get(ApiService);
    }

    @Emitter() // Anonymous action - EmitAction
    public static addTodo({ setState, getState }, { payload }: EmitAction) {
        setState([  ...getState(), payload ]);
    }

    @Emitter({ action: TodoAction }) // Usage custom action 
    public static addTodo({ setState, getState }, { payload }: TodoAction) {
        setState([  ...getState(), payload ]);
    }

    @Emitter({ type: '[Todos] Set todos sync' }) // custom type for devtools
    public static async setTodosSync({ setState }: StateContext<Todo[]>) {
        setState( await this.api.getTodosFromServer(10).toPromise() );
    }

    @Emitter({ type: '[Todos] Set todos async' }) // custom type for devtools
    public static setTodos({ setState }: StateContext<Todo[]>) {
        return this.api.getTodosFromServer(5)
        .pipe( take(1), tap(todos => setState(todos)) );
    }

}
```

```ts
@Component({ 
   selector: 'app-root',
   template: `
     <input #input (blur)='addTodoAction.emit(input.data)'>
   `
})
class AppComponent {

      @PayloadEmitter(TodosState.addTodo)
      public addTodoAction: Emittable<Todo>;

      @PayloadEmitter(TodosState.setTodos)
      public downloadTodo: Emittable<Todo>;

      public loadData() {
           this.downloadTodo.emit().subscribe(
              (data) => console.log(data),
              (error) => this.errorHandler(error)
           );
      }

     private errorHandler(error) {
      // ...
     }
   
}
```

thekiba:
@splincode 
This is a dirty hack, because of the static variable is used to get an instance of API service. We lose the possibility of the decomposition. And also it takes away us from the concept of Action.

I like one of those proposals:
https://github.com/ngxs/store/pull/575#issuecomment-425672679

splincode:
@thekiba I do not think so, because:

1. We can use actions in our implementation
2. The Angular documentation says that we are allowed to use static access to services
3. I do not want to use actions in my projects, since they do not carry any benefit, but only deliver boilerplate (extra classes, extra files). Our plugin will get rid of this, but also it will be possible if you want to use actions.

thekiba:
> If you need dynamic created in runtime, you can use other concepts or
> 
> ```ts
> NgxsModule.forFeature([ DynamicState ])
> ```

This is not a runtime dynamic store.

As I said, the main mistake of such a decision is the impossibility of decomposition. This makes it impossible to use the configuration for your application. And it requires creating an extra boilerplate.

As an example, you cannot reuse written code, and you have to write them again and again, or generate them using schematics.

maktarsis:
@splincode well done

markwhitfeld:
PS. for reference this has become a ngxs labs library found here:
https://github.com/ngxs-labs/emitter

Issue #594, opened 2018-10-05
Rehydrate store from storage without losing initial values (storage-plugin)


viters:
I would like to persist only a part of my state, but in current implementation of storage plugin it is impossible. For example: I have AuthState, where I hold accessToken, loading and error properties. I do not want to persist loading nor error, only accessToken.
Now if I write serializer that saves accessToken, and deserialize it later - I will lose initial loading and error properties.

Moreover, I can define keys of states I want to persist (instead of @@STATE) - but I cannot separate keys in storage from state names (fe. key state-auth would correspond to auth).

Moreover, if I can define keys (meaning I can serialize part of my state) - I should be able to define serialize/deserialize strategies - but it's impossible now because I do not know which key is serialize/deserialize processing.

Possible enhancement:
`key` array should become `keys` map, in which key is "state name" and value is "in storage key". For example: if I have AuthState named "auth" and want to keep in localStorage as "store-auth" it would be
`{ auth: 'store-auth' }`. Because it is a map, there would be only one strategy possible per state.

In serialize function, I should receive two things, instead of one: `(state: any, stateName: string)`. Thanks to this I can write different serialize strategies based on state I am currently processing.

In deserialize function, I should receive three things, instead of one:
`(serializedVal: string, stateName: string, state: any)`. Thanks to this, I can base my deserialize strategy on stateName and keep my initial values from previous state value.

viters:
A good starting point is to pass more data to serialize / deserialize. I can make a PR at the weekend.

PR #597, opened 2018-10-07
feat(storage-plugin): key and previous state in (de)serialize (#594)

viters:
## PR Checklist

## PR Type

```
[ ] Bugfix
[X] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: #594 

## What is the new behavior?

Serialize and deserialize functions will have more args passed. 
Serialize is now `(val, key) => serialized` and deserialize `(val, key, state) => deserialized`.

```
[X] Yes
[ ] No
```

If someone used a function that can take a different number of arguments for serialize and deserialize without passing them explicitly (f.e. `serialize: JSON.parse`) then it should be now used like this: `serialize: (v) => JSON.parse(v)`. It's because `serialize` is no more single-argument and it can cause unexpected behavior. 

## Other information


packages/storage-plugin/tests/storage.plugin.spec.ts
```diff
@@ -259,4 +259,60 @@ describe('NgxsStoragePlugin', () => {
...
+
+  it('should allow to use different strategies based on key', () => {
+    localStorage.setItem('counter', btoa(JSON.stringify({ count: 2137 })));
+
+    TestBed.configureTestingModule({
+      imports: [
+        NgxsModule.forRoot([MyStore]),
+        NgxsStoragePluginModule.forRoot({
+          key: 'counter',
+          serialize: (val, key) => {
```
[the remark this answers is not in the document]
viters: I see no benefit of doing this.

packages/storage-plugin/src/symbols.ts
```diff
@@ -46,12 +46,12 @@ export interface NgxsStoragePluginOptions {
   /**
    * Serailizer for the object before its pushed into the engine.
    */
-  serialize?(obj: any);
+  serialize?(obj: any, key: string);
```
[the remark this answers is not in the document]
viters: If I am correct, this doesn't introduce any improvement considering storage plugin and ngxs store implementation.

viters:
@markwhitfeld 
Well, they are optional.

Consider the following situation:
You have proposed such serialization config: `{ deserialize: JSON.parse }`. And it will work, as long as we use `deserialize` function as single parameter in storage-plugin implementation, like this:
```typescript
val = options.deserialize(val);
```
But in my case, I need to pass several more arguments:
```typescript
val = options.deserialize(val, key, previousState);
```
So, now we can run into a problem. `JSON.parse` or any other function that NGXS user has used without explicit arguments pass, might take more than one parameter. 

Actually, `JSON.parse` might take two parameters, with function on the second one. Here we are in luck, because string as second argument doesn't break `JSON.parse`:
```typescript
JSON.parse('{"x": 1}', 'someKey', {x: 2});
// returns {x: 1}
```

Now consider a function `fn`:
```typescript
function fn(arg1, arg2) {
  if (arg2) {
    arg2();
  }
}

fn(1, 'someKey');
// Uncaught TypeError: arg2 is not a function
```
So, given that some NGXS storage-plugin user had `{deserialize: fn}` - is my pull request a breaking change?

packages/store/src/internal/state-operations.ts
```diff
@@ -12,12 +12,7 @@ import { deepFreeze } from '../utils/freeze';
  */
 @Injectable()
 export class InternalStateOperations {
-
-  constructor(
-    private _stateStream: StateStream,
-    private _dispatcher: InternalDispatcher,
-    private _config: NgxsConfig
-  ) {
+  constructor(private _stateStream: StateStream, private _dispatcher: InternalDispatcher, private _config: NgxsConfig) {
```
markwhitfeld: Please revert this change. Seems like your formatter got too eager.

packages/storage-plugin/src/storage.plugin.ts
```diff
@@ -25,7 +25,8 @@ export class NgxsStoragePlugin implements NgxsPlugin {
 
         if (typeof val !== 'undefined' && val !== null) {
           try {
-            val = options.deserialize(val);
+            const previousState = isMaster ? state : getValue(state, key);
+            val = options.deserialize(val, key, previousState);
```
markwhitfeld: Is the previous state still necesary as a parameter after the other changes below that merge the state (line 45)?

viters: You got me there, I did not see these changes. I think passing previousState is still necessary. Furthermore, I do not agree with line 49, because it's behavior is not something I would expect while in 'redux' mindset. Although I think my solution and solution from line 49 doesn't collide.

packages/store/src/internal/state-operations.ts
viters: It's pretty strange considering I have no globally installed libraries and this change was made automatically before commit (commited via tty).

yarn.lock.readme.md
```diff
@@ -1,25 +0,0 @@
-All of our npm dependencies are locked via the `yarn.lock` file for the following reasons:
```
splincode: Why remove this file?

packages/storage-plugin/src/symbols.ts
```diff
@@ -46,12 +46,12 @@ export interface NgxsStoragePluginOptions {
...
 
   /**
    * Deserializer for the object before its pulled out of the engine.
    */
-  deserialize?(obj: any);
+  deserialize?(obj: any, key: string, state: any);
```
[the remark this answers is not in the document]
viters: Sadly, I also had to change `key?: string` to `key?: any` to cope with backwards compatibility of `deserialize: JSON.parse`

packages/storage-plugin/src/symbols.ts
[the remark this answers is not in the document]
viters: Sadly, I also had to change `key?: string` to `key?: any` to cope with backwards compatibility of `serialize: JSON.stringify`

packages/store/src/internal/state-operations.ts
viters: When I run prettier in your project on this file, the output is:
`constructor(private _stateStream: StateStream, private _dispatcher: InternalDispatcher, private _config: NgxsConfig) {`
This file is changed automatically when husky runs prettier on whole project before commit.

I can try to force-change it via github to omit husky, but should I do it?

packages/storage-plugin/src/symbols.ts
```diff
@@ -44,14 +44,21 @@ export interface NgxsStoragePluginOptions {
...
 
   /**
    * Deserializer for the object before its pulled out of the engine.
+   *
+   * @param obj {any} value of currently processed state slice
+   * @param key {string} key of currently processed state slice
+   * @param state {any} previous value of currently processed state slice
    */
-  deserialize?(obj: any);
+  deserialize?(obj: any, key?: any, state?: any);
```
markwhitfeld: Unfortunately switching the key to any is a bit of a hack to avoid the breaking change. I would prefer that it was the `string` type. I really need to understand the need for this feature though...

viters:
@markwhitfeld 
Someone wants to serialize an object which is instance of some class. In deserialization process, we need to 'revive' this object, so it's instance of class again.

I have this problem, because I use a lot of [Option from fp-ts](https://github.com/gcanti/fp-ts/blob/master/src/Option.ts) which is implemented with classes. Because of that, I need to base my serialize/deserialize on some metadata and it can lead to problems in future:
```typescript
import { none, some } from 'fp-ts/lib/Option';

export function deserialize(val: string): any {
  const obj = JSON.parse(val);

  if (obj.hasOwnProperty('_tag') && obj._tag === 'Some') {
    return some(obj.value);
  } else if (obj.hasOwnProperty('_tag') && obj._tag === 'None') {
    return none;
  }

  return obj;
}
```

I would rather keep track of which state slices are deserialized with certain functions using this slices keys, than rely on potentially unstable object insides.

Another situation, let's say I have state with 5 values that I need to store in localstorage. Let's call it:
`MyState { v1, v2, v3, v4, v5, v6 }`. And all of them have initial state. I want to store first five. 

To keep initial state of 6, now I should config storage-plugin like this: `keys: ['mystate.v1', 'mystate.v2', 'mystate.v3', 'mystate.v4', 'mystate.v5']`. Using strategy + previous state, I could serialize whole `MyState`, omitting `v6` and merge initial state with stored values later in deserialize.

splincode:
@markwhitfeld ping

splincode:
@viters is outdated?

splincode:
@markwhitfeld @viters current PR is relevant?

splincode:
The current PR does not develop, you can rediscover it in the future.

PR #628, opened 2018-10-26
fix(storage-plugin): lazy loaded slices not initialized

enewbury:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

When there is nothing in the store and you are using NgxsPluginModule.forFeature, additional slices that should be initialized in @@UpdateState by the feature are overwritten by the slices in the global state that were saved to storage on the first pass of @@InitState

Issue Number: N/A

## What is the new behavior?
Any slices from a feature that are undefined in the current state are added with their initial value.

```
[ ] Yes
[x] No
```

### DIFF
--- a/packages/storage-plugin/src/storage.plugin.ts
+++ b/packages/storage-plugin/src/storage.plugin.ts
@@ -45,7 +45,7 @@ export class NgxsStoragePlugin implements NgxsPlugin {
           if (!isMaster) {
             state = setValue(state, key, val);
           } else {
-            state = Object.assign(state, val);
+            state = { ...state, ...val };
           }
         }
       }
--- a/packages/storage-plugin/src/storage.plugin.ts
+++ b/packages/storage-plugin/src/storage.plugin.ts
@@ -45,7 +45,7 @@ export class NgxsStoragePlugin implements NgxsPlugin {
           if (!isMaster) {
             state = setValue(state, key, val);
           } else {
-            state = val;
+            state = Object.assign(state, val);
           }
         }
       }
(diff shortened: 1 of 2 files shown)


enewbury:
I am only able to reproduce when using angular router to lazy load new modules. The test files aren't setup for that that more elaborate case, but I can give it a shot if it's absolutely necessary.

splincode:
@enewbury It would be great to write a test for this case.

enewbury:
Test added and confirmed failing before, and not failing after.  Didn't require any big changes to testing file after all.  Cheers guys.

markwhitfeld:
@enewbury Ah man! It looks like it is partially fixed.
It is fixed for the first lazy route that you access but not subsequent ones.
See demo here: https://stackblitz.com/edit/ngxs-localstorage-lazyload-fixed 
Navigate to Lazy route 1 and it is working but then you go to lazy route 2 and there are errors.
Reset your local storage and then try the other way: Navigate to lazy route 2 first and then to lazy route 1 and you will see that lazy route 1 is now broken.
I have already closed and merged this PR so you will have to make a new PR with the changes.
I'm not going to revert the currently merged changes because it is better than before ;-)

enewbury:
@markwhitfeld for some reason, none of the modules are loading their defaults onInit, including the App module.  The reason it's the 2nd module is because the app module does not save its default state, so when you load the first module, it has no existing state to trip on.  If you click the "add" button in the app module, then it adds state to the appstate, then it breaks on the first lazy load module.
Also I just realized this is using storage-plugin version 3.0.0-rc.2 which doesn't have the fix.

enewbury:
confirmed it is working on my project with multiple lazy loaded modules using storage-plugin version 3.2.0-dev.master-e7d5450

Issue #629, opened 2018-10-27
[Feature]: Tree Shaking States (providedIn for State)


splincode:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[ ] Performance issue
[x] Feature request
[ ] Documentation issue or request
[ ] Support request => https://github.com/ngxs/store/blob/master/CONTRIBUTING.md
[ ] Other... Please describe:
</code></pre>

## Current behavior

```ts
// todo.state.ts

@State<string[]>({
  name: 'todo',
  defaults: []
})
export class TodoState {
 ...
}

@State<TodoStateModel>({
  name: 'todos',
  defaults: {
    todo: [],
    pizza: { model: undefined }
  },
  children: [TodoState]
})
export class TodosState {
  ...
}


// app.module.ts

@NgModule({
  imports: [ NgxsModule.forRoot([TodosState, TodoState]) ]
})
export class AppModule {}
```

## Expected behavior

```ts
// todo.state.ts

@State<string[]>({
  name: 'todo',
  defaults: []
})
export class TodoState {
 ...
}

@State<TodoStateModel>({
  name: 'todos',
  defaults: {
    todo: [],
    pizza: { model: undefined }
  },
  children: [TodoState], // default -> providedIn: 'ngxsRoot' for TodoState
  providedIn: 'ngxsRoot'
})
export class TodosState {
  ...
}


// app.module.ts

@NgModule({
  imports: [ NgxsModule.forRoot() ]
})
export class AppModule {}
```

## What is the motivation / use case for changing the behavior?

![image](https://user-images.githubusercontent.com/12021443/47605274-1b2a7f00-da0d-11e8-9e04-231c5840ac4b.png)

```ts
/**
 * Options that can be provided to the store.
 */
export interface StoreOptions<T> {
  /**
   * Name of the state. Required.
   */
  name: string;

  /**
   * Default values for the state. If not provided, uses empty object.
   */
  defaults?: T;

  /**
   * Sub states for the given state.
   */
  children?: any[];

  /**
   * Define states in your root module
   * */
  providedIn?: Type<any> | 'ngxsRoot' | null;
}
```

PR #631, opened 2018-10-29
feat(store): add tree shaking states into integration example

splincode:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

https://github.com/ngxs/store/issues/629

## What is the new behavior?

```ts
// todo.state.ts

@State<string[]>({
  name: 'todo',
  defaults: []
})
export class TodoState {
 ...
}

@State<TodoStateModel>({
  name: 'todos',
  defaults: {
    todo: [],
    pizza: { model: undefined }
  },
  children: [TodoState], 
  provideIn: 'ngxsRoot'
})
export class TodosState {
  ...
}
```

```
@NgModule({
  imports: [ NgxsModule.forRoot() ]
})
export class AppModule {}
```

```
[ ] Yes
[x] No
```

## Other information

Update integration example with tree shaking states

image

### DIFF
--- a/packages/store/src/decorators/state.ts
+++ b/packages/store/src/decorators/state.ts
@@ -1,5 +1,7 @@
 import { ensureStoreMetadata } from '../internal/internals';
-import { StoreOptions, META_KEY } from '../symbols';
+import { META_KEY, StoreOptions } from '../symbols';
+import { NgxsStateProvidersModule } from '../internal/state-providers/state-providers.module';
+import { Type } from '@angular/core';
 
 const stateNameRegex = new RegExp('^[a-zA-Z0-9]+$');
 
@@ -30,6 +32,15 @@ export function State<T>(options: StoreOptions<T>) {
     meta.children = options.children;
     meta.defaults = options.defaults;
     meta.name = options.name;
+
+    if (options.provideIn) {
+      const provideIn: string | Type<unknown> = options.provideIn;
+      const type: Type<unknown> = typeof provideIn !== 'string' ? provideIn : null;
+      const children = options.children || [];
+      const states = [target, ...children];
+      states.forEach((state: Type<unknown>) => NgxsStateProvidersModule.provideInNgxsModule(state, type));
+      NgxsStateProvidersModule.defineStatesByProvideIn(provideIn, states);
+    }
 
     if (!options.name) {
       throw new Error(`States must register a 'name' property`);
--- a/packages/store/src/symbols.ts
+++ b/packages/store/src/symbols.ts
@@ -1,4 +1,4 @@
-import { InjectionToken } from '@angular/core';
+import { InjectionToken, Type } from '@angular/core';
 import { Observable } from 'rxjs';
 
 export const ROOT_STATE_TOKEN = new InjectionToken<any>('ROOT_STATE_TOKEN');
@@ -91,6 +91,11 @@ export interface StoreOptions<T> {
    * Sub states for the given state.
    */
   children?: any[];
+
+  /**
+   * Define states in your root or feature module
+   * */
+  provideIn?: 'ngxsRoot' | Type<unknown> | null;
 }
 
 /**
--- a/integration/app/app.component.scss
+++ b/integration/app/app.component.scss
@@ -57,3 +57,7 @@ button {
   margin: 0 auto;
   border-top: solid 1px #ddd;
 }
+
+.menu a {
+  margin-right: 10px;
+}
--- a/integration/app/app.component.ts
+++ b/integration/app/app.component.ts
@@ -3,7 +3,8 @@ import { Store, Select } from '@ngxs/store';
 import { Observable } from 'rxjs';
 import { FormBuilder, FormArray } from '@angular/forms';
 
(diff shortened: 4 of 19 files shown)


packages/store/src/symbols.ts
```diff
@@ -91,6 +91,11 @@ export interface StoreOptions<T> {
    * Sub states for the given state.
    */
   children?: any[];
+
+  /**
+   * Define states in your root or feature module
+   * */
+  provideIn?: 'ngxsRoot' | Type<unknown> | null;
```
arturovt: Should be `providedIn`

splincode: good catch

packages/store/src/decorators/state.ts
```diff
@@ -1,5 +1,7 @@
 import { ensureStoreMetadata } from '../internal/internals';
-import { StoreOptions, META_KEY } from '../symbols';
+import { META_KEY, StoreOptions } from '../symbols';
+import { NgxsStateProvidersModule } from '../internal/state-providers/state-providers.module';
+import { Type } from '@angular/core';
```
arturovt: Could you move this import to the top?

viters:
What are the benefits of this approach, instead of previous `imports: [ NgxsModule.forRoot([TodosState, TodoState]) ]`?

While I am really keen of angular tree-shakeable services, it seems impossible to me to 'tree-shake' a state. Tree-shaking process isn't only about adding `providedIn`, but about removing unnecessary code. And here is my wrinkle: how can you tell, if state is being used or not (keeping in mind **dispatching actions**).

Beside that, it's true that this approach complies with "new angular way" for services - but why it uses keyword `provideIn` while angular core uses `providedIn`?

arturovt:
@viters

There was a type in the `provideIn` keyword, already fixed.

You are totally right - you cannot tree shake any state, BUT only using native approach like `@Action`.

We've recently published a package that allows you to get rid off actions ([emitter](https://www.npmjs.com/package/@ngxs-labs/emitter)), thus tree shaking works great with this approach.

viters:
@arturovt 

Awesome. Thank you for clarifying.

packages/store/src/decorators/state.ts
```diff
@@ -31,6 +34,15 @@ export function State<T>(options: StoreOptions<T>) {
     meta.defaults = options.defaults;
     meta.name = options.name;
 
+    if (options.providedIn) {
+      const provideIn: string | Type<unknown> = options.providedIn;
```
arturovt: Change to `providedIn`

markwhitfeld:
@splincode Doesn't the lazy example above result in a circular reference?
DetailState --references--> DetailModule
DetailModule --references--> DetailComponent
DetailComponent --references--> DetailState

markwhitfeld:
I feel like this feature is trying to solve something that is not a problem.
The `providedIn` in angular is to provide tree shakeable services exposed from libraries so that when they are included in a client app by being injected into something they will be automatically registered with the root module.
Our state classes are a very different paradigm to services.
There is no clear benefit to NGXS from this feature.

splincode:
@markwhitfeld 

### I. Question
```Doesn't the lazy example above result in a circular reference?```

image

Answer: No. It works exactly the same as in Angular. Just injected into the module.

Analogue: 

```ts
@NgModule({
  imports: [NgxsModule.forFeature([ DetailState ])],
  declarations: [DetailComponent]
})
export class DetailModule {}
```

### II. Question

```I feel like this feature is trying to solve something that is not a problem.```

I write huge applications consisting of a set of modules, sometimes components or modules are deleted and we forget to remove the state from the module. Together with @kyusupov33, we determined that the big problem is the implicit initialization of states.

https://github.com/ngxs/schematics/blob/master/src/templates/starter-kit/store/store.config.ts

image

It turns out that when we have a huge number of states, we are uncomfortable working with it.

### III. Question

```There is no clear benefit to NGXS from this feature.```

Believe me, this feature is worthy to outshine even NGRX. 
 - The bundle will not get states that are not used in the root and child modules if they are not used.
 - Even if you use only the root module, you will simply be able to write ``` providedIn: 'ngxsRoot' ```. 

I am ready to write a huge number of tests for you to believe me.

### Motivation

If we usage Emitter plugin, we are improving for Tree Shaking, because since we use state class references

```ts
@State({  })
export class AppState {

    private static tagService;

    @Emitter(AppStatusState.success)
    public static success: Emittable<NormalizeTag>;

    @Emitter(AppStatusState.failure)
    public static failure: Emittable<string>;

    constructor(tagService) {
      AppState.tagService = tagService;
    }

    @Receiver()
    public static request(ctx: StateContext<TagsStateModel>, { payload }: EmitterAction<NormalizeTag>) {
        return this.tagService.addOne(payload.newTag).pipe(
            tap(tag => this.success.emit({ tag: normalizeTag(tag), parentName: payload.parentName  })),
            catchError(error => this.failure.emit(error.message))
        );
    }

}
```

splincode:
@markwhitfeld  Since I will not be able to get your approval, I will close this thread and later will conduct an experimental development.

https://github.com/ngxs-labs/tree-shaking-state

Issue #622, opened 2018-10-25
Add warning message when usage developmentMode


splincode:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[x] Bug report  
[ ] Performance issue
[ ] Feature request
[ ] Documentation issue or request
[ ] Support request => https://github.com/ngxs/store/blob/master/CONTRIBUTING.md
[ ] Other... Please describe:
</code></pre>

## Current behavior

if we usage developmentMode in production then we have a bad load

```ts
NgxsModule.forRoot(states, { developmentMode: true })
```

## Expected behavior

Add a warning

![image](https://user-images.githubusercontent.com/12021443/47502876-7aec2300-d871-11e8-845c-97542d6a11e2.png)

synec:
This pollutes our test log with message

`HeadlessChrome 71.0.3578 (Mac OS X 10.14.1) WARN LOG: 'RECOMMENDATION: Set developmentMode to true on the NgxsModule when Angular is running in development mode.`

I think the message should be removed again. It's useless and annoying.

splincode:
@synec enable developmentMode ?))

eranshmil:
This should be fixed after the merge of PR #706.

splincode:
this means that immunity test does not work for you

synec:
@eranshmil Thank you for the notice. This will solve my issue.

markwhitfeld:
> This pollutes our test log with message
> 
> `HeadlessChrome 71.0.3578 (Mac OS X 10.14.1) WARN LOG: 'RECOMMENDATION: Set developmentMode to true on the NgxsModule when Angular is running in development mode.`
> 
> I think the message should be removed again. It's useless and annoying.

Agreed. I'm going to release a special bug fix release for the 3.3 branch to sort it out.
The PR number is mentioned above.

markwhitfeld:
@synec This has been fixed in the v3.3.3 release that just went out.
Let me know if you have problems.

synec:
> @synec This has been fixed in the v3.3.3 release that just went out.
> Let me know if you have problems.

Works like a charm! Thank you!

PR #633, opened 2018-10-31
breaking(store): set new default developmentMode

splincode:
Many do not know about this property and mutate their states.

## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[x] Other... Please describe:
```

https://github.com/ngxs/store/issues/622

## What is the new behavior?

Many do not know about this property and mutate their states.

```
[x] Yes
[ ] No
```

### DIFF
--- a/packages/store/src/symbols.ts
+++ b/packages/store/src/symbols.ts
@@ -17,7 +17,7 @@ export class NgxsConfig {
   /**
    * Run in development mode. This will add additional debugging features:
    * - Object.freeze on the state and actions to guarantee immutability
-   * (default: false)
+   * (default: true)
    */
   developmentMode: boolean;
   compatibility: {
@@ -30,6 +30,7 @@ export class NgxsConfig {
   };
 
   constructor() {
+    this.developmentMode = true;
     this.compatibility = {
       strictContentSecurityPolicy: false
     };


splincode:
@markwhitfeld It is very strange that tests pass, because this is a breaking changes. I am ready to write a test for it.

packages/store/src/symbols.ts
```diff
@@ -30,6 +30,7 @@ export class NgxsConfig {
   };
 
   constructor() {
+    this.developmentMode = true;
```
markwhitfeld: I think that this should be determined by angular's dev mode instead of default to true.
See https://angular.io/api/core/isDevMode

Koslun: @markwhitfeld as long as there is some flag to unfreeze the store, this makes sense imo.

With the flag seemingly *only* freezing the store I would further float the idea of changing the name and message to `freeze` or something similar. The current name doesn't really necessarily inform the consumer what is actually taking place imo, though it does make some sense.

Would also be sure to very clearly label this breaking change very clearly in the changelog. In the future we should probably also add another entry in the docs about this. Currently only a short description in the API I believe.

Our experience is basically that I glanced this issue and decided to turn the flag on in anticipation of version 4 without really seeing the description in my editor or going out of my to search for what it meant. Only the next day did we find the `Type error`s pop in one part of our app, where we were apparently mutating a store property due to an incomplete migration to NGXS from no state management. As the error itself does not present itself as immediately pointing to this library I didn't connect it immediately, primarily on my radar because I knew what had been changed for it to break.

So as is intended I believe this will have the intended effect of waking some up to unwarranted mutations of the state.

splincode:
https://github.com/ngxs/store/pull/674

PR #634, opened 2018-11-01
feat(dispatcher): return value of action handler instend of full state

NarHakobyan:
## PR Checklist

## PR Type

Instead of returning full state which developer can get from the state return result of action handler, to be able to dispatch by the chain, for example, use the swtichMap or swtichMapTo to pass the result of action handler to next dispatch

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: N/A

```
[x] Yes
[ ] No
```

use store.selectSnapshot() to get full state instead of using the return value of dispatch

### DIFF
--- a/packages/store/src/actions-stream.ts
+++ b/packages/store/src/actions-stream.ts
@@ -17,6 +17,7 @@ export interface ActionContext {
   status: ActionStatus;
   action: any;
   error?: Error;
+  data?: any;
 }
 
 /**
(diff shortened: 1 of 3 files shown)


markwhitfeld:
Thank you for the idea, but we will leave it as it is for now.
You should not count on it returning a value (as stated in the docs) but haven't done that change because it may be a breaking change for anyone that expected the state value.
The reason for the intended void return is to prevent the use of this observable for the wrong thing.
For example your switchMap example doesn't make sense because the dispatch observable only returns a single value and then completes.

markwhitfeld:
Also note this explanation from the docs:

> The Observable that a dispatch returns has a void type, this is because there can be multiple states that listen to the same @Action, therefore it's not realistically possible to return the state from these actions since we don't know the form of them.

NarHakobyan:
@markwhitfeld  yes, but the observer is returning a single value in any case, and we can combine then to the single value like combineLatest

Issue #606, opened 2018-10-18
Docs Bug?  Return value from store.dispatch


mann-david:
Possible documentation bug?  According to https://ngxs.gitbook.io/ngxs/concepts/store#dispatching-actions, "The Observable that a dispatch returns has a void type".  If my brain isn't totally fried after a long day, it looks like it's returning the whole state instead of void.  Am I missing something?  Did this change recently?

### Versions

```
* ngxs: 3.2.0
* @angular/core: 6.1.10
```

### Repro steps

Stackblitz: https://stackblitz.com/edit/ngxs-simple-izggo9

1. Open the dev tools console
2. Click the Count button
3. The entire state is logged to the console

### Observed behavior
Whole state is returned as payload of Observable

### Expected behavior
According to https://ngxs.gitbook.io/ngxs/concepts/store#dispatching-actions, "The Observable that a dispatch returns has a void type".  I would expect `undefined` to be logged.

### Mention any other details that might be useful (optional)
IMHO, it would be better if the Observable did have a void type.  I can get state other ways if I need it, but this makes it hard to do `withLatestFrom` as the docs show and get whatever just changed.

NarHakobyan:
Hi @mann-david, Yes you are right, it's returning full state instead of void, but I think it will be more comfortable if dispatch subscription will return data from action handler, I have already created PR for that
https://github.com/ngxs/store/pull/634

markwhitfeld:
You should not count on it returning a value as stated correctly in the docs.
We haven't done that change because it may be a breaking change for anyone that expected the state value to be returned as it is.
The reason for the intended void return is to prevent the use of this observable for the wrong thing.
For example user's may be tempted to chain these observables or select state slices, thereby skipping the correct mechanisms for doing these things in the framework.

NarHakobyan:
@markwhitfeld yes, but we are missing return value of Action Handler, it's it's not allowing to do dispatches by chain

splincode:
return value depracated since 4.0.0
https://github.com/ngxs/store/issues/827

PR #635, opened 2018-11-02
Enable patch extensibility

markwhitfeld:
## PR Checklist

## PR Type

```
[ ] Bugfix
[X] Feature
[ ] Build related changes
[ ] CI related changes
[X] Documentation content changes
[ ] Other... Please describe:
```

Issue Number: N/A

## What is the new behavior?

Added the ability to call eht `patchState` function with a patching function which will be given the
existing state and should return the new state. The signature of this function would be:
```(currentState: ReadOnly<T>) => T```
All immutability concerns need to be honoured by this function. 
This change is a general purpose extensibility point addition to the framework.

For example you could use a library like [immer](https://github.com/mweststrate/immer) that can handle these concerns for you and provide a different way of expressing your immutable update through direct mutation of a draft object. Here is the example from above expressed in this way:
```TS
import produce from 'immer';

// in class ZooState ...
@Action(FeedZebra)
feedZebra(ctx: StateContext<ZooStateModel>, action: FeedZebra) {
  ctx.patchState(produce((draft) => {
    draft.zebraFood.push(action.zebraToFeed);
  }));  
}
```
Here the `produce` function from the `immer` library is called with just a single parameter so that it returns its' [curried form](https://github.com/mweststrate/immer#currying) that will take a value and return a new value with all the expressed changes applied.

This approach can also allow for the creation of well named helper functions that can be shared between handlers that require the same type of update. 
The above example could be refactored to this:
```TS
import produce from 'immer';

// in class ZooState ...
@Action(FeedZebra)
feedZebra(ctx: StateContext<ZooStateModel>, action: FeedZebra) {
  ctx.patchState(addToZebraFood(action.zebraToFeed));
}

// defined elsewhere
function addToZebraFood(itemToAdd) {
  return produce((draft) => {
    draft.zebraFood.push(itemToAdd);
  });
}
```

_Note that the use of immer is for demonstration purposes (it has not been added to the framework)._

```
[ ] Yes
[X] No
```


markwhitfeld:
@splincode could you review this to see if there are any obvious issues

packages/store/src/internal/state-context-factory.ts
```diff
@@ -17,44 +17,52 @@ export class StateContextFactory {
   /**
    * Create the state context
    */
-  createStateContext(metadata: MappedStore): StateContext<any> {
+  createStateContext<T>(metadata: MappedStore): StateContext<T> {
```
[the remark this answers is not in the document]
splincode: @markwhitfeld ping

packages/store/src/internal/state-context-factory.ts
```diff
@@ -17,44 +17,52 @@ export class StateContextFactory {
...
       },
-      dispatch(actions: any | any[]): Observable<any> {
+      dispatch(actions: any | any[]): Observable<void> {
         return root.dispatch(actions);
       }
     };
   }
 }
+
+function simplePatch<T>(val: Partial<T>): PatchFunction<T> {
```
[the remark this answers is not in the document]
splincode: @markwhitfeld ping

docs/concepts/state.md
```diff
@@ -150,6 +150,47 @@ feedZebra(ctx: StateContext<ZooStateModel>, action: FeedZebra) {
 }
 ```
 
+The `patchState` function can also be called with a patching function which will be given the
+existing state and should return the new state. 
+All immutability concerns need to be honoured by this function. 
+For example you could use a library like [immer](https://github.com/mweststrate/immer) that can 
+handle these concerns for you and provide a different way of expressing your immutable update 
+through direct mutation of a draft object. Here is the example from above expressed in this way:
+```TS
```
splincode: Add simple example, so that people understand why the operator's patch function appeared

Before

```ts
@Action(MyAction)
public addValue(ctx: StateContext, { payload }: MyAction) {
  ctx.setState({ ...ctx.getState(), value: payload  });
}
```

After

```ts
@Action(MyAction)
public addValue(ctx: StateContext, { payload }: MyAction) {
  ctx.patchState((state) => ({ ...state, value: payload }));
}
```

packages/store/src/internal/state-context-factory.ts
```diff
@@ -17,44 +17,52 @@ export class StateContextFactory {
...
+    const isPrimitive = typeof val !== 'object';
+    if (isArray) {
+      throw new Error('Patching arrays is not supported.');
+    }
+    if (isPrimitive) {
+      throw new Error('Patching primitives is not supported.');
+    }
+    const newState = { ...(<any>existingState) };
+    for (const k in val) {
+      newState[k] = val[k];
```
splincode: image

packages/store/src/internal/state-context-factory.ts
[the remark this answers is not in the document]
splincode:
```ts
private _simplePatch(val: Partial<T>) {
   return (existingState: Readonly<T>) => {
      return Object.assign({}, existingState, val);
   }
 }
```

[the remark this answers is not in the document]
splincode: or return 
```ts
{ ...existingState, ...val };
```

packages/store/tests/action-handler.spec.ts
```diff
@@ -0,0 +1,227 @@
...
+      const { store } = setup({ stores: [FooState] });
+      store.dispatch(new TestAction(21));
+      const stateBeforeAsync = store.selectSnapshot(FooState);
+      // Act
+      tick();
+      // Assert
+      const resultState = store.selectSnapshot(FooState);
+      const expectedState = { name: 'my state', age: 21, updated: true };
+      expect(resultState).toEqual(expectedState);
+      expect(resultState).not.toEqual(stateBeforeAsync);
```
splincode: Where test with Immer?

packages/store/tests/action-handler.spec.ts
```diff
@@ -0,0 +1,227 @@
...
+  }
+
+  type _class = { new (...args: any[]) };
+
+  function setup(config: { stores: _class[] }) {
+    config = config || {
+      stores: []
+    };
+    TestBed.configureTestingModule({
+      imports: [NgxsModule.forRoot(config.stores)],
```
splincode: Usage developmentMode for see warning about mutations

```ts
NgxsModule.forRoot(config.stores, { developmentMode: true })
```

packages/store/tests/action-handler.spec.ts
```diff
@@ -0,0 +1,227 @@
...
+
+describe('Action handlers', () => {
+  class TestAction {
+    static type = 'TestAction';
+    constructor(public payload?: any) {}
+  }
+
+  type _class = { new (...args: any[]) };
+
+  function setup(config: { stores: _class[] }) {
```
splincode: you can choose ```Type<unknown>``` instead ```type _class = { new (...args: any[]) }```;

packages/store/src/internal/state-context-factory.ts
```diff
@@ -17,44 +17,52 @@ export class StateContextFactory {
...
-          clone[k] = val[k];
-        }
+        const clone = patchOperation(local);
 
         const newState = setValue(state, metadata.depth, clone);
         root.setState(newState);
         return newState;
       },
-      setState(val: any): any {
+      setState(val: T): T {
```
splincode: I think PatchFunction it will be the best feature

```ts
setState(val: T | PatchFunction<T>): T {
  let state = root.getState();
  const stateValueOperation = this.getValueByOperation(val, CTX_CONTEXT.SET_VALUE); 
  const stateValue = stateValueOperation(state); // (state) => { ...state }
  state = setValue(state, metadata.depth, stateValue);
  ...
}
```

Before

```ts
@Action(MyAction)
public addValue(ctx: StateContext, { payload }: MyAction) {
  ctx.setState({ ...ctx.getState(), value: payload  });
}
```

After

```ts
@Action(MyAction)
public addValue(ctx: StateContext, { payload }: MyAction) {
  ctx.setState((state) => ({ ...state, value: payload  }));
}
```

```ts
patchState(val: Partial<T> | PatchFunction<T>): T {
  const state = root.getState();
  const local = getValue(state, metadata.depth);
  const stateValueOperation = this.getValueByOperation(val, CTX_CONTEXT.PATCH_STATE); 
  const stateValue = stateValueOperation(local); // this.simplePatch(val)(local);
  const newState = setValue(state, metadata.depth, stateValue);
}
```

```ts
@Action(MyAction)
public addValue(ctx: StateContext, { payload }: MyAction) {
  // state = ['1', '2'], payload = '3'
  ctx.patchState(() => ([ payload ])); // state === ['1', '2', '3']
}
```

docs/concepts/state.md
markwhitfeld: Great idea!

packages/store/src/internal/state-context-factory.ts
markwhitfeld: So, are you saying that it would be better on the `setState` function as opposed to the `patchState` function?
I think I like that. Then the `patchState` function could remain unchanged.

packages/store/src/internal/state-context-factory.ts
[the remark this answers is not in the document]
markwhitfeld: It is a utility function and does not need to be on the class or instance. It could move to a separate file.

packages/store/tests/action-handler.spec.ts
markwhitfeld: Oh great! I didn't know about that type. Thanks

packages/store/tests/action-handler.spec.ts
markwhitfeld: At first I was wondering why you pointed this out, but now I realise the spam that appears in the test run. Thanks!

packages/store/tests/action-handler.spec.ts
markwhitfeld: Not necessary because it was just used as an example in the docs.
I'm also not too keen to take a dependency on `immer` just for the tests.

packages/store/src/internal/state-context-factory.ts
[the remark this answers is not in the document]
splincode: @markwhitfeld <T = any, U = any>

packages/store/src/internal/state-context-factory.ts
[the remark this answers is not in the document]
splincode: @markwhitfeld then move the file

packages/store/src/internal/state-context-factory.ts
splincode: It would be logical to have the same interface for setState, patchState

packages/store/src/internal/state-context-factory.ts
```diff
@@ -17,44 +17,52 @@ export class StateContextFactory {
...
         let state = root.getState();
         state = setValue(state, metadata.depth, val);
         root.setState(state);
         return state;
       },
-      dispatch(actions: any | any[]): Observable<any> {
+      dispatch(actions: any | any[]): Observable<void> {
```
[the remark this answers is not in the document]
eranshmil: I guess the following comment from mark will answer your second question:
https://github.com/ngxs/store/issues/606#issuecomment-435174322

packages/store/src/internal/state-context-factory.ts
markwhitfeld: I think that this makes so much more sense for `setState`. 
Having this option for the `patchState` function could leave some ambiguity... ie. what should I be returning from the function? The new state or the value to be used for the patch?

packages/store/src/internal/state-context-factory.ts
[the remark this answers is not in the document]
markwhitfeld: Thanks @eranshmil . I wrote a similar reply but didn't press send. Doh!

packages/store/src/internal/state-context-factory.ts
splincode: setState -> return new state
patchState -> value for state

markwhitfeld: I don't really see any value in this one:
`patchState -> value for state`
I'll just do the one

splincode:
@markwhitfeld Is development underway?

markwhitfeld:
@splincode yes

packages/store/src/internal/state-context-factory.ts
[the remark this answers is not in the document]
markwhitfeld: @splincode You can add those consistency checks in another PR after this one has gone through 👍

docs/concepts/state.md
markwhitfeld: done!

packages/store/tests/action-handler.spec.ts
markwhitfeld: Done

packages/store/tests/action-handler.spec.ts
markwhitfeld: This is being resolved by another PR

markwhitfeld:
@deebloo @splincode @eranshmil Ready for another review 👍

PR #649, opened 2018-11-08
refactor: typings improvement

### LINKED ISSUE #549: [Question] [Discussion] What is the purpose of 'any' in store select?
Hey there,

Our team currently heavily uses ngxs for state management, and something I've noticed is: typing is not there when we are using the select function(s).

https://github.com/ngxs/store/blob/5c2f5b8475c54e830746eeb3d2bbcea616e5363d/packages/store/src/store.ts#L30

With the code base, as is, we receive the following type:

```
(method) Store.select(selector: any): Observable<any> (+1 overload)
```

screenshot 2018-08-17 11 40 15

As the type is resolved as 'any', this actually does mean, in fact, that the selector will resolve to `anything` we want. That is dangerous as any can actually match to anything and everything. This is not the intended behavior observing the source code, from what I see.

However, deleting line 30 results in this behavior:

```
(method) Store.select<Partial<UserPreferences> | undefined>(selector:(state: any, ...states: any[]) => Partial<UserPreferences> | undefined): Observable<Partial<UserPreferences> | undefined> 
```

screenshot 2018-08-17 11 42 34

A snippet of our AuthState:
```typescript
export class AuthState {
    @Selector()
    static preferences(state: AuthStateModel): Partial<UserPreferences> | undefined {
        return state.user ? state.user.preferences : undefined;
    }
}
```

This is arguably the perfect/desired behavior that we are after, as the typescript compiler will fill the types from the given arguments, rather than explicitly passing the type via `store.select<T>() => T`;

Is this behavior already built and we might have somehow wrecked our code base, or is it appropriate to delete the 'any' typing? 

Or, thirdly, is the desired behavior to make the developer explicitly, state the type, rather than implicitly read the type via the arguments?

e-cloud:
fill explicit types for all source codes including test code for better
typesafety. No logic changed.

Besides, there are a little test code changed to conform types.

fix #549

## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

No type safety

Issue Number: #549 

## What is the new behavior?
mostly type safety

```
[ ] Yes
[ ] No
[x] maybe
```

If the user turn on `strict` mode for `tsconfig.json`, he/she might get some type errors


splincode:
@e-cloud Sorry, there are too many changes. Do small pull request...

e-cloud:
@splincode It's hard split this one into several commits. Since I changed the types of store packages, the other packages should be changed too for the valid build and tests.

It's very hard to split them as well as satisfy the CI checks.

If you are fine with the fail checks for split commits, I could split the commit for each package.

eranshmil:
There's some changes (like the decorators in https://github.com/ngxs/store/pull/649/files#diff-fa3d8bfb37529a5145f2d1d32df80f6f) that unnecessary for this PR, that's also caused by issue in prettier.
I fixed that in PR #651, so after it will be merged, update your local yarn dependencies and revert this unnecessary changes. Maybe the PR will be much smaller.

e-cloud:
@splincode @markwhitfeld @eranshmil pr rebased and fix some format problem by old prettier. Please help to review.

eranshmil:
It's still have a lot of format related changes, so you have an auto format in your ide?

e-cloud:
I think it has no problem. Once a line is too long prettier will chop it down, like function params.

Not from my IDE.

packages/store/src/store.ts
```diff
@@ -6,13 +7,14 @@ import { getSelectorFn } from './utils/selector-utils';
...
+import { StateClassStatic } from './internal/internals';
 
 @Injectable()
-export class Store {
+export class Store<T> {
   constructor(
     private _ngZone: NgZone,
-    private _stateStream: StateStream,
-    private _internalStateOperations: InternalStateOperations
+    private _stateStream: StateStream<T>,
```
markwhitfeld: A type `T` des not make sense for the StateStream because the storage structure of the application state is determined by the individual state declarations at run time and a client application should not have any need to map out the structure of the entire application state because of the abstraction provided through selectors.

packages/store/src/store.ts
```diff
@@ -6,13 +7,14 @@ import { getSelectorFn } from './utils/selector-utils';
...
+    private _internalStateOperations: InternalStateOperations<T>
```
markwhitfeld: No type `T` needed here either. Same reasonings as above.

packages/store/src/store.ts
```diff
@@ -25,8 +27,10 @@ export class Store {
   /**
    * Selects a slice of data from the store.
    */
-  select<T>(selector: (state: any, ...states: any[]) => T): Observable<T>;
-  select(selector: string | any): Observable<any>;
+  select<R, K>(selector: (state: T, ...subState: K[]) => R): Observable<R>;
+  select<R, S>(selector: (...states: S[]) => R): Observable<R>;
+  select<R, S>(memorizedSelector: (state: S) => R): Observable<R>;
+  select<R>(selector: string | StateClassStatic): Observable<R>;
```
markwhitfeld: None of these revised selector typings give any substantial advantage over the exisitng ones as only the output type of the selector functions matters for determining the resulting Observable.

packages/store/src/store.ts
```diff
@@ -48,17 +52,21 @@ export class Store {
   /**
    * Select one slice of data from the store.
    */
-  selectOnce<T>(selector: (state: any, ...states: any[]) => T): Observable<T>;
-  selectOnce(selector: string | any): Observable<any>;
+  selectOnce<R, K>(selector: (state: T, ...subState: K[]) => R): Observable<T>;
+  selectOnce<R>(selector: string | StateClassStatic): Observable<R>;
+  selectOnce<R, S>(selector: (...states: S[]) => R): Observable<R>;
+  selectOnce<R, S>(memorizedSelector: (state: S) => R): Observable<R>;
```
markwhitfeld: As above.

packages/store/src/store.ts
```diff
@@ -48,17 +52,21 @@ export class Store {
...
 
   /**
    * Select a snapshot from the state.
    */
-  selectSnapshot<T>(selector: (state: any, ...states: any[]) => T): T;
-  selectSnapshot(selector: string | any): any;
+  selectSnapshot<R, K>(selector: (state: T, ...subState: K[]) => R): R;
+  selectSnapshot<R, S>(selector: (...states: S[]) => R): R;
+  selectSnapshot<R, S>(memorizedSelector: (state: S) => R): R;
+  selectSnapshot<R>(selector: string | StateClassStatic): R;
```
markwhitfeld: As above

e-cloud:
As stated by @byronmejia at https://github.com/ngxs/store/issues/549#issuecomment-422286117, adding these types are trying to use the power of type system to improve application code rubustness.

1. talking about the select function signatures. I think it's worthy to adding them. There is an any type in second signature of the old code, making the typing nullified.

2. Frankly speaking, I haven't considered that would make some confusion for the client user. For your concern, it is worth considering about it. However, IMO, you may put the incidental before the fundamental --

* At the very beginning, what we should do is to make the framework typesafe, but not to fulfill what clients want to do. For that reason and **the natural spreading feature of generic type**, generic type for those class is inevitable(`BehaviorSubject -> StateStream -> Store`). 

* Second, since the fundamental we talking about is the single shared state management, it would be easy to infer, that the generic type of singleton store should be the single application state, not any substate of it.

* BTW, the store from ngrx is also designed with generic type.

* For the misuse from client side, we could improve the documentation to guide them.

3. Although there are changes across many files, there are not much real changes.

The core changes just include:

* generic type for classes
* `select` signature improvement
* decorator signature improvement
* `IAction` interface

All the other changes are just some derived changes from the above. Reviewers could quickly go through those generic `<any>` types and `non-null-assertion` changes.

I should make another PR first to improve the project template.

markwhitfeld:
One of the goals for the 4.0.0 release will be improved type safety throughout (as stricter types will create compile failures for some users so I consider this sort of thing breaking changes) so I am really interested in ways that we can improve the current type safety throughout the codebase (internal and public).

byronmejia:
I agree with @markwhitfeld - Don't get me wrong, typing improvement is a massive change so PRs will naturally be a bit big.

Is there any logical places we can split it up and send smaller PRs though to help you out?

e-cloud:
close this as it has too much outdated info

PR #656, opened 2018-11-12
feat(project): improve build and configurations

e-cloud:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [x] Build related changes
- [x] CI related changes
- [ ] Documentation content changes
- [ ] Other... Please describe:

some non-up-to-date deps and not good enough tools configurations.

## What is the new behavior?
improvement over the problem. updated deps and clearer configurations.

- [ ] Yes
- [x] No


karma.conf.js
```diff
@@ -17,12 +18,17 @@ module.exports = function(config) {
...
+      reports: ['html', 'lcovonly'],
       fixWebpackSourcePaths: true
     },
+    angularCli: {
+      environment: 'production'
+    },
     reporters: ['progress', 'kjhtml'],
     port: 9876,
     colors: true,
+    logLevel: config.LOG_INFO,
```
splincode: why you need usage logLevel instead?

```ts
browserConsoleLogOptions: {
  level: CI ? 'error' : 'debug'
}
```

splincode:
You should not mix changes in one branch.

docs/concepts/select.md
```diff
@@ -264,9 +264,9 @@ export class ZooState {
 }
 ```
 
-When using the `Selector` decorator along with a state class, it will still
-inject the state class's state first followed by the other selectors in the order
-they were passed in the signature.
+> NOTE: When using the `Selector` decorator along with a state class, **it will still
```
splincode: Take out in a separate PR

tsconfig.build.json
```diff
@@ -1,21 +1,34 @@
...
+    "lib": [
+      "dom",
+      "es2017"
+    ],
+    "paths": {
+      "@ngxs/*": ["@ngxs/*"]
+    },
+  },
+  "angularCompilerOptions": {
+    "annotateForClosureCompiler": true,
```
splincode: What are the advantages?

tsconfig.lib.json
```diff
@@ -0,0 +1,30 @@
...
+    "sourceMap": true,
+    "inlineSources": true,
+    "importHelpers": true,
+    "types": [],
+    "lib": [
+      "dom",
+      "es2017"
+    ]
+  },
+  "angularCompilerOptions": {
```
splincode: Why not immediately take out a separate file `tsconfig.json` and not inherit from it?

karma.conf.js
e-cloud: @splincode these two are with different scope. loglevel is for karma internal log, browserConsoleLogOptions is for reporter.

tsconfig.build.json
e-cloud: this will generate some information in output, which could be used by user using closure.

karma.conf.js
splincode: but we will be moving to jest soon and karma internal log was not really needed

tsconfig.build.json
splincode: Does it increase the bundle?

splincode: good

e-cloud:
@splincode could you review again?

tsconfig.lib.json
splincode: angularCompilerOptions move to tsconfig.json

splincode:
@markwhitfeld could you take a look?

e-cloud:
@splincode @markwhitfeld any progress?

e-cloud:
@markwhitfeld please have a review

markwhitfeld:
@e-cloud I'm happy if the others are happy, just testing locally before I merge...

PR #663, opened 2018-11-19
fix(store): included underscore in stateNameRegex

JasonAtallah:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: N/A

underscores in state name return an error saying the state name isn't valid because it is not a 'valid object property name'

## What is the new behavior?
underscores are allowed in state name

```
[ ] Yes
[x] No
```

### DIFF
--- a/packages/store/src/decorators/state.ts
+++ b/packages/store/src/decorators/state.ts
@@ -1,7 +1,7 @@
 import { ensureStoreMetadata } from '../internal/internals';
 import { StoreOptions, META_KEY } from '../symbols';
 
-const stateNameRegex = new RegExp('^[a-zA-Z0-9]+$');
+const stateNameRegex = new RegExp('^[a-zA-Z0-9_]+$');
 
 /**
  * Error message
--- a/packages/store/tests/store.spec.ts
+++ b/packages/store/tests/store.spec.ts
@@ -14,7 +14,7 @@ describe('Store', () => {
   interface SubStateModel {
     hello: boolean;
     world: boolean;
-    baz?: SubSubStateModel;
+    baz_?: SubSubStateModel;
   }
 
   interface StateModel {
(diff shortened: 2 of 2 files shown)


JasonAtallah:
@splincode, the error isn't logical as an underscore is a valid object property name.

Use case:
When trying to construct a payload through substates for an API call that has underscores in property names (common practice for APIs built in Python), this error adds unnecessary logic for me to find and replace field names before making the request.

packages/store/tests/store.spec.ts
```diff
@@ -28,7 +28,7 @@ describe('Store', () => {
   }
 
   @State<SubSubStateModel>({
-    name: 'baz',
+    name: 'baz_',
```
splincode: Could you return the changes in this test and write a new one that corrects your error?

splincode: Could you return the changes in this test and write a new one that corrects your error?

packages/store/src/decorators/state.ts
```diff
@@ -1,7 +1,7 @@
 import { ensureStoreMetadata } from '../internal/internals';
 import { StoreOptions, META_KEY } from '../symbols';
 
-const stateNameRegex = new RegExp('^[a-zA-Z0-9]+$');
+const stateNameRegex = new RegExp('^[a-zA-Z0-9_]+$');
```
splincode: Is this a breaking change?

JasonAtallah: No

packages/store/tests/store.spec.ts
markwhitfeld: Agreed, @JasonAtallah please make a copy of an existing test and test your own scenario in a test specifically for that scenario (leaving the existing tests as they were).
Thanks!

JasonAtallah: @splincode, @markwhitfeld I reverted the changes I made to existing tests and created a new test with a new state. The only change I made to the current tests was adding my new state to the root state.

Issue #221, opened 2018-04-05
Storage-Plugin should support async storage engines


BeaveArony:
Some storage engines are promise or callback based. Like LocalForage, idb-keyval or native mobile storage implementations.

Probably the main challenge is to delay the first rendering of the app until the state is actually rehydrated from an async storage. redux-persist does this with react.
I dunno, but maybe ngxs could hook into the router to cancel the first navigation attempt, waiting for some callback from the plugin to redo the navigation once rehydration is completed? But not every app uses the router ...

Another useful feature would be to define blacklisted keys. Imagine you have a big feature state you want to persist, except one key (ie. loading or pending). You would have to define a huge list of keys to make it happen, right?

AlgoTrader:
Render is probably not the state management business. App should decide itself how to do it. If you have async state, it is better to provide default state immediatly and then patch it with real state forcing re-render

amcdnl:
I think the vast majority of cases will want sync. I imagine you have a indexdb use case or something. It would be easy to make a new plugin to support this but not something that is priority on my list right now.

BeaveArony:
Thanks for the answers.
I don't know, if the vast majority will (want) localStorage. It's slow (you need to JSON.stringify()/parse() everything), blocks the UI-Thread and is very limited with mostly 5MB of data and is only supported in browsers. As soon as you want to use NativeScript or IndexedDB or whatever comes in the future, you will most likely need async and the ability to react on the rehydration event.

I don't think that NGXS should bring all of those features by default, but since it provides quite some abstraction of redux and the bootstrapping process, it should offer some kind of entry into the initialization point of the store, so we could wait on some callback to render our app for example. Maybe I'm missing something and it's easily doable?

Plugins are comparable to middleware in redux. You hook into each action to do something before and/or after and you might even swallow it. Is it possible to hook into the createStore() process to alter the initial state from within a plugin?

amcdnl:
Closing and marking for future.

alejandro-heredia-coderoad-com:
Couldn't agree more with @BeaveArony, NGXS is just too powerful, smart and stable to only support sync storage. Async storage engines should be priority.

kctang:
I am using NGXS in an Ionic application that needs to support various devices. LocalStorage support in older devices is quite 'unpredictable' to say the least (even iOS 10.3.3)! So, having the option to swap storage engine with things like Sqlite or IndexDB will be critical.

Any chance we can up the priority of this item?

[...]

Issue #465, opened 2018-06-26
[FR] SSR


John0x:
## Feature Request
It would be great to have some builtin way to handle SSR or at least have a section in the docs for it

amcdnl:
I have no experience with SSR, can you write this up for the community?

odahcam:
Basically you gotta polyfill or check if browser specific things exists before start using them, like the `localStorage`, it is not available in the server and throws an error when running in the server.

odahcam:
For example, in these lines:
https://github.com/ngxs/store/blob/b42b4a405a3b11fbdf4243c4a3821228c069477d/packages/storage-plugin/src/storage.module.ts#L23-L31

We should verify if both `localStorage` and `sessionStorage` are defined before returning any, otherwise we get errors like that:

```
$ yarn serve:universal
yarn run v1.7.0
$ cd dist && node server
Node Express server listening on http://localhost:4000
ReferenceError: localStorage is not defined
    at engineFactory (\path\to\my\project\Havan.Vc.Interface.PWA\dist\server.js:228592:9)
    at _callFactory (\path\to\my\project\Havan.Vc.Interface.PWA\dist\server.js:12348:20)
    at _createProviderInstance (\path\to\my\project\Havan.Vc.Interface.PWA\dist\server.js:12304:26)
    at initNgModule (\path\to\my\project\Havan.Vc.Interface.PWA\dist\server.js:12234:32)
    at new NgModuleRef_ (\path\to\my\project\Havan.Vc.Interface.PWA\dist\server.js:12960:9)
    at Object.createNgModuleRef (\path\to\my\project\Havan.Vc.Interface.PWA\dist\server.js:12949:12)
    at NgModuleFactory_.create (\path\to\my\project\Havan.Vc.Interface.PWA\dist\server.js:15491:25)
    at \path\to\my\project\Havan.Vc.Interface.PWA\dist\server.js:8221:43
    at ZoneDelegate.invoke (\path\to\my\project\Havan.Vc.Interface.PWA\dist\server.js:533:26)
    at Object.onInvoke (\path\to\my\project\Havan.Vc.Interface.PWA\dist\server.js:7732:33)
```

odahcam:
**Workaround for `localStorage` in server-side:**

#### app.module

```typescript
// app.module

// @angular
import { NgModule } from '@angular/core';
import { BrowserModule } from '@angular/platform-browser';

// @env
import { environment } from '@env/environment';

// State management
[...]

PR #669, opened 2018-11-22
feat(storage-plugin): add support for async custom storage engine

TomDemulierChevret:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Currently the storage plugin only works with synchronous storage engines (localStorage, sessionStorage or any custom storage engine that implements ``StorageEngine``).

Issue Number: #221 

## What is the new behavior?
The ``StorageEngine`` API has been modified to accept asynchronous storage engines. It's ``getItem`` method now returns an ``Observable``.

At state initialisation, the ``StoragePlugin`` now retrives the stored keys asynchronously but keep the retrieval order (and therefore the migration order).

This was made easy by the fact that the current plugin pipe (with each plugin having a ``handle`` method) is based on ``Observable``.

```
[] Yes
[x] No
```

## Other information

Here is a small example based on my current Ionic 4 app.
My custom storage engine uses the [Ionic storage service](https://beta.ionicframework.com/docs/building/storage/) (which uses internally [localForage](https://github.com/localForage/localForage) for ``IndexedbDB``/``WebSQL``/``localStorage`` and [cordova-sqlite-storage](https://github.com/litehelpers/Cordova-sqlite-storage) for ``SQLite``).

### app.module.ts
```typescript
@NgModule({
  declarations: [AppComponent],
  entryComponents: [],
  imports: [
    BrowserModule,
    IonicModule.forRoot(),
    IonicStorageModule.forRoot(),
    AppRoutingModule,
    NgxsModule.forRoot([YourState]),
      serialize: (obj: any): any => obj,
      deserialize: (obj: any): any => obj
    }),
  ],
  providers: [
    {
      provide: STORAGE_ENGINE,
      useClass: StorageService
    }
  ],
  bootstrap: [AppComponent]
})
export class AppModule {}
```

### storage.service.ts
```typescript
@Injectable()
export class StorageService implements AsyncStorageEngine {
  constructor(private storage: Storage) {}

  length(): Observable<number> {
    return from(this.storage.length());
  }

  getItem(key: any): Observable<any> {
    return from(this.storage.get(key));
  }

  setItem(key: any, val: any): void {
    this.storage.set(key, val);
  }

  removeItem(key: any): void {
    this.storage.remove(key);
  }

  clear(): void {
    this.storage.clear();
  }

  key(val: number): Observable<string> {
    return from(this.storage.keys().then(keys => keys[val]));
  }

}
```


packages/storage-plugin/tests/storage.plugin.spec.ts
```diff
@@ -218,29 +219,13 @@ describe('NgxsStoragePlugin', () => {
         }
       };
 
-      get length() {
```
splincode: You can not delete detail test

packages/storage-plugin/src/symbols.ts
```diff
@@ -58,12 +59,26 @@ export const NGXS_STORAGE_PLUGIN_OPTIONS = new InjectionToken('NGXS_STORAGE_PLUG
...
+ * StorageEngine API is similar to the WebStorage API (localStorage or sessionStorage) but can be used with asynchronous
+ * storage engines such as IndexedDB
+ */
 export interface StorageEngine {
-  readonly length: number;
-
-  getItem(key): any;
+  getItem(key): Observable<any>;
   setItem(key, val): void;
-  removeItem(key): void;
```
splincode: Return back

packages/storage-plugin/tests/storage.plugin.spec.ts
TomDemulierChevret: It was to be consistent with the StorageEngine API.
I will add it back when the comment below is resolved.

packages/storage-plugin/src/symbols.ts
TomDemulierChevret: I removed som methods from ``StorageEngine`` since they were unused and would have forced people to make a more complex than necessary custom storage engine implementation.

Is there any plans to use theses methods in the future ?
Do you me to add back all of them or onlye ``removeItem`` ?

splincode: in your case, these are critical changes and those who use it can break their application.

splincode:
you must make a parameterized choice so that there is no breaking changes

```ts
export enum NgxsStorageType {
  ASYNC = 'async',
  SYNC = 'sync'
}
```

```ts
@NgModule({
  declarations: [AppComponent],
  entryComponents: [],
  imports: [
    BrowserModule,
    IonicModule.forRoot(),
    IonicStorageModule.forRoot(),
    AppRoutingModule,
    NgxsModule.forRoot([YourState]),
      key: ['yourstate.key'],
      type: NgxsStorageType.ASYNC, // Default 'sync'
      serialize: (obj: any): any => obj,
      deserialize: (obj: any): any => obj
    }),
  ],
  providers: [
    {
      provide: STORAGE_ENGINE,
      useClass: StorageService
    }
  ],
  bootstrap: [AppComponent]
})
export class AppModule {}
```

TomDemulierChevret:
You're right, it shouldn't introduce breakind changes.

After some analysis, my current implementation doesn't really work (it only did the trick because my async storage was nearly instantaneous).

I'll rewrite it and see if deeper changes are needed for the PR.

markwhitfeld:
@TomDemulierChevret 
Just a thought... We could leave the `StorageEngine` interface alone and then create a new `AsyncStorageEngine` with the tweaked method signatures.
Then, your async adjustments to the plugin code (once corrected) could just use an `AsyncStorageEngineProxy` around the supplied storage engine. This proxy would delegate its methods to the underlying storage engine that has been supplied by the user and if the return value is not an observable then the value would be wrapped using the `of` operator.
Something like this:
```TS
export class AsyncStorageEngineProxy implements AsyncStorageEngine {
  constructor(private _storageEngine: StorageEngine | AsyncStorageEngine) {}
  
  private asObservableValue(value) {
    return (value instanceOf Observable) ? value : of(value);
  }

  length(): Observable<number> {
    return asObservableValue(this._storageEngine.length);
  }
  getItem(key): Observable<any> {
    return asObservableValue(this._storageEngine.getItem(key));
  }
  setItem(key, val): void {
    this._storageEngine.setItem(key, val);
  }
  removeItem(key): void {
    this._storageEngine.removeItem(key);
  }
  clear(): void {
    this._storageEngine.clear();
  }
  key(val: number): Observable<string> {
    return asObservableValue(this._storageEngine.key(val));
  }
}
```
**Note:** this code was written in the comment so please don't assume it is correct.

TomDemulierChevret:
@markwhitfeld Yeah, I was thinking of going somewhere like that (there is already a wrapper/proxy for sync storage engines).
Althought I'm not sure testing if the value is an Observable everytime the method is called is a good idea performance-wise.
Since the storage engine is not supposed to be changed during runtime (is it ?) we could just make another test at startup and then provide the appropriate proxy.

But currently I need to modify the `compose` function that allows to chain plugins `handle` methods so it can handles async plugins.

TomDemulierChevret:
Actually I didn't have to rewrite anything outside of the storage plugin.
The issue with my test was that the ``Increment`` actions where launched before the init was concluded (and therefore the previous state retrieval from the storage).

markwhitfeld:
@TomDemulierChevret I think that if it is at all possible to get away from the need to specify if the storage engine is async or not then that would be first prize!
Maybe we could detect the type by calling length on the current engine and see if that is a number or not (async).

markwhitfeld:
Ps. Does the async storage return an Observable or Promise? A Promise would make more sense. What platform are we dealing with?

splincode:
@markwhitfeld Observable support Promise -> Observable.prototype.toPromise

markwhitfeld:
@splincode Yeah, I know but what does the targeted use case need? This plugin would only ever want a single value and not a stream as far as I can see.

TomDemulierChevret:
@markwhitfeld I agree, it would be detect automatically the engine type. I'll see what I can do.

Regarding ``Promise`` vs ``Observable``, I prefer the second one for 2 reasons :
1. They integrate much more nicely into the existing plugin flow (since the ``next`` argument/method returns an Observable). Using ``Promise`` for the storage internal API would force to call rxjs ``from`` constructor to convert the ``Promise`` into and ``Observable`` which would add unnecessary complexity to the ``handle`` method in the plugin.
2. Most Angular async API uses ``Observable`` regardless of the number of returned values. Using ``Observable`` is therefore more consistent with the ecosystem.
For example, a user could easily use Angular ``HttpClient`` to implement a custom storage engine that store the states in a server.

markwhitfeld:
Sounds like good reasoning!
See what you can do for detection. 
It would be great to add a sample to the docs of an example integration once this goes through

TomDemulierChevret:
Just added the auto detection.

For the doc sample, would my example in the first post be enough ?

markwhitfeld:
I think it would be ok, but if there was an example that for not require you to understand the ionic storage api then that would be better.
Ps. I think that the integration test failures are to do with your changes. Potentially the sequencing of things has changed. It may be showcasing an example where this change subtly changes the order or timing of things

TomDemulierChevret:
After some debugging, you're right : the integration test failures are linked to my changes.

The big issue is that my changes broke something that **shouldn't have worked in the first place**.
It's linked to how the ``form-plugin`` react to value changes and how the ``dispatcher`` handle multiple actions dispatched at the same time.

1) The directive ``FormDirective`` (in ``packages\form-plugin\src\directive.ts``) ensure the synchronisation between the state by dispatching 3 actions (``UpdateFormValue``, ``UpdateFormDirty`` & ``UpdateFormErrors``) at the same time :
```typescript
@Directive({ selector: '[ngxsForm]' })
export class FormDirective implements OnInit, OnDestroy {
  // irrelevant stuff //

  ngOnInit() {
    // irrelevant stuff //

    // On first state change, sync form model, status and dirty with state
    this._store
      .select(state => getValue(state, `${this.path}`))
      .pipe(
        takeUntil(this._destroy$),
        first()
      )
      .subscribe(state => {
        this._store.dispatch([
          new UpdateFormValue({
            path: this.path,
            value: this._formGroupDirective.form.getRawValue()
          }),
          new UpdateFormStatus({
            path: this.path,
            status: this._formGroupDirective.form.status
          }),
          new UpdateFormDirty({
            path: this.path,
            dirty: this._formGroupDirective.form.dirty
          })
        ]);
      });
    
    // irrelevant stuff //

    this._formGroupDirective.valueChanges
      .pipe(
        debounceTime(this.debounce),
        takeUntil(this._destroy$)
      )
      .subscribe(() => {
        const value = this._formGroupDirective.control.getRawValue();
        this._updating = true;
        this._store
          .dispatch([
            new UpdateFormValue({
              path: this.path,
              value
            }),
            new UpdateFormDirty({
              path: this.path,
              dirty: this._formGroupDirective.dirty
            }),
            new UpdateFormErrors({
              path: this.path,
              errors: this._formGroupDirective.errors
            })
          ])
          .subscribe({
            error: () => (this._updating = false),
            complete: () => (this._updating = false)
          });
      });
  }
}
```

2. The provider ``InternalDispatcher`` (in ``packages\store\src\internal\dispatcher.ts``) handle these 3 actions dispatched by using ``forkJoin`` which will execute 3 times the ``dispatchSingle`` (once for each action) method **in parallel**.
The ``dispatchSingle`` method will get the current state and pass it to the composition of plugins ``handle`` methods to take care of one action. In our case, all 3 actions will therefore receive the same state (for example with ``model`` still set to undefined).
Once every plugin has done its side effet, the last function from the composition will be called and will set the state to its new value (and some plugin will then trigger a second side effet, but it's irrelevant here) :
```typescript
@Injectable()
export class InternalDispatcher {
  // irrelevant stuff //

  /**
   * Dispatches event(s).
   */
  dispatch(event: any | any[]): Observable<any> {
    const result: Observable<any> = this._ngZone.runOutsideAngular(() => {
      if (Array.isArray(event)) {
        return forkJoin(event.map(a => this.dispatchSingle(a)));
      } else {
        return this.dispatchSingle(event);
      }
    });

    // irrelevant stuff //
  }

  private dispatchSingle(action: any): Observable<any> {
    const prevState = this._stateStream.getValue();
    const plugins = this._pluginManager.plugins;

    return (compose([
      ...plugins,
      (nextState, nextAction) => {
        if (nextState !== prevState) {
          this._stateStream.next(nextState);
        }
        const actionResult$ = this.getActionResultStream(nextAction);
        actionResult$.subscribe(ctx => this._actions.next(ctx));
        this._actions.next({ action: nextAction, status: ActionStatus.Dispatched });
        return this.createDispatchObservable(actionResult$);
      }
    ])(prevState, action) as Observable<any>).pipe(shareReplay());
  }

  // irrelevant stuff //
}
```

3) The provider ``NgxsFormPlugin`` (in ``packages\form-plugin\src\form.plugin.ts``) will handle the actions related to forms (in our case the 3 actions dispatched earlier  ``UpdateFormValue``, ``UpdateFormDirty`` & ``UpdateFormErrors``). Each action will get the previous state and update the relevant stuff inside depending of its type and then pass the updated state to the returned ``next`` function :
```typescript

@Injectable()
export class NgxsFormPlugin implements NgxsPlugin {
  constructor() {}

  handle(state, event, next) {
    const type = getActionTypeFromInstance(event);

    let nextState = state;

    if (type === UpdateFormValue.type || type === UpdateForm.type) {
      const { value } = event.payload;
      const payloadValue = Array.isArray(value) ? [...value] : { ...value };

      nextState = setValue(nextState, `${event.payload.path}.model`, payloadValue);
    }

    // irrelevant stuff //

    if (type === UpdateFormErrors.type || type === UpdateForm.type) {
      nextState = setValue(nextState, `${event.payload.path}.errors`, {
        ...event.payload.errors
      });
    }

    if (type === UpdateFormDirty.type || type === UpdateForm.type) {
      nextState = setValue(nextState, `${event.payload.path}.dirty`, event.payload.dirty);
    }

    // irrelevant stuff //

    return next(nextState, event);
  }
}
```

We can sum up the flow like this :
1) action A, B & C get dispatched
2) action A, B & C received the current state S
3) action A updates the state S to state SA, completes and returns SA
4) action B updates the state S to state SB, completes and returns SB
5) action C updates the state S to state SC, completes and returns SC

So in our case, only the result from the action ``UpdateFormDirty`` will be taken into account in the updated state.

Before my modifications the flow ought to be something like this :
1) action A, B & C get dispatched
2) action A receive the current state S
3) action A updates the state S to state SA, completes and returns SA
2) action B receive the current state SA
4) action B updates the state SA to state SAB, completes and returns SAB
2) action C receive the current state SAB
5) action C updates the state S to state SABC, completes and returns SABC

But since the ``dispatched`` uses ``forkJoin`` **it shouldn't have worked** and only did because the ``dispatchSingle`` method updated the state fast enough so that the next action will get the updated state.

markwhitfeld:
Thank you for the in depth investigation! 
The current flow that results in SABC is correct. There is no such thing as parallel execution in javascript. It is single threaded. I'll have a read through your code to see if I can spot where it is breaking the flow of the state through a sequence of actions.

Ps. I had a laugh with your example because our national TV broadcaster is called SABC🤣

TomDemulierChevret:
Of course parallel execution does not exist in JS since it's single threaded (except of course if you use webworkers).
But the rxjs operator ``forkJoin`` is literally made to executes ``Observable`` in a pseudo parallel way.

If you have 2 ``Observable`` A & B and do ``forkJoin(A, B)``, you are guaranteed that the beginning of A will execute before the beginning of B but **your are not** guaranteed that the end of A will execute before the beginning of B. Yet the current implementation of the ``form-plugin`` is based on it.

TomDemulierChevret:
The correct way to dispatch an array of actions in sequence would be (given actions [A, B, C] something like :
```typescript
this.dispatchSingle(A)
  .pipe(
    concatMap(() => this.dispatchSingle(B)),
    concatMap(() => this.dispatchSingle(C))
  )
```
Since we don't know the number of actions, the more generic way would be :
```typescript
of(null)
  .pipe(
    pipe.apply(this, event.map(a => concatMap(() => this.dispatchSingle(a))))
  )
```

I tested locally this one and it solves the integration tests but it makes 3 unit tests fail (all linked to actions dispatched in array), so it's obvioulsy not the right way to fix the issue.

splincode:
@TomDemulierChevret storage plugin can be supported Server Side Rendering

TomDemulierChevret:
@splincode I'm not sure to understand your post.
Do you mean the current plugin support SSR and the PR should not break it.
Or is it question wether this PR can help the support of SSR ?

On a side-note, can you resolve the changes you requested ? Or do you need other changes ?

splincode:
@TomDemulierChevret I think it would be great if you also added support SSR

splincode:
https://github.com/ngxs/store/issues/465

TomDemulierChevret:
Well, I'm not sure we can support SSR directly since we don't which technology will be used on the server side (and therefore how the state can be stored).

But, since the plugin now support asynchronous get/set, it's easy to write a custom engine storage based on Angular ``HttpClient`` and then use https://github.com/angular/universal/blob/master/docs/transfer-http.md.

But i'm not really experienced with SSR, so there could be other problems that I don't know.

TomDemulierChevret:
1. For the test, please read the posts above (starting with this one https://github.com/ngxs/store/pull/669#issuecomment-443512539)
2. For SSR, it's quite out of scope of this PR and I don't think I'm experienced enough with it to advance on this topic.

TomDemulierChevret:
@markwhitfeld Had you had time to look at the synchronisation problem ?

@splincode Did you read my previous comments about the eventual solutions to fix the failing tests ? What do you think of them ?

splincode:
@TomDemulierChevret You can rewrite tests for correct works, but this is breaking changes for 3.x

TomDemulierChevret:
The problem is not how the test are written but how the form-plugin works.
And I agree that my proposed changes could be breaking changes.

splincode:
I suggest rewriting it to you in a separate PR

markwhitfeld:
@TomDemulierChevret I was also confused with the SSR comments. It doesn't relate to this PR directly, but I guess that the thought is that the idea might influence the thinking about this PR through introducing another possible use case.

Regarding the synchronisation problem, I'll hopefully have a look in the next few days. Very busy with all the year end stuff going on.

splincode:
@markwhitfeld @TomDemulierChevret current PR is relevant?

marcjulian:
I would like to see this PR in a next release. I am using NGXS in a new Ionic 4 App and I would love to test this feature.

signuspl:
> I would like to see this PR in a next release. I am using NGXS in a new Ionic 4 App and I would love to test this feature.

I'm in the same situation (Ionic 4 + NXGS + PouchDB as storage). Just use the ngxsOnInit in your state class to load your state from database at startup or init it with some data. I just dispatch an action to load my data from database and use some "init" flag inside state to show that state is initalized:

`ngxsOnInit(ctx: StateContext<AppStateModel>) {
    // console.log('State initialized, now getting animals');
    ctx.dispatch(new LoadConfig());
  }

@Action(LoadConfig)
  loadConfig(ctx: StateContext<AppStateModel>, action: LoadConfig) {
    const state = ctx.getState();
    this.storage.get('app').then((data) => {
      // console.log('----- init app state');
      // console.log(data);
      ctx.setState({...data, ...{init: true}});
      ctx.dispatch(new LoadRestaurants());
      ctx.dispatch(new LoadUsers());
    }, (err) => {
      console.log(err);
      ctx.patchState({
        init: true
      });
    });
  }
`
Hope this will help you until they release async storage solution.

splincode:
The current PR does not develop, you can rediscover it in the future.

TomDemulierChevret:
Hi,

I was on holyday for 2 weeks and was therefore unable to answer comments.
There was no developments recently as I was waiting input from your side. ;)

markwhitfeld:
Hope you had a great break!
I think that the development of this as a labs project is a brilliant idea. No risk of introducing breaking changes. Thanks @TomDemulierChevret for doing the ground work on this! I would recommend that you track the labs project closely to ensure that is solves your problem too. 
Thanks @marcjulian for taking this on 👍 .

marcjulian:
No problem, thanks for a great head start @TomDemulierChevret 👍 
Feedback is appreciated ;)

TomDemulierChevret:
Don't hesitate if you need some tests done on a real app. 👍

markwhitfeld:
@TomDemulierChevret ping me in the ngxs slack and I'll add you to the labs channel for this plugin.
I couldn't find your username

larscom:
What is the status? i'm kinda interested... :-)

splincode:
@larscom https://github.com/ngxs-labs/async-storage-plugin

larscom:
> @larscom https://github.com/ngxs-labs/async-storage-plugin

Thanks! will check it out :-)

Issue #692, opened 2018-12-03
The way that keep the state on HMR


junimohano:
Is there any way to keep state on HMR?

Always refreshes the state on HMR as it is really inconvenient.

I couldn't find any resources so far.

Impossible to create features?

kayvanbree:
Angular already has hot module replacement right? What's the point of using it the way described in the docs?

splincode:
`@ngxs/hmr` coming soon

kayvanbree:
@junimohano This can be closed right?

splincode:
No

eranshmil:
The plugin is not published yet, because there's an issue with lazy-loaded modules.
#719

Issue #567, opened 2018-09-13
Documentation needed for Angular HMR


matulkum:
How do I setup ngxs when using Hot Module Replacement (as described for example here: https://github.com/angular/angular-cli/wiki/stories-configure-hmr)?

Currently I end up with a new global NGXS instance on every update.

splincode:
@marcus-sa I will add documentation soon

markwhitfeld:
Thanks @splincode

markwhitfeld:
PR merged that includes the requested documentation. Closing the issue.
@matulkum Please test and give your feedback.

matulkum:
Unless I am missing something, the guide on ngxs.gitbook is pretty much just a copy of the recipe page from the angular-cli documentation.

Still the problem persist, that I end up with a new global NGXS instance on every reload.

splincode:
@matulkum Did you want your state to be saved when the application is restarted? I do not think it's a good idea

splincode:
@matulkum you need a create new issue and suggest solutions

matulkum:
I do not have a solution.

This issue can not be open again. Otherwise I would just copy paste this issue text?

splincode:
This issue was not formed by pattern.
https://github.com/ngxs/store/blob/master/.github/ISSUE_TEMPLATE.md

... and not clear what problem she solves

splincode:
I think it is impossible at the moment, or send a pull request

markwhitfeld:
Hmmm... this could be to do with the dev tools plugin... or it could be an issue with the Redux dev tools themselves. Would need an investigation.

splincode:
@matulkum https://github.com/ngxs/store/pull/705/files

Polyterative:
Is this possible? Being HRM a potential change in the code, maintaining the state while changing the code could corrupt the data store. Someone please correct me if I'm wrong

splincode:
@Polyterative 

```
$ npm install @ngxs/hmr-plugin
```

https://ngxs.gitbook.io/ngxs/v/master/plugins/hmr

PR #698, opened 2018-12-03
 fix(store): [SSR] - Action dispatch does not work properly

Gorniv:
## PR Checklist
Please check if your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://github.com/ngxs/store/blob/master/CONTRIBUTING.md#commit
- [x] Tests for the changes have been added (for bug fixes / features)
- [x] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?
```
[x] Bugfix
[ ] Feature
[ ] Code style update (formatting, local variables)
[ ] Refactoring (no functional changes, no api changes)
[ ] Build related changes
[ ] CI related changes
[ ] Documentation content changes
[ ] Other... Please describe:
```
## What is the current behavior?
https://github.com/ngxs/store/issues/696
https://github.com/ngxs/store/issues/465
https://github.com/angular/universal/issues/1061

## What is the new behavior?
run dispatch in ngZone for server side

## Does this PR introduce a breaking change?
```
[ ] Yes
[x] No
```

PR #707, opened 2018-12-12
feat(hmr-plugin): support manage state when usage hmr

splincode:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

https://github.com/ngxs/store/issues/692
https://github.com/ngxs/store/issues/567

## What is the new behavior?

hmr

```
[ ] Yes
[x] No
```

## Other information

main.ts

```ts
import { hmrNgxsBootstrap } from '@ngxs/hmr-plugin';
import { environment } from './environments/environment';

const bootstrap = () => platformBrowserDynamic().bootstrapModule(AppBrowserModule);

if (environment.hmr) {
  if (module['hot']) {
    hmrNgxsBootstrap(module, bootstrap);
  } else {
    console.error('HMR is not enabled for webpack-dev-server!');
    console.log('Are you using the --hmr flag for ng serve?');
  }
} else {
  bootstrap().catch(err => console.log(err));
}

```

app.module.ts

```ts
import { StateStream } from '@ngxs/store';
import { NgxsHmrPlugin, NgxsStoreSnapshot } from '@ngxs/hmr-plugin';

@NgModule({ .. })
export class AppModule implements NgxsHmrPlugin {
 
  public hmrNgxsStoreOnInit(state: StateStream, snapshot: NgxsStoreSnapshot) {
    console.log('[NGXS HMR]: Current state', state.getValue());
    console.log('[NGXS HMR]: Previous state', snapshot);
    state.next({ ...state.getValue(), ...snapshot });
  }

  public hmrNgxsStoreOnDestroy(state: StateStream): NgxsStoreSnapshot  {
    console.log('[NGXS HMR]: Return the necessary HMR state for reuse...');
    return state.getValue();
  }

}
```


integration/app/app.browser.module.ts
```diff
@@ -18,4 +20,17 @@ import { AppModule } from './app.module';
   ],
   providers: [{ provide: 'ORIGIN_URL', useValue: location.origin }]
 })
-export class AppBrowserModule {}
+export class AppBrowserModule implements NgxsHmrPlugin {
```
markwhitfeld: Maybe `NgxsHmrPlugin` could be better named `NgxsHmrLifeCycle`?

packages/hmr-plugin/package.json
```diff
@@ -0,0 +1,19 @@
...
+    "@angularclass/hmr": ">=2.1.3 <3.0.0"
+  },
+  "ngPackage": {
+    "lib": {
+      "flatModuleFile": "ngxs-store",
+      "entryFile": "index.ts",
+      "umdModuleIds": {}
+    },
+    "dest": "../../@ngxs/hmr-plugin"
+  }
```
markwhitfeld: The ngPackage config doesn't seem correct.
Other plugins have this:
```TS
"ngPackage": {
    "lib": {
      "flatModuleFile": "ngxs-storage-plugin",
      "entryFile": "index.ts",
      "umdModuleIds": {
        "@ngxs/store": "ngxs-store"
      }
    },
    "dest": "../../@ngxs/storage-plugin"
  }
```

docs/advanced/hmr.md
```diff
@@ -145,6 +124,29 @@ if (environment.hmr) {
...
+    state.next({ ...state.getValue(), ...snapshot });
+  }
+
+  public hmrNgxsStoreOnDestroy(state: StateStream): NgxsStoreSnapshot  {
+    console.log('[NGXS HMR]: Return the necessary HMR state for reuse...');
+    return state.getValue();
+  }
+
+}
+```
```
markwhitfeld: I would prefer if we could get the users away from direct interaction with the state stream.
I think we should rather give them a `StateContext` for these interactions:
```TS
import { StateContext } from '@ngxs/store';
import { NgxsHmrLifeCycle, NgxsStoreSnapshot } from '@ngxs/hmr-plugin';

@NgModule({ .. })
export class AppModule implements NgxsHmrLifeCycle {

  // If this hook is not supplied then it defaults to doing a patchState with the hmr snapshot
  public hmrNgxsStoreOnInit(ctx: StateContext<any>, snapshot: NgxsStoreSnapshot) {
    console.log('[NGXS HMR]: Current state', cts.getState();
    console.log('[NGXS HMR]: Previous state', snapshot);
    // Added something exra to the state here for demo purposes
    ctx.patchState({ ...snapshot, hmr: { status: 'loaded' } });
    // They could use setState if they want to replace the entire state
    // They could also potentially even dispatch some custom actions that let the State Classses choose what to do with the hmr state
  }

  // If this hook is not supplied then it defaults to snapshotting the entire state
  public hmrNgxsStoreOnDestroy(ctx: StateContext<any>): NgxsStoreSnapshot {
    console.log('[NGXS HMR]: Return the necessary HMR state for reuse...');
    return ctx.getState();
  }
}
```
I am in two minds about the ctx for the destroy because that would also make `setState`, `patchState`, and `dispatch` available. I can't think of a usecase for using those at the destroy phase. That being said, maybe it could be useful to have. Not sure what you could still do while the destroy is happening.

PS. Note that I also renamed `NgxsHmrPlugin` to `NgxsHmrLifeCycle` above.

packages/hmr-plugin/src/symbols.ts
```diff
@@ -0,0 +1,21 @@
+import { StateStream } from '@ngxs/store';
+
+export interface NgxsStoreSnapshot {
+  [key: string]: any;
+}
```
markwhitfeld: Do we need an interface for the snapshot?

integration/app/app.component.ts
```diff
@@ -68,14 +66,12 @@ export class AppComponent implements OnInit {
...
-      console.log('HMR enabled');
-    }
-  }
   ngOnInit(): void {
-    this.store.dispatch(new AddTodo('ngOnInit todo'));
+    const payload = 'ngOnInit todo';
+    const state: string[] = this.store.selectSnapshot(TodoState);
+    if (!state.includes(payload)) {
+      this.store.dispatch(new AddTodo(payload));
+    }
```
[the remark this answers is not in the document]
splincode: Past author of pull request
https://github.com/ngxs/store/pull/698/files#diff-bbb25ab674d66eec8397551726ba20d5R33

force state overwriting each time it is updated
it's bad and I corrected it

integration/app/todo.state.ts
```diff
@@ -30,8 +30,12 @@ export class TodoState implements NgxsOnInit {
...
   }
 
-  ngxsOnInit(ctx: StateContext<string[]>) {
-    ctx.setState(['NgxsOnInit todo']);
+  ngxsOnInit({ getState, setState }: StateContext<string[]>) {
+    const state: string[] = getState();
+    const payload = 'NgxsOnInit todo';
+    if (!state.includes(payload)) {
+      setState([...state, payload]);
+    }
```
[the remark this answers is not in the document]
splincode: this is not necessary, this is a bug that the previous author introduced

packages/hmr-plugin/src/symbols.ts
splincode: We need to understand that we are returning a state

integration/app/todo.state.ts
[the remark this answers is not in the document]
markwhitfeld: Does the ngxsOnInit get run after the state has been restored using HMR?
If so, then they need some way of determining if they need to do what they have intended to do on a fresh state.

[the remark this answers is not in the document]
splincode: No, I can return these changes in the app.component file, but this is a bug of the previous author, who added a test for ssr

there is a hmr life cycle, and I believe that there is no need to complicate and write more functions to determine what has changed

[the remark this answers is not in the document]
markwhitfeld: I think we need to think through the potential use cases of how someone might be using the `ngxsOnInit` and how it applies to HMR. For example they could be using it to set up some initial state through some dispatched events. In this case they may want to prevent this from happening if loading as a result of HMR.
On the other hand the user could be starting a timer to refresh a security token in a specific interval. They would want this code to execute, HMR or not.

The decisions above need to be made if the `ngxsOnInit` is run AFTER the state is restored by HMR. But, if the `ngxsOnInit` is run BEFORE the state is restored by HMR then it becomes a lot simpler. Hopefully my question make sa bit more sense given this context...

> Does the ngxsOnInit get run after the state has been restored using HMR?
> If so, then they need some way of determining if they need to do what they have intended to do on a fresh state.

So I guess what I am trying to say is that we have one of 2 options: 
1. Ensure that the HMR state is restored only after the `ngxsOnInit` methods are run
2. Otherwise we need to provide a way for the user to query if HMR has restored state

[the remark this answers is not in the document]
eranshmil: And this is the reason I stopped working on my own implementation a few days ago.
The most interesting case is:
- Access a route that lazy load a state and make a change in this state.
- Move to other route and make a change in the files to activate the HMR.
- Wait for the HMR to finish reloading.
- Enter the same lazy loaded route.

splincode:
@markwhitfeld @eranshmil I managed to ensure that we kept state during the open session (browser tabs) when reloading lazy modules

docs/plugins/hmr.md
```diff
@@ -145,6 +124,30 @@ if (environment.hmr) {
...
+### Update src/app/app.module.ts to manage the state in HMR lifecycle:
+
+```ts
+import { StateOperations } from '@ngxs/store';
+import { NgxsHmrLifeCycle, NgxsStoreSnapshot } from '@ngxs/hmr-plugin';
+
+@NgModule({ .. })
+export class AppBrowserModule implements NgxsHmrLifeCycle<NgxsStoreSnapshot> {
+
+  public hmrNgxsStoreOnInit(ctx: StateOperations<NgxsStoreSnapshot>, snapshot: NgxsStoreSnapshot) {
```
markwhitfeld:
```suggestion
  public hmrNgxsStoreOnInit(ctx: StateContext<NgxsStoreSnapshot>, snapshot: NgxsStoreSnapshot) {
```
Could we rather use a `StateContext` as opposed to a `StateOperations`?
`StateContext` is already public, has a `patchState` method and would already be familiar to the users too.
I would prefer that `StateOperations` does not become part of the public API because its form is likely to change in the future.

docs/plugins/hmr.md
```diff
@@ -145,6 +124,30 @@ if (environment.hmr) {
...
+    console.log('[NGXS HMR] Current state', ctx.getState());
+    console.log('[NGXS HMR] Previous state', snapshot);
+    ctx.setState({ ...ctx.getState(), ...snapshot });
+  }
+
+  public hmrNgxsStoreBeforeOnDestroy(ctx: StateOperations<NgxsStoreSnapshot>): NgxsStoreSnapshot  {
```
markwhitfeld:
```suggestion
  public hmrNgxsStoreBeforeOnDestroy(ctx: StateContext<NgxsStoreSnapshot>): NgxsStoreSnapshot  {
```

integration/app/app.browser.module.ts
```diff
@@ -18,4 +20,16 @@ import { AppModule } from './app.module';
   ],
   providers: [{ provide: 'ORIGIN_URL', useValue: location.origin }]
 })
-export class AppBrowserModule {}
+export class AppBrowserModule implements NgxsHmrLifeCycle<NgxsStoreSnapshot> {
+  public hmrNgxsStoreOnInit(ctx: StateOperations<NgxsStoreSnapshot>, snapshot: NgxsStoreSnapshot) {
```
markwhitfeld:
```suggestion
  public hmrNgxsStoreOnInit(ctx: StateContext<NgxsStoreSnapshot>, snapshot: NgxsStoreSnapshot) {
```

integration/app/app.browser.module.ts
```diff
@@ -18,4 +20,16 @@ import { AppModule } from './app.module';
...
+    console.log('[NGXS HMR] Current state', ctx.getState());
+    console.log('[NGXS HMR] Previous state', snapshot);
+    ctx.setState({ ...ctx.getState(), ...snapshot });
+  }
+
+  public hmrNgxsStoreBeforeOnDestroy(ctx: StateOperations<NgxsStoreSnapshot>): NgxsStoreSnapshot {
```
markwhitfeld:
```suggestion
  public hmrNgxsStoreBeforeOnDestroy(ctx: StateContext<NgxsStoreSnapshot>): NgxsStoreSnapshot {
```

integration/app/app.browser.module.ts
```diff
@@ -18,4 +20,16 @@ import { AppModule } from './app.module';
   ],
   providers: [{ provide: 'ORIGIN_URL', useValue: location.origin }]
 })
-export class AppBrowserModule {}
+export class AppBrowserModule implements NgxsHmrLifeCycle<NgxsStoreSnapshot> {
+  public hmrNgxsStoreOnInit(ctx: StateOperations<NgxsStoreSnapshot>, snapshot: NgxsStoreSnapshot) {
+    console.log('[NGXS HMR] Current state', ctx.getState());
+    console.log('[NGXS HMR] Previous state', snapshot);
+    ctx.setState({ ...ctx.getState(), ...snapshot });
```
markwhitfeld:
```suggestion
    ctx.patchState(snapshot);
```

packages/hmr-plugin/index.ts
```diff
@@ -0,0 +1,4 @@
+/**
+ * The public api for consumers of @ngxs/logger-plugin
```
markwhitfeld:
```suggestion
 * The public api for consumers of @ngxs/hmr-plugin
```

packages/hmr-plugin/src/hmr-bootstrap.ts
```diff
@@ -0,0 +1,28 @@
...
+  bootstrap().then((ref: NgModuleRef<any>) => (ngModule = hmrDoBootstrap(ref)));
+
+  module.hot.dispose(() => {
+    if (autoClearLogs) {
+      console.clear();
+      console.log('[NGXS HMR] clear old logs...');
+    }
+
+    const snapshot: Partial<NgxsStoreSnapshot> = hmrBeforeOnDestroy<NgxsStoreSnapshot>(ngModule);
+    setStateInHmrStorage(snapshot);
```
markwhitfeld: Small refactor...
These 2 lines could also be moved to the './internal' file as a `hmrDoDispose` method (or some other similar name).
Then the logic about the storage can remain encapsulated in the one file.
Potentially even the `validateExistingHmrStorage` could be called from within './internal' too (maybe in the bootstrap/dispose).

packages/hmr-plugin/src/internal.ts
```diff
@@ -0,0 +1,51 @@
...
+
+import { HmrNgxsStoreOnDestroyFn, HmrNgxsStoreOnInitFn, NGXS_HMR, NgxsStoreSnapshot } from './symbols';
+import { StateOperations } from '@ngxs/store/src/internal/internals';
+
+export function hmrDoBootstrap<T = NgxsStoreSnapshot>(ref: NgModuleRef<any>): NgModuleRef<any> {
+  const hmrNgxsStoreOnInitFn: HmrNgxsStoreOnInitFn<T> = ref.instance[NGXS_HMR.hmrNgxsStoreOnInit];
+
+  if (typeof hmrNgxsStoreOnInitFn === 'function') {
+    const stateOperation: StateOperations<NgxsStoreSnapshot> = getStateOperations(ref);
+    stateOperation && hmrNgxsStoreOnInitFn(getStateOperations(ref), getStateFromHmrStorage());
```
markwhitfeld:
```suggestion
    stateOperation && hmrNgxsStoreOnInitFn(stateOperation, getStateFromHmrStorage());
```
use the variable from above... as mentioned in another comment this should rather be a `StateContext` reference too.

packages/hmr-plugin/src/internal.ts
```diff
@@ -0,0 +1,51 @@
...
+  return ref;
+}
+
+export function hmrBeforeOnDestroy<T = NgxsStoreSnapshot>(ref: NgModuleRef<any>): Partial<T> {
+  let resultSnapshot: Partial<T> = {};
+  const hmrNgxsStoreOnDestroyFn: HmrNgxsStoreOnDestroyFn<T> = ref.instance[NGXS_HMR.hmrNgxsStoreBeforeOnDestroy];
+
+  if (typeof hmrNgxsStoreOnDestroyFn === 'function') {
+    const stateOperation: StateOperations<NgxsStoreSnapshot> = getStateOperations(ref);
+    resultSnapshot = (stateOperation && hmrNgxsStoreOnDestroyFn(getStateOperations(ref))) || resultSnapshot;
```
markwhitfeld:
```suggestion
    resultSnapshot = (stateOperation && hmrNgxsStoreOnDestroyFn(stateOperation)) || resultSnapshot;
```
same as above

packages/hmr-plugin/src/internal.ts
```diff
@@ -0,0 +1,51 @@
...
+  return internalFactory && internalFactory.getRootStateOperations();
+}
+
+/**
+ * Session storage: max size - 5 MB, in future need usage IndexDB (50MB)
+ * Session storage is used so that lazy modules can also be updated.
+ */
+export function validateExistHmrStorage() {
+  const hmrStorageDoesNotExist = !(sessionStorage.getItem(NGXS_HMR.SNAPSHOT_KEY) || null);
+  hmrStorageDoesNotExist && sessionStorage.setItem(NGXS_HMR.SNAPSHOT_KEY, JSON.stringify({}));
```
markwhitfeld: An `if` statement might make this code read easier.

packages/hmr-plugin/src/symbols.ts
```diff
@@ -0,0 +1,31 @@
...
+  SNAPSHOT_KEY = '__NGXS_HMR_SNAPSHOT__'
+}
+
+export interface NgxsHmrLifeCycle<T = NgxsStoreSnapshot> {
+  /**
+   * hmrNgxsStoreOnInit is called when the AppModule on init
+   * @param context - current state from Store
+   * @param snapshot - previous state from Store after last hmr on destroy
+   */
+  hmrNgxsStoreOnInit(context: StateOperations<T>, snapshot: Partial<T>): void;
```
markwhitfeld:
```suggestion
  hmrNgxsStoreOnInit(context: StateContext<T>, snapshot: Partial<T>): void;
```

packages/hmr-plugin/src/symbols.ts
```diff
@@ -0,0 +1,31 @@
...
+
+  /**
+   * hmrNgxsStoreOnInit is called when the AppModule on destroy
+   * @param context - current state from Store
+   */
+  hmrNgxsStoreBeforeOnDestroy(context: StateOperations<T>): T;
```
markwhitfeld:
```suggestion
  hmrNgxsStoreBeforeOnDestroy(context: StateContext<T>): T;
```

packages/hmr-plugin/src/symbols.ts
```diff
@@ -0,0 +1,31 @@
...
+}
+
+export type HmrNgxsStoreOnInitFn<T> = (context: StateOperations<T>, snapshot: Partial<T>) => void;
```
markwhitfeld: Did you know that you can alternatively refer to this signature like this:
```TS
const hmrNgxsStoreOnInitFn: NgxsHmrLifeCycle<T>['hmrNgxsStoreOnInit'] = ref.instance[NGXS_HMR.hmrNgxsStoreOnInit];
```
So, you could avoid having to expost this signature again and just use the interface method signature that is already exposed.
... same for the on destroy method.

packages/store/src/plugin_api.ts
```diff
@@ -1,5 +1,7 @@
 export { NgxsModule } from './module';
 export { NGXS_PLUGINS, NgxsPlugin, NgxsNextPluginFn } from './symbols';
+export { InternalStateOperations } from './internal/state-operations';
+export { StateOperations } from './internal/internals';
```
markwhitfeld: I would prefer that we didn't expose these internals.

splincode:
sorry but i don't know how to do this i tried (for using StateContext)

packages/hmr-plugin/src/internal.ts
markwhitfeld: Fixed in my recent commit to this branch.

packages/hmr-plugin/src/symbols.ts
markwhitfeld: Fixed in my recent commit to this branch.

packages/hmr-plugin/src/hmr-bootstrap.ts
markwhitfeld: Fixed in my recent commit to this branch.

packages/hmr-plugin/src/internal.ts
markwhitfeld: Fixed in my recent commit to this branch.

packages/hmr-plugin/src/internal.ts
markwhitfeld: Fixed in my recent commit to this branch.

markwhitfeld:
Got the State Context stuff working but getting some strange typings issues now.

markwhitfeld:
Ok, sorted out the typings issues.
Regarding the lazy loaded states, do you think that we can defer that fix to another PR or should we solve it here? This PR is startng to get quite large.
What we have here does work for non-lazy loaded states.
We won't publish the package until the lazy loaded issue is resolved.
The other thing that needs to be addressed is that this plugin has no tests.

docs/plugins/hmr.md
markwhitfeld: Done in latest commits to this branch.

docs/plugins/hmr.md
markwhitfeld: Done in latest commits to this branch.

integration/app/app.browser.module.ts
markwhitfeld: Done in latest commits to this branch.

integration/app/app.browser.module.ts
markwhitfeld: Done in latest commits to this branch.

packages/hmr-plugin/src/symbols.ts
markwhitfeld: Done in latest commits to this branch.

packages/hmr-plugin/src/symbols.ts
markwhitfeld: Done in latest commits to this branch.

packages/store/src/plugin_api.ts
markwhitfeld: Done in latest commits to this branch.

integration/app/app.browser.module.ts
markwhitfeld: Done in latest commits to this branch.

eranshmil:
> Ok, sorted out the typings issues.
> Regarding the lazy loaded states, do you think that we can defer that fix to another PR or should we solve it here? This PR is startng to get quite large.
> What we have here does work for non-lazy loaded states.
> We won't publish the package until the lazy loaded issue is resolved.
> The other thing that needs to be addressed is that this plugin has no tests.

Yeah, probably it would be better to separate to another PR.

Issue #710, opened 2018-12-14
feat(ofAction): ofActionCompleted


Newbie012:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[ ] Performance issue
[X] Feature request
[ ] Documentation issue or request
[ ] Support request => https://github.com/ngxs/store/blob/master/CONTRIBUTING.md
[ ] Other... Please describe:
</code></pre>

## Current behavior

When I just want to know when an action has been ended (no matter if it's successful. errored or canceled), I need to merge all of them into one observable, and then continue. 

## Expected behavior

A method called `ofActionComplete()` which will emit after an action has been completed no matter if it was successful, errored or canceled.

## What is the motivation / use case for changing the behavior?

Short example - I have a form with a `loading` indicator. `onDispatch` will set it to true. After the action has been completed, I would like to set the `loading` indicator back to false.

## Environment

<pre><code>
Libs:
- @angular/core version: 3.3.0
- @ngxs/store version: 3.3.0

For Tooling issues:
- Node version: XX  
- Platform:

eranshmil:
I agree that this feature should be added, but in your use case I prefer creating isLoading key in the state and using it with an async pipe.

markwhitfeld:
@Newbie012 Yes! Thank you for raising this. It was supposed to be added in v3 but it never happened.

It should rather be called `ofActionCompleted` to match the tense of the other ofAction methods and also to avoid a conflict with the `ofActionComplete` from v2 that was renamed to `ofActionSuccessful` v3.

Would you like to have a go at doing a PR for adding `ofActionCompleted`?
You will need to also add some tests to `packages\store\tests\action.spec.ts`

Newbie012:
@eranshmil According to the last 2 apps I built, I had to do multiple load indicators. Therefore, I moved these indicators to the component level (I made a small state-management system in my components).

@markwhitfeld As I mentioned in the PR, I agree with you. Although, I think that `AuthStatus.Completed` should be kept.

Issue #518, opened 2018-07-23
ofActionErrored 


torbonaut:
When observing errored Actions it would be useful to catch the error that made the action fail. Can the error be provided?

### Versions

```
* ngxs: 3.0.1
* @angular/core: 6.0.3
```

### Desired behavior

Error available within ofActionErrored

splincode:
Add stackblitz for repro

IlCallo:
This would a really good idea.
Use case: login which can fail for multiple reason, so there could be multiple `throw` statement inside a single action, each with a different message.
When listening to the Action stream it's necessary to understand _which_ error are we dealing with.

Ex.
```typescript
    this.actions.pipe(
      ofActionErrored(Login)
    ).subscribe((action, error) => console.log(error));
```

IlCallo:
I checked the code, it seems to me that this is actually just a matter of exposing the full `ActionContext` instead of mapping it to just the action, changing this

https://github.com/ngxs/store/blob/4ba1032c4b99bd52254ea804344512c5382ba1d1/packages/store/src/of-action.ts#L57

removing altogether `mapAction`.

Or, if you _really_ want to hide the status label, it's just a matter of changing the mapping there

https://github.com/ngxs/store/blob/4ba1032c4b99bd52254ea804344512c5382ba1d1/packages/store/src/of-action.ts#L69-L71

maybe like

```typescript
function mapAction() {
  return map(({action, error}: ActionContext) => { action, error );
}
```

@markwhitfeld shall I make a PR?

garthmason:
This would be very useful - any chance it'll make it in the next release?

markwhitfeld:
@splincode  Any chance you could have a look at how we could do this without introducing a breaking change?

splincode:
@markwhitfeld I will try to see soon

markwhitfeld:
This will be achievable using the `ofActionCompleted` pipe when v3.4 is released.
Use the `@dev` version now to try it out. 
See this comment for the signature of what is returned by this pipe: https://github.com/ngxs/store/issues/710#issuecomment-447647487
This will be in the docs once PR #726 is merged

IlCallo:
🎉🎉

Even though, the semplicity of returning the error via ofActionErrored would be the best.

If you consider this to be a breaking change,  can we hope to see it in version 4.0 maybe?

PR #712, opened 2018-12-16
feat(of-action): implement ofActionCompleted

Newbie012:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: https://github.com/ngxs/store/issues/710
https://github.com/ngxs/store/issues/518

## What is the new behavior?

* `ofActionCompleted`: triggers when an action has been completed whether it was successful or not

```
[ ] Yes
[x] No
```

### DIFF
--- a/packages/store/src/actions-stream.ts
+++ b/packages/store/src/actions-stream.ts
@@ -10,7 +10,8 @@ export const enum ActionStatus {
   Dispatched = 'DISPATCHED',
   Successful = 'SUCCESSFUL',
   Canceled = 'CANCELED',
-  Errored = 'ERRORED'
+  Errored = 'ERRORED',
+  Completed = 'COMPLETED'
 }
 
 export interface ActionContext {
--- a/packages/store/src/internal/state-factory.ts
+++ b/packages/store/src/internal/state-factory.ts
@@ -131,8 +131,7 @@ export class StateFactory {
           this.invokeActions(this._actions, action!).pipe(
             map(() => <ActionContext>{ action, status: ActionStatus.Successful }),
             defaultIfEmpty(<ActionContext>{ action, status: ActionStatus.Canceled }),
-            catchError(error => of(<ActionContext>{ action, status: ActionStatus.Errored, error })),
-            finalize(() => this._actions.next(<ActionContext>{ action, status: ActionStatus.Completed }))
+            catchError(error => of(<ActionContext>{ action, status: ActionStatus.Errored, error }))
           )
         )
       )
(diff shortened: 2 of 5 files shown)


packages/store/src/internal/state-factory.ts
```diff
@@ -131,7 +131,8 @@ export class StateFactory {
           this.invokeActions(this._actions, action!).pipe(
             map(() => <ActionContext>{ action, status: ActionStatus.Successful }),
             defaultIfEmpty(<ActionContext>{ action, status: ActionStatus.Canceled }),
-            catchError(error => of(<ActionContext>{ action, status: ActionStatus.Errored, error }))
+            catchError(error => of(<ActionContext>{ action, status: ActionStatus.Errored, error })),
```
splincode: Could you add a test for this case?

splincode: Could you add a test for this case?

splincode: https://github.com/ngxs/store/issues/518

packages/store/src/internal/state-factory.ts
```diff
@@ -131,7 +131,8 @@ export class StateFactory {
...
+            finalize(() => this._actions.next(<ActionContext>{ action, status: ActionStatus.Completed }))
```
markwhitfeld: I don't think we should add a new lifecycle event, but rather filter for any of the existing ones.

packages/store/src/actions-stream.ts
```diff
@@ -10,7 +10,8 @@ export const enum ActionStatus {
   Dispatched = 'DISPATCHED',
   Successful = 'SUCCESSFUL',
   Canceled = 'CANCELED',
-  Errored = 'ERRORED'
+  Errored = 'ERRORED',
+  Completed = 'COMPLETED'
```
markwhitfeld: We don't need to define a new completed status, the existing ones infer completion

splincode: @markwhitfeld I do not agree with you, because COMPLETED status does not promise that he is successful or failed

packages/store/src/internal/state-factory.ts
splincode: On the contrary, thanks to such processing, we solve the problem https://github.com/ngxs/store/issues/518

packages/store/src/actions-stream.ts
Newbie012: Even tho they infer, I made the COMPLETED status for the readability. It's not crucial, I can remove it. But I my suggestion is to keep it.

packages/store/src/internal/state-factory.ts
Newbie012: @markwhitfeld Agreed.
@splincode the new `ofActionCompleted` should solve your given problem.

packages/store/src/actions-stream.ts
markwhitfeld: @splincode These statuses represent the possible lifecycle status values. If we leave this in here then we should have 2 events added to the actions stream when completion happens. Not something that need to be in that stream.

@newbie012 you can leave it in here and I'll follow up your PR with another commit that sorts it the statuses and optimises the performance of the checks in the ofAction pipe. 👍

PR #714, opened 2018-12-17
style(prettier): fix format script

eranshmil:
## PR Checklist
Please check if your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://github.com/ngxs/store/blob/master/CONTRIBUTING.md#commit
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

```
[ ] Bugfix
[ ] Feature
[x] Code style update (formatting, local variables)
[ ] Refactoring (no functional changes, no api changes)
[ ] Build related changes
[ ] CI related changes
[ ] Documentation content changes
[ ] Other... Please describe:
```

## What is the current behavior?
Without a quote in the package.json, only two levels of files are being formatted.

## What is the new behavior?
All relevant files are being formatted.

## Does this PR introduce a breaking change?
```
[ ] Yes
[x] No
```

PR #717, opened 2018-12-18
Revert "style(prettier): fix format script"

markwhitfeld:
Reverts ngxs/store#714

This new single quoted form does not work in windows.

### DIFF
--- a/package.json
+++ b/package.json
@@ -77,7 +77,7 @@
     "// - UTILS": "Utility Scripts",
     "lint": "ng lint",
     "docs": "typedoc packages/store/src --options typedoc.json",
-    "format": "prettier --write '{packages,integration}/**/*.ts'"
+    "format": "prettier --write {packages,integration}/**/*.ts"
   },
   "private": true,
   "devDependencies": {


markwhitfeld:
@eranshmil What OS are you on?
I am using windows and the single quotes broke the command:
```
> yarn format
yarn run v1.12.3
$ prettier --write '{packages,integration}/**/*.ts'
[error] No matching files. Patterns tried: '{packages,integration}/**/*.ts' !**/node_modules/** !./node_modules/** !**/.{git,svn,hg}/** !./.{git,svn,hg}/**
error Command failed with exit code 2.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```
When I removed the single quotes it did the formatting perfectly n-levels deep.
I also tried double quotes as opposed to single quotes and it also works fine.
Maybe you could try double quotes in your OS? If that works then maybe we could go with that.

...Merging this PR now to avoid the broken state.

eranshmil:
I'm using macOS.
I've found this issue https://github.com/prettier/prettier/issues/4804, and it's working the following way:
```json
"format": "prettier --write \"{packages,integration}/**/*.ts\""
```

Issue #545, opened 2018-08-15
Proposal: State Operators


markwhitfeld:
### Versions

```
* ngxs: v3.2.0
* @angular/core: v6.1
```

### Background
The NGXS patchState method is used to do immutable object updates to the container state slice without the typical long-handed syntax. This is very neat and convenient because you do not have to use the getState and setState as well as the Object.assign(…)or the spread operator to update the state. The patchState method only offers a shallow patch and as a result is left wanting in more advanced scenarios. 
What if it is possible to make the patchState method extensible to create a more advanced yet simple mechanism for state updates?… Let me introduce patch operators…

### Proposal
The basic idea of this proposal is that we could describe the modifications to the state using curried functions that are given any inputs that they need to describe the change and are assigned to the part of the state tree that they should be applied to. The curried function is then run using the state slice that they are assigned to and the returned value is used as the new value for that state slice. 

The function that produces the curried function that is still awaiting the slice of state to be applied to is called a patch operator. The type of this curried function could be defined as follows:

```TS
type op<T> = (state: T) => T;
```

We can make a number of these patch operator functions in order to help with common scenarios for state manipulation. Some ideas are:

- `patch<T>(value: T) : op<T>`
- `set<T>(value: T) : op<T>`
- `iif<T>(condition: Predicate<T>, then: op<T>, else: op<T>) : op<T>`

And some array operators:
- `patchItem<T>(selector: number | Predicate<T>, operator: op<T>) : op<T[]>`
- `removeItem<T>(selector: number | Predicate<T>) : op<T[]>`
- `insertItem<T>(value: T, beforePosition? : number) : op<T[]>`

### Example:

We could express some modifications to a state as follows:
```TS
ctx.patchState({
  greeting: 'Howzit',
  token: patch({
    lastUpdate: new Date(),
    active: true,
    origin: iif( origin => !origin.ip, patch({ ip: '192.168.1.1'})
  }),
  roles: [
    removeItem(3),
    insertItem({id:55, name: 'Admin'}),
    patchItem( x => x.id === 64, {name:'Editor'})
  ]
});
```
Please add any other ideas for operators to the comments.
Feedback welcome!

PS. Related Meduim posts here: 
Initial proposal: https://medium.com/@mark.whitfeld/ngxs-proposal-patch-operators-75b24d309b91
Announcement before v3.4: https://medium.com/ngxs/ngxs-state-operators-8b339641b220

garthmason:
[...]

PR #736, opened 2019-01-02
feat: implement patch state operator

markwhitfeld:
## PR Checklist

## PR Type

```
[ ] Bugfix
[X] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: #545
Now that PR #635 has been implemented we can start adding the standard state operators.
This `patch` operator is the first of these.

## What is the new behavior?
Introduced the `patch` operator which will allow for deeper patching as well as the ability to easily use state operators for properties within a state object.

I have added a use of this operator to the Integration example application (albeit a bit contrived to have such a deep state).  _Note that I haven't added anything to the docs in this PR so that the docs can be updated on the release of the major version to avoid confusing the users._

I have added a sub package called `@ngxs/store/operators` to encapsulate these operatros into their own area. The alternative to this subpackage would be to add them to a separate package called `@ngxs/state-operators` but this would now give the users an additional install step to take to use these operators, which is not ideal (so we are trying out the sub-package approach for now).

**NOTE:** This PR is to allow us to explore how this sub-package feels to use in the dev branch before committing to an approach.

```
[ ] Yes
[X] No
```


markwhitfeld:
Tried to fix the code climate issues, but they are caused by differences between what prettier and code climate think are good practice. I have gone with the prettier style enforced during commit.

splincode:
I would also like to have a description in the documentation, for what it may come in handy.

```ts
@Action(SetPrefix)
  public setPrefix({ getState, setState, patchState }: StateContext<TodoStateModel>) {
    setState(
      patch({
        pizza: patch({
          model: patch({
            toppings: (topping: any) => 'Mr. ' + topping
          })
        })
      })
    );
  }
```

What is better than this?

```ts
setState(({ pizza }) => ({ 
    pizza: { 
       model: {  toppings: 'Mr. ' + pizza.model.toppings } 
    })
);
```

I think it would be better to do so merge operator

```ts
// state = { a: { a: 1, b: 2 } }
// b = { a: { c: 3} }
setState(state => merge(state, b)) // { a: { a: 1, b: 2, c: 2 } }
```

```ts
setState(state => replace(state, 'pizza.model.toppings', (toppings) => 'Mr' + toppings))
```

integration/app/store/todos/todos.state.ts
```diff
@@ -22,9 +24,15 @@ export class TodosState {
...
-    patchState({ pizza: { model: { toppings: 'Mr. ' + pizza1 } } });
+    setState(
+      patch({
+        pizza: patch({
+          model: patch({
+            toppings: (topping: any) => 'Mr. ' + topping
+          })
+        })
+      })
+    );
```
deebloo: how well does the type inference work when combining patches like this?

splincode: I think bad, but maybe we can usage generic for patch<T>

markwhitfeld: @deebloo It works very well. It took me 3 hours to get the operator done with tests and about 12 frustrating hours to get the type checking working nicely. Have a look at the bigger examples in the tests. 
You can supply a type T here for the patch operator. Unfortunately this integration example lacked decent types and is deeply nested, not the ideal use case for the patch operator.

I agree that operators like those suggested by @splincode would be better suited for this particular scenario.

splincode:
@markwhitfeld Unfortunately, I can not find an application and I have never seen people do this.

```ts
const original: Model = {
             a: 1,
             b: {
               hello: {
                 person: {
                   name: 'you'
                 }
               },
               goodbye: {
                 person: {
                   name: 'Mark'
                 }
               }
             }
           }
```

I think you just need to use the operator merge

```ts
function mergeDeep(...objects) {
  const isObject = obj => obj && typeof obj === 'object';
  
  return objects.reduce((prev, obj) => {
    Object.keys(obj).forEach(key => {
      const pVal = prev[key];
      const oVal = obj[key];
      
      if (Array.isArray(pVal) && Array.isArray(oVal)) {
        prev[key] = pVal.concat(...oVal);
      }
      else if (isObject(pVal) && isObject(oVal)) {
        prev[key] = mergeDeep(pVal, oVal);
      }
      else {
        prev[key] = oVal;
      }
    });
    
    return prev;
  }, {});
}
```

image

Issue #574, opened 2018-09-21
State composition, why default state is not inherited?


gremo:
Maybe I misunderstood the the state composition feature. I have a `GenericListingStateModel` which provides `defaults` (model omitted for brevity):

```typescript
@State<GenericListingStateModel<T>>({
  name: 'genericList',
  defaults: {
    entities: {},
    currentPage: 1,
    currentLimit: 10,
    currentPageResults: [],
  }
})
export class GenericListingState<T> { }
```

Then I use the state composition to inherit the state model (and the defualts):

```typescript
@State<GenericListingStateModel<Customer>>({ name: 'customers' })
export class CustomerState extends GenericListingState<Customer> { }
```
Why the state of `CustomerState` **is empty and doesn't inherit all the defaults** defined in the `GenericListingState`?

matulkum:
I think because you overwrote the @State decorator and if I am not mistaken there is no inheritance in decorators, therefore it would not be possible to inherit the _defaults_ .

gremo:
Well, without overriding `@State` I can't assigna name to my `CustomerState `. So... what's the point of state composition as explained [here](https://ngxs.gitbook.io/ngxs/advanced/composition)?

matulkum:
you still can, you just have to apply the whole decorator again. So just pass in your defaults into the decorator of your inheriting class again.

Again: I did not use inheritance in that case myself. Someone please correct me if I am stating something false ;)

gremo:
@matulkum duplcation of code. Defaults must be applied both to the parent class and to the child class (correct me if I'm wrong).

matulkum:
yeah.. kind of. But it is not so bad. I guess you could do it like this:
```js
const defaults = {
  entities: {},
  currentPage: 1,
  currentLimit: 10,
  currentPageResults: [],
}

@State<GenericListingStateModel<T>>({
  name: 'genericList',
  defaults,
})
export class GenericListingState<T> { }

@State<GenericListingStateModel<Customer>>({
  name: 'customers',
  defaults,
})
export class CustomerState extends GenericListingState<Customer> { }
```

juliusstoerrle:
@gremo you mentioned a state composition feature, could you (maybe by providing a link to the docs), show which feature you ment?

gremo:
@juliusstoerrle https://ngxs.gitbook.io/ngxs/advanced/composition

juliusstoerrle:
Thanks, I never saw that feature until today..
[...]

Issue #494, opened 2018-07-12
[Discussion] Children SubStates shared between two top-level States


IlCallo:
This is not really a feature or bug request, nor a technical question.
Is more a "what do you think about..." thing.

My use case:
Among other parts, I'm using NGXS to keep track of the currently logged user preferences.
Other than that, I'm also keeping track of the **default** preferences for that user's role which have the same structure and is used for "reset to default" functionalities (not total, it's a reference to do a diff with single sub-objects).

Currently my code is like this

```typescript
/* State */

const preferencesSubStates = [
    Module1PreferencesState,
    Module2PreferencesState,
];

@State<AppPreferences>({
    name: 'preferences',
    children: preferencesSubStates,
})
export class PreferencesState {
    // Things
}

@State<AppPreferences>({
    name: 'rolePreferences',
    children: preferencesSubStates,
})
export class RolePreferencesState {
    // More things
}
```

```typescript
/* Module */

@NgModule({
  imports: [
    NgxsModule.forFeature([
      PreferencesState,
      RolePreferencesState,
      Module1PreferencesState,
      Module2PreferencesState,
    ]),
    NgxsStoragePluginModule.forRoot({
      key: [ 'preferences', 'rolePreferences'],
    }),
  ],
})
export class MyModule { }
```

This actually works pretty well, but then I asked myself: "preferences updates is working, but how does it know about which one of the top-level State must be used for Modules `@Select/@Selectors`?!"

Turns out the answer is the declaration order inside `NgxsModule.forFeature` (the first wins).

Would you like to enhance support for this scenario? Do think this is a proper use of your library or too much an edge case? Generally speaking, what do you think about this pattern?
Obviously I cannot use `@Select` for the second top-level state and I'm forced to use inline functions or strings, which is a pity but I can live with it.

This could be adressable with parameters on `@Select` decorators (I saw someone already proposed those), but there are probably other possibilities to support this scenario.

markwhitfeld:
PS. with the current `@dev` version you could actually have a reasonable selector using a new construct I added last week. `createSelector` allows for the creation of a selector at run time. So you could have this as a selector:
```typescript
@State<Module1Preferences>({
    name: 'module1Pref'
})
export class Module1PreferencesState {

    @Selector()
    getPreferenceFrom(parentClass: any) {
        return createSelector([parentClass], (parent) => parent['module1Pref'] );
    }

    // Even more things
}

@Component( /* ... */)
class SomeComponent {

[...]

PR #750, opened 2019-01-06
feat(store): inheritance state options

splincode:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

https://github.com/ngxs/store/issues/574

## What is the new behavior?

```ts
@State({
  name: 'a',
  defaults: { value: 'a' }
})
class MyState {}

@State<MyStateModel>({
  name: 'b'
})
class MyOtherState extends MyState {} // can provide defaults: { value: 'a' }
```

```
[ ] Yes
[x] No
```

### DIFF
--- a/docs/advanced/composition.md
+++ b/docs/advanced/composition.md
@@ -3,11 +3,16 @@ You can compose multiple stores together using class inheritance. This is quite 
 
 ```TS
 @State({
-  name: 'zoo'
+  name: 'zoo',
+  defaults: {
+    type: null 
+  }
 })
 class ZooState {
   @Action(Eat)
-  eat() {}
+  eat(ctx: StateContext) {
+    ctx.setState({ type: 'eat' });
+  }
 }
 
 @State({
@@ -15,8 +20,11 @@ class ZooState {
 })
 class StLouisZooState extends ZooState {
   @Action(Drink)
-  drink() {}
+  drink() {
+    ctx.setState({ type: 'drink' });
+  }
 }
 ```
 
-Now when `StLouisZooState` is invoked, it will share the actions of the `ZooState`.
+Now when `StLouisZooState` is invoked, it will share the actions of the `ZooState`. 
+Also all state options are inherited.
--- a/docs/advanced/composition.md
+++ b/docs/advanced/composition.md
@@ -20,7 +20,7 @@ class ZooState {
 })
 class StLouisZooState extends ZooState {
   @Action(Drink)
-  drink() {
+  drink(ctx: StateContext) {
     ctx.setState({ type: 'drink' });
   }
 }
--- a/packages/store/src/decorators/state.ts
+++ b/packages/store/src/decorators/state.ts
@@ -1,4 +1,9 @@
-import { ensureStoreMetadata, MetaDataModel, StateClass } from '../internal/internals';
+import {
+  ensureStoreMetadata,
+  MetaDataModel,
+  MutateMetaOptions,
+  StateClass
+} from '../internal/internals';
 import { META_KEY, META_OPTIONS_KEY, StoreOptions } from '../symbols';
 import { StoreValidators } from '../utils/store-validators';
 
@@ -6,11 +11,11 @@ import { StoreValidators } from '../utils/store-validators';
  * Decorates a class with ngxs state information.
  */
 export function State<T>(options: StoreOptions<T>) {
-  function mutateMetaData(
-    meta: MetaDataModel,
-    stateClass: StateClass,
-    optionsWithInheritance: StoreOptions<T>
-  ) {
+  function mutateMetaData({
+    meta,
+    stateClass,
+    optionsWithInheritance
+  }: MutateMetaOptions<T>): void {
     const { children, defaults, name } = optionsWithInheritance;
     StoreValidators.checkCorrectStateName(name);
 
@@ -24,12 +29,12 @@ export function State<T>(options: StoreOptions<T>) {
     meta.name = name;
   }
 
-  return function(target: StateClass) {
+  return (target: StateClass): void => {
     const meta: MetaDataModel = ensureStoreMetadata(target);
     const stateClass: StateClass = Object.getPrototypeOf(target);
     const inheritanceOptions: Partial<StoreOptions<T>> = stateClass[META_OPTIONS_KEY] || {};
     const optionsWithInheritance: StoreOptions<T> = { ...inheritanceOptions, ...options };
-    mutateMetaData(meta, stateClass, optionsWithInheritance);
+    mutateMetaData({ meta, stateClass, optionsWithInheritance });
     target[META_OPTIONS_KEY] = optionsWithInheritance;
   };
 }
--- a/packages/store/src/decorators/state.ts
+++ b/packages/store/src/decorators/state.ts
@@ -4,7 +4,7 @@ import { StoreValidators } from '../utils/store-validators';
 
 interface MutateMetaOptions<T> {
   meta: MetaDataModel;
-  stateClass: StateClass;
+  inheritedStateClass: StateClass;
   optionsWithInheritance: StoreOptions<T>;
 }
 
@@ -12,13 +12,18 @@ interface MutateMetaOptions<T> {
  * Decorates a class with ngxs state information.
  */
 export function State<T>(options: StoreOptions<T>) {
+  function getStateOptions(stateClass: StateClass): StoreOptions<T> {
+    const inheritanceOptions: Partial<StoreOptions<T>> = stateClass[META_OPTIONS_KEY] || {};
+    return { ...inheritanceOptions, ...options };
+  }
+
   function mutateMetaData(params: MutateMetaOptions<T>): void {
-    const { meta, stateClass, optionsWithInheritance } = params;
+    const { meta, inheritedStateClass, optionsWithInheritance } = params;
     const { children, defaults, name } = optionsWithInheritance;
     StoreValidators.checkCorrectStateName(name);
 
-    if (stateClass.hasOwnProperty(META_KEY)) {
-      const parentMeta: Partial<MetaDataModel> = stateClass[META_KEY] || {};
+    if (inheritedStateClass.hasOwnProperty(META_KEY)) {
+      const parentMeta: Partial<MetaDataModel> = inheritedStateClass[META_KEY] || {};
       meta.actions = { ...meta.actions, ...parentMeta.actions };
     }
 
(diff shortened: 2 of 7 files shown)


packages/store/src/decorators/state.ts
```diff
@@ -1,42 +1,28 @@
...
 /**
  * Decorates a class with ngxs state information.
  */
 export function State<T>(options: StoreOptions<T>) {
-  return function(target: any) {
-    const meta = ensureStoreMetadata(target);
+  return function(target: StateClass) {
+    const meta: MetaDataModel = ensureStoreMetadata(target);
+    const targetReference: StateClass = Object.getPrototypeOf(target);
+    const inheritanceOptions: Partial<StoreOptions<T>> = target[META_OPTIONS_KEY] || {};
```
markwhitfeld:
```suggestion
    const inheritanceOptions: Partial<StoreOptions<T>> = targetReference[META_OPTIONS_KEY] || {};
```
Surely we want the options from the prototype?
Also maybe `inheritedStateClass` might be a better name than `targetReference`?

packages/store/src/internal/internals.ts
```diff
@@ -189,7 +200,9 @@ export function buildGraph(stateClasses: StateClass[]): StateKeyGraph {
   const findName = (stateClass: StateClass) => {
     const meta = stateClasses.find(g => g === stateClass);
     if (!meta) {
-      throw new Error(`Child state not found: ${stateClass}`);
+      throw new Error(
+        `Child state not found: ${stateClass}. \nYou may have forgotten to add states to module`
```
markwhitfeld:
```suggestion
        `Child state not found: ${stateClass}. \r\nYou may have forgotten to add states to module`
```

packages/store/src/decorators/state.ts
```diff
@@ -1,42 +1,28 @@
...
+    const { children, defaults, name } = { ...inheritanceOptions, ...options };
```
markwhitfeld: Is it valid to inherit the children? This might lead to a child state share between two top level states which has some caveats. See discussion in issue #494

splincode: we can cover with tests such a case, just give an example

splincode: all properties should be inherited, since the essence of the task in this

markwhitfeld: The `ChildSate` will then 'exist' in multiple places in the tree but will have only one MappedStore representation. This will result in the action handlers only being executed for the first place that it was added and the `store.select(ChildState)` only being able to select from only that one state.
I guess that this is an existing issue so it does not need to be addressed by this PR, but this PR makes it more likely that this issue will be encountered so it pushes the priority up for solving it. This would be encountered when you inherit multiple times from a state that has a child state. I guess that the duplicate check may catch this scenario, but we would have to test it. Could you write a test for this?

packages/store/tests/inheritance.spec.ts
```diff
@@ -0,0 +1,49 @@
...
+    defaults: { value: 'a' },
+    children: [MyChildAState]
+  })
+  class MyState {}
+
+  @State<MyStateModel>({
+    name: 'b',
+    children: [MyChildBState]
+  })
+  class MyOtherState extends MyState {}
```
markwhitfeld: I guess for the inherited children scenario we would have a test that has another variant of this inherited state that does not override the children. The test would essentially just be defining how ngxs currently handles this for now (an error thrown or just silently accept it) - a characterization test.

docs/advanced/composition.md
```diff
@@ -3,20 +3,28 @@ You can compose multiple stores together using class inheritance. This is quite
...
-  name: 'zoo'
+  name: 'zoo',
+  defaults: {
+    type: null 
+  }
 })
 class ZooState {
   @Action(Eat)
-  eat() {}
+  eat(ctx: StateContext) {
```
eranshmil: I think we should use destructing (`{ setState }`) in the docs.

docs/advanced/composition.md
```diff
@@ -3,20 +3,28 @@ You can compose multiple stores together using class inheritance. This is quite
...
+  }
 }
 
 @State({
   name: 'stlzoo'
 })
 class StLouisZooState extends ZooState {
   @Action(Drink)
-  drink() {}
+  drink() {
```
eranshmil: ctx is missing.
Also, see the comment above.

packages/store/src/decorators/state.ts
```diff
@@ -1,42 +1,28 @@
...
+import { META_KEY, META_OPTIONS_KEY, StoreOptions } from '../symbols';
+import { StoreValidators } from '../utils/store-validators';
 
 /**
  * Decorates a class with ngxs state information.
  */
 export function State<T>(options: StoreOptions<T>) {
-  return function(target: any) {
-    const meta = ensureStoreMetadata(target);
+  return function(target: StateClass) {
```
eranshmil: There's a reason why not change that to arrow function?

packages/store/src/decorators/state.ts
```diff
@@ -1,42 +1,28 @@
...
-      };
+    if (targetReference.hasOwnProperty(META_KEY)) {
+      const parentMeta: Partial<MetaDataModel> = targetReference[META_KEY] || {};
+      meta.actions = { ...meta.actions, ...parentMeta.actions };
     }
 
-    meta.children = options.children;
-    meta.defaults = options.defaults;
-    meta.name = options.name;
+    meta.children = children;
```
eranshmil: Isn't it possible to use spread?
```typescript
meta = {
  ...meta,
  children,
  defaults,
  name
}
```

docs/advanced/composition.md
markwhitfeld: There was a desion a while back to remove the destructuring because it is not as familiar to most devs as a normal parameter. See PR #320. We should still use `ctx`.

packages/store/src/decorators/state.ts
splincode: nope, because we need mutate meta

splincode:
@markwhitfeld @eranshmil updated

packages/store/tests/inheritance.spec.ts
```diff
@@ -0,0 +1,95 @@
...
+
+    const store = TestBed.get(Store);
+    const state = store.selectSnapshot(MyState);
+    const otherState = store.selectSnapshot(MyOtherState);
+
+    expect(state).toEqual({ value: 'a', child_a: 'child_a' });
+    expect(otherState).toEqual({ value: 'a', child_b: 'child_b' });
+  });
+
+  it('should be correct first inherited without other states', () => {
```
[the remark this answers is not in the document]
splincode: done

splincode:
@markwhitfeld updated

packages/store/src/decorators/state.ts
```diff
@@ -24,12 +29,12 @@ export function State<T>(options: StoreOptions<T>) {
     meta.name = name;
   }
 
-  return function(target: StateClass) {
+  return (target: StateClass): void => {
```
eranshmil: Because we don't use void on a regular basis, I think we should avoid adding that.

packages/store/src/decorators/state.ts
```diff
@@ -1,16 +1,21 @@
...
-  function mutateMetaData(
-    meta: MetaDataModel,
-    stateClass: StateClass,
-    optionsWithInheritance: StoreOptions<T>
-  ) {
+  function mutateMetaData({
+    meta,
+    stateClass,
+    optionsWithInheritance
+  }: MutateMetaOptions<T>): void {
```
eranshmil: Because we don't use void on a regular basis, I think we should avoid adding that.

packages/store/src/decorators/state.ts
```diff
@@ -1,42 +1,37 @@
...
-      throw new Error(stateNameErrorMessage(options.name));
-    }
+    meta.children = children;
+    meta.defaults = defaults;
+    meta.name = name;
+  }
+
+  return (target: StateClass): void => {
+    const meta: MetaDataModel = ensureStoreMetadata(target);
+    const stateClass: StateClass = Object.getPrototypeOf(target);
```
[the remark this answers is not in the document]
markwhitfeld: It is the prototype of the target class, therefore is it not a class we have inherited from (or extended)? I was trying to avoid the word `parent` because that could refer to the state graph.
`inherited` is the best I can come up with to describe the extends relationship.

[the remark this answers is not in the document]
splincode: done

packages/store/src/decorators/state.ts
splincode: I think it's time for us to use return types everywhere in the future to configure tslint

packages/store/src/decorators/state.ts
splincode: I think it's time for us to use return types everywhere in the future to configure tslint

splincode:
@markwhitfeld update

packages/store/src/decorators/state.ts
```diff
@@ -1,42 +1,42 @@
...
+    return { ...inheritanceOptions, ...options };
+  }
+
+  function mutateMetaData(params: MutateMetaOptions<T>): void {
+    const { meta, inheritedStateClass, optionsWithInheritance } = params;
+    const { children, defaults, name } = optionsWithInheritance;
+    StoreValidators.checkCorrectStateName(name);
+
+    if (inheritedStateClass.hasOwnProperty(META_KEY)) {
+      const parentMeta: Partial<MetaDataModel> = inheritedStateClass[META_KEY] || {};
```
markwhitfeld:
```suggestion
      const inheritedMeta: Partial<MetaDataModel> = inheritedStateClass[META_KEY] || {};
```

packages/store/src/decorators/state.ts
```diff
@@ -1,42 +1,42 @@
...
-
-    // Handle inheritance
-    if (Object.getPrototypeOf(target).hasOwnProperty(META_KEY)) {
-      const parentMeta = Object.getPrototypeOf(target)[META_KEY];
-
-      meta.actions = {
-        ...meta.actions,
-        ...parentMeta.actions
-      };
+  function getStateOptions(stateClass: StateClass): StoreOptions<T> {
```
markwhitfeld:
```suggestion
  function getStateOptions(inheritedStateClass: StateClass): StoreOptions<T> {
```
Rename this parameter too for consistency.

splincode:
@markwhitfeld update

splincode:
@eranshmil ping

PR #751, opened 2019-01-07
fix(store): action completion error type

eranshmil:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

```typescript
export interface ActionCompletion<T = any> {
  action: T;
  result: {
    successful: boolean;
    canceled: boolean;
    error?: Error;
  };
}
```

The most common use case for the ofActionCompleted is to handle an HttpErrorResponse from the backend.

## What is the new behavior?
No need to use hacks like:
```typescript
const {
  error: { message }
} = result.error as any;
```

```
[ ] Yes
[x] No
```

### DIFF
--- a/packages/store/src/operators/of-action.ts
+++ b/packages/store/src/operators/of-action.ts
@@ -3,12 +3,12 @@ import { map, filter } from 'rxjs/operators';
 import { getActionTypeFromInstance } from '../utils/utils';
 import { ActionContext, ActionStatus } from '../actions-stream';
 
-export interface ActionCompletion<T = any> {
+export interface ActionCompletion<T = any, E = Error> {
   action: T;
   result: {
     successful: boolean;
     canceled: boolean;
-    error?: Error;
+    error?: E;
   };
 }
 
@@ -101,11 +101,11 @@ function filterStatus(allowedTypes: FilterMap, allowedStatuses?: FilterMap) {
 function mapActionResult(): OperatorFunction<ActionContext, ActionCompletion> {
   return map(({ action, status, error }: ActionContext) => {
     return <ActionCompletion>{
-      action: action,
+      action,
       result: {
         successful: ActionStatus.Successful === status,
         canceled: ActionStatus.Canceled === status,
-        error: error
+        error
       }
     };
   });


packages/store/src/operators/of-action.ts
```diff
@@ -8,7 +8,7 @@ export interface ActionCompletion<T = any> {
   result: {
     successful: boolean;
     canceled: boolean;
-    error?: Error;
+    error?: any;
```
splincode: why any?

splincode: can we somehow learn how to use the type we need

eranshmil: I don't think so, because you can also have ReferenceError when you expect to something else.

splincode:
```
export interface ActionContext<T = any, U extends Error> {
  status: ActionStatus;
  action: T;
  error?: U;
}
```

eranshmil: Error is an interface, so I guess that supposed to be `U implements Error`.
It might get messy if someone, for some reason, except multiple errors in the same action.

@markwhitfeld What is your preferred way?

markwhitfeld: @eranshmil Is the problem with the Error type that a user may throw a string as an error instead?
(which doesn't extend error)

eranshmil: That might be as a result of a runtime error (such as ReferenceError), a custom thrown error, HttpErrorResponse. (And I guess that there's more use cases)

markwhitfeld: @eranshmil Those usually extend `Error` but I see your point about the casting, and we also cannot guarantee that what was thrown does extend `Error`.
How about something like this:
```TS

export interface ActionCompletion<T = any, TError = Error> {
  action: T;
  result: {
    successful: boolean;
    canceled: boolean;
    error?: TError;
  };
}
```
PS I called it `TError` because it is a bit more descriptive than `U`. Up to you though.

eranshmil: Usually I'd agree with you about type safety, but it might be problematic if an unexpected error has happened.
Does it necessary mean that the error would implement the Error interface?

markwhitfeld: No, in my sample it is not requiring that it extends it, merely that if you don't specify the `TError` then it defaults to be of type `Error`.
ie. this would be totally valid:
```TS
let info: ActionCompleted<LoggedIn, string> = ...
```

splincode: @markwhitfeld I think good case if we usage implements Error, because we understand in advance that this is a type with an error
`export interface ActionCompletion<T = any, E extends Error = Error> {`

markwhitfeld:
@splincode In reference to your comment https://github.com/ngxs/store/pull/751#discussion_r246275737
It is forseeable that an error could be thrown that does not extend the `Error` interface, so this would not be ideal as a type restriction.

Issue #739, opened 2019-01-03
Run outside ngzone


uahic:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[ x] Performance issue
[ ] Feature request
[x] Documentation issue or request
[ ] Support request => https://github.com/ngxs/store/blob/master/CONTRIBUTING.md
[ ] Other... Please describe:
</code></pre>

## Current behavior

NgZone is triggered on each state update

## Expected behavior

No NgZone involvement (optionally)

## What is the motivation / use case for changing the behavior?

I am using NgXS currently only as a model/state for shapes drawn with pixi.js. So basically several arrays of points and whether it is selected or not along with an id for each shape (polygonline). Now I am experimenting whether it is performance-wise OK to update the state when dragging a shape basically with the frequency of mouse-move events (throttled by pixi.js, but still quite alot) or if I need to hold updates back until the dragging ends (which adds more local states and complexity). I've noticed that ngZone gets triggered each time an update occurs, I've tried to wrap all NgXS dispatch and subscribe calls into the runOutsideAngular call but its not working, probably there are some explicit calls in the NgXS code to run within the ngZone? or is it just RxJS itself ?

PIXI.js itself wrapped such that it runs in my application outside of ngZone which works fine

## Environment

<pre><code>
Libs:
- @angular/core version: X.Y.Z
- @ngxs/store version: X.Y.Z

Browser:
- [ x] Chrome (desktop) version XX
- [ ] Chrome (Android) version XX
- [ ] Chrome (iOS) version XX
- [ ] Firefox version XX
- [ ] Safari (desktop) version XX
- [ ] Safari (iOS) version XX
- [ ] IE version XX
- [ ] Edge version XX
 

Linux (Ubuntu 16.04)
Others:

</code></pre>

splincode:
Please add repo or stackblitz for reproduce

uahic:
Here is a stackblitz example 
https://stackblitz.com/edit/ngxs-simple-hgumzb

splincode:
you are right, all events take place in the zone
https://github.com/ngxs/store/blob/master/packages/store/src/store.ts#L45

splincode:
https://github.com/ngxs/store/blob/master/packages/store/src/operators/zone.ts

uahic:
[...]

Issue #487, opened 2018-07-09
Feature request : Option to turn off 'run outside of zone' feature


msegmx:
(from [Stackoverflow](https://stackoverflow.com/questions/51235284/http-requests-made-from-ngxs-state-doesnt-get-detected-by-angular-zone-related))

I'm using ngx-progressbar and it works fine with http request started from within services, components or resolvers.

Note that no manual triggering of the progress bar (via service, etc) during http requests is needed. It is triggered automatically.

Unfortunately it doesn't work as expected when making an http request from within an NGXS State :

[stackblitz](https://stackblitz.com/edit/ngx-progressbar-bat5zp?file=src%2Fapp%2Fngxs.ts)

I created for every case a button :

"Make Request (Dispatch to store, w/o zone)"
This does not work, no progress bar appears (or it appears but hangs and does not complete to 100%)

```
@Action(LoadUrl)
load({ setState }: StateContext<string>) {
    return this.http.get<any>('https://jsonplaceholder.typicode.com/posts/1').pipe(tap(res => console.log(res)));
}
```
"Make Request (Component)"
This does work as expected, progress bar appears

```
@Component({
  selector: 'my-app',
  templateUrl: './app.component.html',
  styleUrls: ['./app.component.css']
})
export class AppComponent {
// ...
  makeRequestComponent() {
    this.http.get<any>('https://jsonplaceholder.typicode.com/posts/1').pipe(tap(res => console.log(res))).subscribe();
  }
}
```
"Make Request (Dispatch to store, w/ zone)"
I asked the developer of ngx-progressbar and he actually found the issue and solved it by wrapping the http call inside zone. This does work, progress bar appears :

```
@State<string>({
  name: 'testState',
  defaults: ''
})
export class TestUrlState {

  constructor(private http: HttpClient, private zone: NgZone) { }

  @Action(LoadUrl)
  load({ setState }: StateContext<string>) {

    this.zone.run(() => {

      this.http.get<any>('https://reqres.in/api/users?delay=2').pipe(
        tap(res => console.log(res))
      ).subscribe();

    });
  }
}
```

Please add an option to turn off the 'run outside of zone' feature.

markwhitfeld:
@amcdnl I think that this is a but different. The ngx-progressbar is expecting the angular HttpInterceptor to be executed in the angular zone but because our actions are executing the http call outside of the zone this does not work correctly.
[...]

PR #761, opened 2019-01-10
perf(store): add `outsideZone` option to the config

arturovt:
## PR Checklist

## PR Type
```
[ ] Bugfix
[x] Feature
[x] Perfomance issue
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

## Current behavior

https://github.com/ngxs/store/issues/739
https://github.com/ngxs/store/issues/487

```
[ ] Yes
[x] No
```

## Motivation
NGXS performs operations inside Angular zone, even if you explicitly call `runOutsideAngular` - e.g. `select` still will be performed inside zone, because of `enterZone` operator.

`outsideZone` is an advanced option for developers who manually control the CD mechanism. E.g. all views in the app are marked as `OnPush` and explicit subscriptions inside views are avoided. All subscribtions are performed using `async` pipe. `async` is an impure pipe and also calls `ViewRef.markForCheck()` under the hood.


splincode:
@markwhitfeld this is a very important feature, I think it's worth accepting, if in the following PR we can fix the functionality if necessary

packages/store/src/internal/dispatcher.ts
```diff
@@ -96,14 +106,13 @@ export class InternalDispatcher {
...
       .pipe(
         exhaustMap((ctx: ActionContext) => {
-          switch (ctx.status) {
-            case ActionStatus.Successful:
-              return of(this._stateStream.getValue());
-            case ActionStatus.Errored:
-              return throwError(ctx.error);
-            default:
-              return empty();
+          if (ctx.status === ActionStatus.Successful) {
```
markwhitfeld: Why has this been changed from the switch to the if? Please change it back.

packages/store/src/operators/wrap.ts
```diff
@@ -0,0 +1,58 @@
...
+
+class WrapOperator<T> implements Operator<T, T> {
+  constructor(private outsideZone: boolean | null, private zone: NgZone) {}
+
+  public call(subscriber: Subscriber<T>, source: Observable<T>): Subscription {
+    return source.subscribe(new WrapSubscriber(subscriber, this.outsideZone, this.zone));
+  }
+}
+
+class WrapSubscriber<T> extends Subscriber<T> {
```
markwhitfeld: What is the reason for the vastly different approach to what was used with the enterZone pipe?
This is a lot more code.

packages/store/src/internal/dispatcher.ts
arturovt: @markwhitfeld Isn't that the same behavior, except `if-else` doesn't lead to nesting like `switch-case`?

packages/store/src/operators/wrap.ts
arturovt: @markwhitfeld I can re-write `wrap` operator in the same way as `enterZone` was implemented

packages/store/src/internal/dispatcher.ts
splincode: https://github.com/ngxs/store/issues/739
https://github.com/ngxs/store/issues/487

in that we can now control the management of the zone

This feature will allow you to disable extra detect changes in angular.

packages/store/src/operators/wrap.ts
splincode: https://github.com/ngxs/store/issues/739
https://github.com/ngxs/store/issues/487

in that we can now control the management of the zone

This feature will allow you to disable extra detect changes in angular.

splincode: revert by https://github.com/ngxs/store/pull/763

packages/store/src/internal/dispatcher.ts
splincode: revert by https://github.com/ngxs/store/pull/763

eranshmil:
The new file, `docs/advanced/options.md`, is not linked in the docs.
Is that on purpose?

splincode:
@eranshmil fixed
https://ngxs.gitbook.io/ngxs/v/master/advanced/options

eranshmil:
Thanks, but I don't think this should be under the `advanced` section.
Maybe inside the installation page.

Btw, I didn't tested it out throughly but when adding `outsideZone: true`, there's an error:
`Error: Right-hand side of 'instanceof' is not an object`

splincode:
@arturovt `Error: Right-hand side of 'instanceof' is not an object`

> Thanks, but I don't think this should be under the advanced section.
Actually quite suitable, as the module setting is advanced parameters.

eranshmil:
I got this error using the old stackblitz template. After upgrading all the deps, it's not happening anymore.

@markwhitfeld We should update the stackblitz template created by Austin.
https://stackblitz.com/edit/ngxs-simple

markwhitfeld:
@eranshmil yes, but I think it should be based off of a github folder so that it is easier to update.
Regarding this zone stuff, I'll be implementing a strategy pattern approach for solving this instead of the current solution

Issue #762, opened 2019-01-11
Can't build production when using documented dynamic selectors


kayvanbree:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[ ] Performance issue
[ ] Feature request
[x] Documentation issue or request
[ ] Support request => https://github.com/ngxs/store/blob/master/CONTRIBUTING.md
[ ] Other... Please describe:
</code></pre>

## Current behavior

When using a dynamic selector, the Angular compiler will complain about using Lambda function within a static function. It won't compile if you follow the documentation:
```
static pandas(type: string) {
    return createSelector([ZooState], (state: string[]) => {
      return state.filter(s => s.indexOf('panda') > -1)
                  .filter(s => s.indexOf(type) > -1);
    });
  }
```

## Expected behavior

I found out that adding `// @dynamic` above the class will let the Angular compiler ignore this warning, however I don't know if this is the correct way to deal with this problem. 
```
// @dynamic
@State<string[]>({
  name: 'animals',
  defaults: []
})
export class ZooState {
  static pandas(type: string) {
    return createSelector([ZooState], (state: string[]) => {
      return state.filter(s => s.indexOf('panda') > -1)
                  .filter(s => s.indexOf(type) > -1);
    });
  }
}
```
I would like to see the documentation updated to include something about this problem. This could either be telling people to add `// @dynamic` to the class,  or anything else if you guys have a better solution.

## What is the motivation / use case for changing the behavior?

I tried to build my library that I am refactoring to use NGXS. I couldn't find this problem on this repository nor on the website, but I think it should be addressed somewhere.

## Environment

<pre><code>
Libs:

```
    "@angular/core": "^6.0.3",
    "@ngxs/devtools-plugin": "^3.3.4",
    "@ngxs/store": "^3.3.4",
```

splincode:
Add repo for reproduce

kayvanbree:
We use a private SVN repo... :(

I'll create one to demo it.

kayvanbree:
Looks like it only happens when you use it in a library. 

You can find a reproduction at https://github.com/kayvanbree/ngxs-dynamic-selector-build-error, with the branch 'fixed' containing the `// @dynamic`, which counters the problem.

Here's the build error:

```
[...]

PR #770, opened 2019-01-14
docs(store): improve info about using lambdas in static methods

markwhitfeld:
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[X] Documentation content changes
[ ] Other... Please describe:
```

Issue Number: #762 
Other PR: #766 

## What is the new behavior?

Improved grammar issues, moved to new section at the bottom because it pertains to all selectors that contain a lambda in the `return` statement, not just dynamic selectors and corrected alternative fix after some investigation.

### DIFF
--- a/docs/concepts/select.md
+++ b/docs/concepts/select.md
@@ -190,43 +190,6 @@ export class ZooState {
 }
 ```
 
-**Pay attention!** If you specify `strictMetadataEmit` in the Angular compiler options - you have to include `@dynamic` comment before class expression. The Angular's `MetadataCollector` from the `@angular/compiler-cli` package reports about using lambdas in static methods:
-
-```TS
-// @dynamic
-@State<string[]>({
-  name: 'animals',
-  defaults: []
-})
-export class ZooState {
-
-  static pandas(type: string) {
-    return createSelector([ZooState], (state: string[]) => {
-      ....
-    });
-  }
-
-}
-```
-
-Either use a plain function expression except of lambda:
-
-```TS
-@State<string[]>({
-  name: 'animals',
-  defaults: []
-})
-export class ZooState {
-
-  static pandas(type: string) {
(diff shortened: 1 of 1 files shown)


docs/concepts/select.md
```diff
@@ -331,3 +294,85 @@ export class CityService {
 ```
 
 now we can use this `zooThemeParks` selector anywhere in our application.
+
+## Special Considerations
+
+### Angular Libraries: Use of lambdas in static functions
+
+_If you are building an Angular lib directly so that it can be deployed to npm. `strictMetadataEmit` will most likely be set to true in the Angular compiler options and, as a result, Angular's `MetadataCollector` from the `@angular/compiler-cli` package will report the following issue with using lambdas in static methods:_
```
eranshmil: What about linking to https://angular.io/guide/aot-compiler#strictmetadataemit?

docs/concepts/select.md
```diff
@@ -331,3 +294,85 @@ export class CityService {
...
+
+>  Metadata collected contains an error that will be reported at runtime: Lambda not supported.`
+
+This error would be reported for each of the selectors defined below but, as demonstrated in the sample, you can prevent this by including the `// @dynamic` comment before the class expression and decorators:
```
eranshmil: Btw, I think this comment also supported when placed above the selector itself.

markwhitfeld: Didn't seem to work in my tests. I read that it needs to be above an exported symbol. ie. above the declaration of something that has an associated `export`.

docs/concepts/select.md
markwhitfeld: Good point. Done 👍

markwhitfeld:
cc @splincode and @arturovt 
Note the corrected details on this issue

PR #782, opened 2019-01-18
chore(store): add typing for the Action decorator

deebloo:
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ x] Other... Please describe:
```
Add updated types for Action

Issue Number: N/A

## What is the new behavior?
The Action decorator doesn't accept "any" anymore it accepts and ActionDef

```
[ ] Yes
[ x] No
```

### DIFF
--- a/packages/store/src/symbols.ts
+++ b/packages/store/src/symbols.ts
@@ -44,6 +44,12 @@ export class NgxsConfig {
   }
 }
 
+export interface ActionDef {
+  type: string;
+
+  new (...args: any[]): any;
+}
+
 export type StateOperator<T> = (existing: Readonly<T>) => T;
 
 /**
--- a/packages/store/src/decorators/action.ts
+++ b/packages/store/src/decorators/action.ts
@@ -1,11 +1,11 @@
 import { ensureStoreMetadata } from '../internal/internals';
-import { ActionOptions } from '../symbols';
+import { ActionOptions, ActionDef } from '../symbols';
 
 /**
  * Decorates a method with a action information.
  */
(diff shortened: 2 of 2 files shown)


packages/store/src/symbols.ts
```diff
@@ -44,6 +44,12 @@ export class NgxsConfig {
   }
 }
 
+export interface ActionDef {
```
arturovt: you could also do `export type ActionDef<T = any> = Type<T> & { type: string };`

deebloo: true we could

arturovt: @deebloo `store/src/store.ts` uses `Type` from Angular lib, but it's up to you :wink:

eranshmil: I think we should decrease the angular dependency, in order to someday make the library available also for other libraries/frameworks.

splincode: @eranshmil hardly

eranshmil: All the angular stuff should be removed from the core and moved to a plugin.

arturovt:
@eranshmil I understand what you're talking about, but the DI in this library works only thanks to Angular

eranshmil:
> @eranshmil I understand what you're talking about, but the DI in this library works only thanks to Angular

Are you talking about the DI in the applications that use NGXS?
The DI is not relevant in React application, for instance.

arturovt:
@eranshmil I'm talking about the DI in the NGXS library

Issue #741, opened 2019-01-03
Add warnings for duplicate states


splincode:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[x] Bug report  
[ ] Performance issue
[ ] Feature request
[ ] Documentation issue or request
[ ] Support request => https://github.com/ngxs/store/blob/master/CONTRIBUTING.md
[ ] Other... Please describe:
</code></pre>

## Current behavior

Duplicates state are deleted while there is no check
https://stackblitz.com/edit/ngxs-simple-e6j3vw?file=app%2Fapp.state.ts

## Expected behavior

- If state already exists in state tree, need show warning
- For HMR it is not necessary to reload existing states

splincode:
Related: https://github.com/ngxs/store/issues/719

Issue #719, opened 2018-12-19
HMR not working with lazy-loaded states


eranshmil:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[x] Bug report  
[ ] Performance issue
[ ] Feature request
[ ] Documentation issue or request
[ ] Support request => https://github.com/ngxs/store/blob/master/CONTRIBUTING.md
[ ] Other... Please describe:
</code></pre>

## Current behavior

The state is being reset after @@UPDATE_STATE action.

## Expected behavior

The state will remain the same after the reload.

## Minimal reproduction of the problem with instructions
1. Navigate to a lazy-loaded module and update the relevant state.
1. Trigger HMR.
1. The state will reset to the default values after @@UPDATE_STATE.

All the relevant information and repro could be found here: https://github.com/eranshmil/ngxs-hmr-playground

![](https://github.com/eranshmil/ngxs-hmr-playground/raw/master/assets/hmr.gif)

## Environment

<pre><code>
Libs:
- @angular/core version: 7.1.3
- @ngxs/store version: 3.3.3-dev.master-bd16c95

Browser:
- [x] Chrome (desktop) version 70.0.3538.110
- [ ] Chrome (Android) version XX
- [ ] Chrome (iOS) version XX
- [ ] Firefox version XX
- [ ] Safari (desktop) version XX
- [ ] Safari (iOS) version XX
- [ ] IE version XX
- [ ] Edge version XX
 
For Tooling issues:
- Node version: 11.3.0
- Platform:  Mac
</code></pre>

splincode:
I will deal with this issue soon

splincode:
@eranshmil can check again

eranshmil:
Working now. :)

Issue #179, opened 2018-03-27
Sub states cannot have same name in diferrent lazy loaded modules.


dreamer99x:
### Versions

```
* ngxs: 2.0.0-rc.23
* @angular/core: 5.2.8
```

### Repro steps

See the sample:
https://stackblitz.com/edit/ngxs-lazy-load

* There are two lazy loaded modules: UserModule, TodoModule. 
* UserModule has UserState named 'user' with sub state named 'list'.
* TodoModule has TodoState named 'todo' with sub state named 'list'.

### Observed behavior

When application starts, TodoModule is loaded.
When you click on User link, UserModule is loaded and application fails on:

```
Error: Uncaught (in promise): Error: Store has already been added: list
Error: Store has already been added: list
    at _loop_1 (state-factory.ts:62)
    at StateFactory.add (state-factory.ts:45)
    at StateFactory.addAndReturnDefaults (state-factory.ts:91)
    at NgxsFeatureModule.initStates (module.ts:60)
    at new NgxsFeatureModule (module.ts:54)
    at _createClass (ng_module.js:152)
    at _createProviderInstance$1 (ng_module.js:119)
    at initNgModule (ng_module.js:70)
    at new NgModuleRef_ (refs.js:1033)
    at createNgModuleRef (refs.js:1026)
    at _loop_1 (state-factory.ts:62)
    at StateFactory.add (state-factory.ts:45)
    at StateFactory.addAndReturnDefaults (state-factory.ts:91)
    at NgxsFeatureModule.initStates (module.ts:60)
    at new NgxsFeatureModule (module.ts:54)
    at _createClass (ng_module.js:152)
    at _createProviderInstance$1 (ng_module.js:119)
    at initNgModule (ng_module.js:70)
    at new NgModuleRef_ (refs.js:1033)
    at createNgModuleRef (refs.js:1026)
    at resolvePromise (zone.js:809)
    at resolvePromise (zone.js:775)
    at eval (zone.js:858)
    at ZoneDelegate.invokeTask (zone.js:421)
    at Object.onInvokeTask (ng_zone.js:585)
    at ZoneDelegate.invokeTask (zone.js:420)
    at Zone.runTask (zone.js:188)
    at drainMicroTaskQueue (zone.js:595)
    at ZoneTask.invokeTask [as invoke] (zone.js:500)
    at invokeTask (zone.js:1517)
```

amcdnl:
Can't do this. This is actually in the docs too.

[...]

PR #791, opened 2019-01-25
feat(store): reusable states

splincode:
## PR Checklist

## PR Type

```
[x] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: 
https://github.com/ngxs/store/issues/179
https://github.com/ngxs/store/issues/741
https://github.com/ngxs/store/issues/719

## What is the new behavior?

throw an error, since the previous state will be overwritten

```
[ ] Yes
[x] No
```

### DIFF
--- a/packages/store/src/internal/state-factory.ts
+++ b/packages/store/src/internal/state-factory.ts
@@ -16,18 +16,22 @@ import {
   findFullParentPath,
   isObject,
   MappedStore,
+  MetaDataModel,
   nameToState,
+  ObjectKeyMap,
   propGetter,
   StateClass,
+  StateKeyGraph,
   StatesAndDefaults,
   topologicalSort
 } from './internals';
-import { getActionTypeFromInstance, setValue } from '../utils/utils';
+import { getActionTypeFromInstance, getValue, setValue } from '../utils/utils';
 import { ofActionDispatched } from '../operators/of-action';
 import { ActionContext, ActionStatus, InternalActions } from '../actions-stream';
 import { InternalDispatchedActionResults } from '../internal/dispatcher';
 import { StateContextFactory } from '../internal/state-context-factory';
 import { StoreValidators } from '../utils/store-validators';
+import { InternalStateOperations } from '../internal/state-operations';
 
 /**
  * State factory class
@@ -46,7 +50,8 @@ export class StateFactory {
     private _parentFactory: StateFactory,
     private _actions: InternalActions,
     private _actionResults: InternalDispatchedActionResults,
-    private _stateContextFactory: StateContextFactory
+    private _stateContextFactory: StateContextFactory,
+    private _internalStateOperations: InternalStateOperations
   ) {}
 
   private _states: MappedStore[] = [];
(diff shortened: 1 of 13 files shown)


packages/store/src/utils/store-validators.ts
```diff
@@ -14,4 +17,33 @@ export abstract class StoreValidators {
...
+    const statesNames: Map<StateName, StateClassName> = new Map();
+
+    // https://jsperf.com/for-vs-foreach/75
+    for (let i = 0, size: number = states.length; i < size; i++) {
+      const state: StateClass = states[i]!;
+      const meta: MetaDataModel = this.validateStateMeta(state);
+      const stateName: string = <string>meta!.name;
+
+      if (statesNames.has(stateName)) {
+        const previousStateName: string = <string>statesNames.get(stateName);
```
arturovt:
```TS
const previousStateName = stateNames.get(stateName) as string;
```

packages/store/src/utils/store-validators.ts
```diff
@@ -14,4 +17,33 @@ export abstract class StoreValidators {
...
   }
+
+  public static validateStateNames(states: StateClass[]): Map<StateName, StateClassName> {
+    const statesNames: Map<StateName, StateClassName> = new Map();
+
+    // https://jsperf.com/for-vs-foreach/75
+    for (let i = 0, size: number = states.length; i < size; i++) {
+      const state: StateClass = states[i]!;
+      const meta: MetaDataModel = this.validateStateMeta(state);
+      const stateName: string = <string>meta!.name;
```
arturovt:
```TS
const stateName = meta!.name as string;
```

packages/store/src/utils/store-validators.ts
splincode: done

splincode:
@markwhithfeld ping

tsconfig.json
```diff
@@ -15,6 +15,7 @@
     "moduleResolution": "node",
     "emitDecoratorMetadata": true,
     "experimentalDecorators": true,
+    "downlevelIteration": true,
```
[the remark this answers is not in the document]
splincode: Issue: https://github.com/Microsoft/TypeScript/issues/27008

I use merge two maps
https://github.com/ngxs/store/pull/791/files#diff-62bb81999aabeb9aaa84c63a01b0670eR64

splincode:
@markwhitfeld @arturovt I have improved the formation of the state tree so that the hmr-plugin will now work perfectly

tsconfig.json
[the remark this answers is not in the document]
arturovt: @markwhitfeld in simple terms `downlevelIteration` option is useful when you target `es5` and you have to do some iteration over ES2015 collection. TypeScript compiler emits optimized function that contains contains a proper implementation of the iteration protocol.

packages/store/src/internal/state-factory.ts
```diff
@@ -49,83 +50,98 @@ export class StateFactory {
...
+  }
+
+  private static immutableDefaults(defaults: any): any {
+    let value = {};
+
+    if (Array.isArray(defaults)) {
+      value = [...defaults];
+    } else if (isObject(defaults)) {
+      value = { ...defaults };
+    } else if (defaults === undefined) {
```
arturovt: do you need this condition?

packages/store/src/internal/state-factory.ts
```diff
@@ -195,4 +211,21 @@ export class StateFactory {
...
+  }
+
+  /**
+   * @description
+   * the method checks if the state has already been added to the tree
+   * and completed the life cycle
+   * @param name
+   * @param depth
+   */
+  private hasWillMounted(name: string, depth: string): boolean {
```
arturovt: `hasBeenMounted`

packages/store/src/utils/store-validators.ts
```diff
@@ -14,4 +17,32 @@ export abstract class StoreValidators {
       throw new Error(this.stateNameErrorMessage(name));
     }
   }
+
+  public static getValidStateNamesMap(states: StateClass[]): Map<StateName, StateClassName> {
+    const uniqueStateNamesMap: Map<StateName, StateClassName> = new Map();
+
+    for (let i = 0, size: number = states.length; i < size; i++) {
```
arturovt: maybe call it `length` - more semantic maybe? and also do you need to specify the `number` type here?

packages/store/src/internal/state-factory.ts
splincode: Why not?

arturovt: i mean you declare a `value` variable before all these conditions and assign to `{}` and also in this condition you assign to `{}`, maybe better:
```TS
let value = {};

if (Array.isArray(defaults)) {
  value = [...defaults];
} else if (isObject(defaults)) {
  value = { ...defaults };
} else if (defaults !== undefined) {
  value = defaults;
}
```

markwhitfeld: Just my 2c, I would say that having multiple returns would actually be better here. 
Guard clauses are one of the definite exceptions to the guidelines from Code Complete.
Could be expressed as:
```TS
if (Array.isArray(defaults)) return [...defaults];
if (isObject(defaults)) return { ...defaults };
if (typeof defaults === 'undefined') return {};
return defaults;
```

packages/store/src/utils/store-validators.ts
```diff
@@ -14,4 +17,32 @@ export abstract class StoreValidators {
       throw new Error(this.stateNameErrorMessage(name));
     }
   }
+
+  public static getValidStateNamesMap(states: StateClass[]): Map<StateName, StateClassName> {
```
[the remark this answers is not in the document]
markwhitfeld: Would not detect duplicates from feature states.

tsconfig.json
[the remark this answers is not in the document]
markwhitfeld: I see that the reason this would have been added was to iterate over the `Map`.

Issue #794, opened 2019-01-27
Regression with action type


splincode:
## I'm submitting a...

<pre><code>
[x] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[ ] Performance issue
[ ] Feature request
[ ] Documentation issue or request
[ ] Support request => https://github.com/ngxs/store/blob/master/CONTRIBUTING.md
[ ] Other... Please describe:
</code></pre>

## Current behavior

![image](https://user-images.githubusercontent.com/12021443/51796801-d7ad0000-2209-11e9-8fe1-2b493d970069.png)

## Expected behavior

@deebloo in v3.3.0
earlier I could safely use objects

deebloo:
Fix is in master now

PR #800, opened 2019-01-27
fix(store): ease ActionDef signature

deebloo:
## PR Checklist

## PR Type
Fix for #794

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: #794

```
[ ] Yes
[x] No
```

### DIFF
--- a/packages/store/src/decorators/action.ts
+++ b/packages/store/src/decorators/action.ts
@@ -4,7 +4,10 @@ import { ActionOptions, ActionDef } from '../symbols';
 /**
  * Decorates a method with a action information.
  */
-export function Action(actions: ActionDef | ActionDef[], options?: ActionOptions) {
+export function Action(
+  actions: ActionDef | ActionDef[] | { type: string } | { type: string }[],
+  options?: ActionOptions
+) {
   return function(target: any, name: string, _descriptor: TypedPropertyDescriptor<any>) {
     const meta = ensureStoreMetadata(target.constructor);
 
@@ -14,14 +17,6 @@ export function Action(actions: ActionDef | ActionDef[], options?: ActionOptions
 
     for (const action of actions) {
       const type = action.type;
-
-      if (!action.type) {
-        throw new Error(
-          `Action ${
-            'name' in action ? action['name'] : JSON.stringify(action)
-          } is missing a static "type" property`
-        );
-      }
 
       if (!meta.actions[type]) {
         meta.actions[type] = [];
--- a/packages/store/src/decorators/action.ts
+++ b/packages/store/src/decorators/action.ts
@@ -16,7 +16,11 @@ export function Action(actions: ActionDef | ActionDef[], options?: ActionOptions
       const type = action.type;
 
       if (!action.type) {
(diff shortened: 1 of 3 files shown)


deebloo:
@markwhitfeld @splincode are we sure that we want to support object literals for the Action decorator?

splincode:
@deebloo this is very convenient if I don’t want to create a separate action class for this!

```ts
@State(..)
export class CounterState {
  public static increase: string = 'increase';

  @Action({ type: CounterState.increase })
  public increase(ctx) {
    ctx.setState((state) => state++);
  }

}
```

app.component.ts 

```ts
class AppComponent {
  ...  
  ngOnInit() {
    this.store.dispatch({ type: CounterState.increase });
  }
 
}
```

markwhitfeld:
@splincode Does that code sample currently work in plain ngxs?

splincode:
@markwhitfeld Yes, we use it in our projects, because we are uncomfortable creating action classes

markwhitfeld:
Yeah, in my mind it is not a recommended approach but if it is a currently supported then we should continue to support it.
It is not ideal, but as you can see with @splincode it is something that some people prefer.
One of my hopes with v4.0.0 is that we can make the `type` property optional again. I have some ideas how we can acheive this in a production optimised minified code compatible way.

I think it is ok for us to continue to maintain support for the object literal form as an alternative approach in future major versions too.

markwhitfeld:
@deebloo The issue with the fisrt attempt to remove the `type` property is that it relied on the name of the class and this would get scrambled and potentially have duplicates when the code was minified for production.
The basic idea to solve this is that if you pass an action that does not contain a `type` property and it is a class then you can generate a type name and add this type name to the class' prototype object.

The generated name could be based on the class name but then adds some unique identifier (an incrementing number). This would make it useful in dev but also unique in minified prod.

Does this make sense?

deebloo:
Oh yeah. For a brief time we had that. There was a counter used to tag action classes but it was removed. I don't remember the reason though.

splincode:
@markwhitfeld need to fix this before the release

eranshmil:
> Would it be possible for you to find the commit?

I think that's all the relevant PRs:
https://github.com/ngxs/store/pull/212
https://github.com/ngxs/store/pull/211
https://github.com/ngxs/store/pull/210
https://github.com/ngxs/store/pull/209
https://github.com/ngxs/store/pull/208
https://github.com/ngxs/store/pull/207
https://github.com/ngxs/store/pull/206

deebloo:
@splincode ill get to add the tests today

packages/store/src/decorators/action.ts
```diff
@@ -15,10 +18,6 @@ export function Action(actions: ActionDef | ActionDef[], options?: ActionOptions
     for (const action of actions) {
       const type = action.type;
 
-      if (!action.type) {
-        throw new Error(`Action ${action.name} is missing a static "type" property`);
-      }
-
```
deebloo: no need for the runtime check if the compiler will catch the error first

PR #811, opened 2019-02-04
feat: add execution strategy to decouple from zones

markwhitfeld:
## PR Checklist

## PR Type

```
[ ] Bugfix
[X] Feature
[X] Refactoring (no functional changes, no api changes)
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Numbers: #739  #487 
Removing approach introduced in PR #761 in favor of using a strategy approach

## What is the new behavior?

Added a new module option called `executionStrategy`.
This is an example of using the `NoopNgxsExecutionStrategy` so that NGXS doesn't mess with zones.
```
NgxsModule.forRoot([CounterState], {
          executionStrategy: NoopNgxsExecutionStrategy
        })
```

```
[ ] Yes
[X] No
```
Well, it does break something that only exists in the dev version now.
The `outsideZones` option has been removed.


packages/store/src/symbols.ts
```diff
@@ -147,3 +153,13 @@ export interface NgxsAfterBootstrap {
...
 
 export type NgxsLifeCycle = Partial<NgxsOnInit> & Partial<NgxsAfterBootstrap>;
+
+export type FunctionCall = <T>(func: (...args: any[]) => T) => T;
+
+/*
+ * Execution strategy interface
+ */
+export interface NgxsExecutionStrategy {
+  enter<T>(func: (...args: any[]) => T): T;
```
splincode: why do you do not use everywhere your FunctionCall type ?

packages/store/src/symbols.ts
```diff
@@ -147,3 +153,13 @@ export interface NgxsAfterBootstrap {
 }
 
 export type NgxsLifeCycle = Partial<NgxsOnInit> & Partial<NgxsAfterBootstrap>;
+
+export type FunctionCall = <T>(func: (...args: any[]) => T) => T;
```
splincode: Rename to 

```ts
type Callback<T = any, U = any> = (...args: U[]) => T;
```

packages/store/src/execution/dispatchOutsideZoneNgxsExecutionStrategy.ts
```diff
@@ -0,0 +1,52 @@
...
+import { NgxsExecutionStrategy } from '../symbols';
+import { ConfigValidator } from '../internal/config-validator';
+
+@Injectable()
+export class DispatchOutsideZoneNgxsExecutionStrategy implements NgxsExecutionStrategy {
+  constructor(private _ngZone: NgZone, @Inject(PLATFORM_ID) private _platformId: Object) {
+    this.verifyZoneIsNotNooped(this._ngZone);
+  }
+
+  enter<T>(func: (...args: any[]) => T): T {
```
splincode: enter<T>(callback: Callback<T>)

and others

packages/store/src/execution/dispatchOutsideZoneNgxsExecutionStrategy.ts
```diff
@@ -0,0 +1,52 @@
...
+    if (isPlatformServer(this._platformId)) {
+      return this.runInsideAngular(func);
+    } else {
```
eranshmil: You can get rid of the else and just leave the return statement.

packages/store/src/execution/dispatchOutsideZoneNgxsExecutionStrategy.ts
```diff
@@ -0,0 +1,52 @@
...
+  ɵNoopNgZone as NoopNgZone,
+  Inject
+} from '@angular/core';
+import { isPlatformServer } from '@angular/common';
+
+import { NgxsExecutionStrategy } from '../symbols';
+import { ConfigValidator } from '../internal/config-validator';
+
+@Injectable()
+export class DispatchOutsideZoneNgxsExecutionStrategy implements NgxsExecutionStrategy {
```
eranshmil: Maybe remove the `Ngxs` from the class name?

markwhitfeld:
@ngxs/core Are you happy with the name `NoopNgxsExecutionStrategy` for the strategy that does nothing except call on to the provided function?

Also the interface is called `NgxsExecutionStrategy`. I manged this in the same way that the plugin interface is named with a `Ngxs` prefix. Therefore subsequent classes that implement this interface are ended with `NgxsExecutionStrategy`.
Are you happy with this naming?

Ps. I will add the docs commit once naming is decided.

splincode:
@markwhitfeld seems to be OK

packages/store/src/execution/dispatchOutsideZoneNgxsExecutionStrategy.ts
markwhitfeld: I followed the convention of the `NgxsPlugin` interface to start with `Ngxs`.
So, because the interface is called `NgxsExecutionStrategy` the class would end with that by convention. ... ?

packages/store/src/symbols.ts
markwhitfeld: I think it gave me problems. Will look again.

packages/store/src/symbols.ts
markwhitfeld: Not necessarily a callback. It is to represent a function parameter to a higher order function.

packages/store/src/execution/dispatchOutsideZoneNgxsExecutionStrategy.ts
markwhitfeld: A callback would be something that would happen on completion of some work IMO.
Here the function is describing the work to do, not something to call aftet the work is complete.
There is a distinction in my mind... but maybe it is just me?

packages/store/src/execution/dispatchOutsideZoneNgxsExecutionStrategy.ts
markwhitfeld: Will do.

arturovt:
@markwhitfeld Looks great! Just add `NgxsExecutionStrategy` to the `public_api`.

packages/store/src/execution/dispatchOutsideZoneNgxsExecutionStrategy.ts
markwhitfeld: Done

packages/store/src/execution/dispatchOutsideZoneNgxsExecutionStrategy.ts
markwhitfeld: See my comment below

packages/store/src/symbols.ts
markwhitfeld: Got rid of this because it is not used. Made signature simpler so that this is not necessary as well.

packages/store/src/symbols.ts
markwhitfeld: Got rid of this because it is not used. Made signature simpler so that this is not necessary as well.

markwhitfeld:
@splincode Review again please 👍

packages/store/src/execution/dispatchOutsideZoneNgxsExecutionStrategy.ts
```diff
@@ -0,0 +1,51 @@
...
+  NgZone,
+  PLATFORM_ID,
+  Injectable,
+  ɵNoopNgZone as NoopNgZone,
+  Inject
+} from '@angular/core';
+import { isPlatformServer } from '@angular/common';
+
+import { NgxsExecutionStrategy } from './symbols';
+import { FunctionExpression } from 'typescript';
```
splincode: FunctionExpression - where usage?

packages/store/src/internal/config-validator.ts
```diff
@@ -1,11 +1,11 @@
-import { Injectable, NgZone, isDevMode, ɵNoopNgZone as NoopNgZone } from '@angular/core';
+import { Injectable, isDevMode } from '@angular/core';
 
 import { isAngularInTestMode } from '../utils/angular';
 import { NgxsConfig } from '../symbols';
 
 @Injectable()
 export class ConfigValidator {
-  constructor(private config: NgxsConfig, private zone: NgZone) {}
+  constructor(private config: NgxsConfig) {}
```
splincode:
```ts
private _config: NgxsConfig
```

packages/store/src/execution/dispatchOutsideZoneNgxsExecutionStrategy.ts
splincode: Rename `DispatchOutsideZoneNgxsExecutionStrategy` to `DispatchOutsideZone`

packages/store/src/execution/internalNgxsExecutionStrategy.ts
```diff
@@ -0,0 +1,18 @@
...
+
+import { NgxsExecutionStrategy, NGXS_EXECUTION_STRATEGY } from './symbols';
+
+@Injectable()
+export class InternalNgxsExecutionStrategy implements NgxsExecutionStrategy {
+  constructor(
+    @Inject(NGXS_EXECUTION_STRATEGY) private _executionStrategy: NgxsExecutionStrategy
+  ) {}
+
+  enter<T>(func: () => T): T {
```
splincode: rename  `func`  to `tick`

packages/store/src/operators/leaveNgxs.ts
```diff
@@ -0,0 +1,26 @@
+import { MonoTypeOperatorFunction, Observable, Observer } from 'rxjs';
+import { NgxsExecutionStrategy } from '../execution/symbols';
+
+/**
+ * Returns operator that will run
+ * `subscribe` outside of the ngxs execution context
+ */
+export function leaveNgxs<T>(
```
splincode: could it be better to call it another way?
`ngxsStreamListener`

packages/store/src/symbols.ts
```diff
@@ -37,7 +39,7 @@ export class NgxsConfig {
    * server side rendering, as dispatching on the server is still wrapped into `zone.run`.
```
splincode: need fix comments

packages/store/src/execution/dispatchOutsideZoneNgxsExecutionStrategy.ts
markwhitfeld: Argh! Must have been added when I was exploring the typescript types.
I'll remove.

packages/store/src/internal/config-validator.ts
markwhitfeld: Not really related to my changes, but I'll do the rename anyway

packages/store/src/execution/internalNgxsExecutionStrategy.ts
markwhitfeld: Huh? I don't understand why `tick` is a better name?

packages/store/src/operators/leaveNgxs.ts
markwhitfeld: Could you explain why?

packages/store/src/symbols.ts
markwhitfeld: Will do

packages/store/src/operators/leaveNgxs.ts
splincode: Because basically we do a wrapper over the behavior subject stream

packages/store/src/execution/dispatchOutsideZoneNgxsExecutionStrategy.ts
markwhitfeld: Ok, then how would we name the `NoopNgxsExecutionStrategy`?

packages/store/src/execution/internalNgxsExecutionStrategy.ts
splincode: image

better anologue

splincode: Because in any case, we do it for the angular zones.

packages/store/src/execution/dispatchOutsideZoneNgxsExecutionStrategy.ts
splincode: `NoopNgZone`, `OutsideZoneNoop` ?

markwhitfeld: The `NoopNgxsExecutionStrategy` has got nothing to do with Zones. The only thing that has something to do with zones is the `DispatchOutsideZoneNgxsExecutionStrategy`.

packages/store/src/internal/config-validator.ts
markwhitfeld: Done

packages/store/src/execution/internalNgxsExecutionStrategy.ts
markwhitfeld: But this function is doing the work, not the change detection. Looking at the description `tick` definitely doesn't suit this.

packages/store/src/operators/leaveNgxs.ts
markwhitfeld: I think it should be named after what it acheives as opposed to how it does it.

packages/store/src/symbols.ts
markwhitfeld: Done

markwhitfeld:
@splincode I have updated the docs. I don't think that examples of a custom implementation is necessary because this is an advanced and very edge case customisation.

packages/store/src/execution/dispatchOutsideZoneNgxsExecutionStrategy.ts
splincode: ok

packages/store/src/execution/internalNgxsExecutionStrategy.ts
splincode: ok

packages/store/src/operators/leaveNgxs.ts
splincode: ok

markwhitfeld:
Yippee!!! Now to get the 3.4 release and announcement post ready!

PR #815, opened 2019-02-05
refactor(store): decomposition ngxs module

splincode:
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

```
[ ] Yes
[x] No
```


packages/store/src/modules/ngxs-feature.module.ts
```diff
@@ -0,0 +1,42 @@
...
+    // add stores to the state graph and return their defaults
+    const results: StatesAndDefaults = factory.addAndReturnDefaults(flattenedStates);
+
+    internalStateOperations.setStateToTheCurrentWithNew(results);
+
+    // dispatch the update action and invoke init and bootstrap functions after
+    lifecycleStateManager.ngxsBootstrap(new UpdateState(), results);
+  }
+
+  private static statesFlat(states: StateClass[][] = []): StateClass[] {
```
arturovt: `flattenStates`

splincode: done

splincode:
@markwhitfeld ping

markwhitfeld:
@splincode I'm delaying these refactors until after the release just to be safe.

splincode:
@markwhitfeld what's problem?

splincode:
@markwhitfeld can we keep this thread? so that it does not hang

splincode:
@markwhitfeld ping

packages/store/src/module.ts
```diff
@@ -1,108 +1,39 @@
...
-      internalStateOperations.setStateToTheCurrentWithNew(results);
-      // dispatch the update action and invoke init and bootstrap functions after
-      lifecycleStateManager.ngxsBootstrap(new UpdateState(), results);
-    }
-  }
-}
-
-export type ModuleOptions = Partial<NgxsConfig>;
-
-export function ngxsConfigFactory(options: ModuleOptions): NgxsConfig {
```
markwhitfeld: Both these factories (`ngxsConfigFactory` and `appBootstrapListenerFactory`) are better kept in the same file as the `NgxsModule` because they aren't used from anywhere else. Also, moving these away could cause problems with AOT compilation (...or maybe that has been fixed in the Angular compiler now).

I think that the `ModuleOptions` and the `ROOT_OPTIONS` should also remain in this file because they are concepts that are local to the bootstrapping methods in this file (they are only exported to satisfy the Angular compiler).

splincode: Why do we have to move the ModuleOptions file? It should be public (expose from symbol.ts), as we describe it in the documentation for use it.

packages/store/src/module.ts
```diff
@@ -137,17 +72,17 @@ export class NgxsModule {
...
           provide: NgxsConfig,
-          useFactory: ngxsConfigFactory,
-          deps: [ROOT_OPTIONS]
+          useFactory: NgxsModule.ngxsConfigFactory,
+          deps: [NgxsModule.ROOT_OPTIONS]
         },
         {
           provide: APP_BOOTSTRAP_LISTENER,
-          useFactory: appBootstrapListenerFactory,
+          useFactory: NgxsModule.appBootstrapListenerFactory,
```
markwhitfeld: Is this safe for AOT compilation?

splincode: yes

splincode:
@markwhitfeld ping

splincode:
@markwhitfeld ping

packages/store/src/symbols.ts
```diff
@@ -7,11 +7,14 @@ import { DispatchOutsideZoneNgxsExecutionStrategy } from './execution/dispatch-o
...
 export const ROOT_STATE_TOKEN = new InjectionToken<any>('ROOT_STATE_TOKEN');
 export const FEATURE_STATE_TOKEN = new InjectionToken<any>('FEATURE_STATE_TOKEN');
+export const NGXS_PLUGINS = new InjectionToken('NGXS_PLUGINS');
+
 export const META_KEY = 'NGXS_META';
 export const META_OPTIONS_KEY = 'NGXS_OPTIONS_META';
 export const SELECTOR_META_KEY = 'NGXS_SELECTOR_META';
 
-export const NGXS_PLUGINS = new InjectionToken('NGXS_PLUGINS');
+export type NgxsModuleOptions = Partial<NgxsConfig>;
```
markwhitfeld: The `NgxsModuleOptions` does not need to move here. It is best kept close to it's only usage in `NgxsModule`.

markwhitfeld: It does not need to be exposed publically.

splincode: Why?

splincode: done

splincode:
@markwhitfeld thank you very much )))

PR #820, opened 2019-02-07
WIP: feat(websocket-plugin): dispatch WebSocketDisconnected Action upon socket disconnect; dispatch WebsocketMessageError on received message error

kwalski:
## PR Checklist
Please check if your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://github.com/ngxs/store/blob/master/CONTRIBUTING.md#commit
- [x] Tests for the changes have been added (for bug fixes / features)
- [x] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

```
[x] Bugfix
[x] Feature
[ ] Code style update (formatting, local variables)
[ ] Refactoring (no functional changes, no api changes)
[ ] Build related changes
[ ] CI related changes
[x] Documentation content changes
[ ] Other... Please describe:
```

## What is the current behavior?
* Dispatches WebSocketDisconnected action upon disconnect to allow user write a reducer/effect to track websocket connection state and retry if required. Note that this is plugin dispatched and hence different to DisconnectWebSocket which can be programatically dispatched
* WebsocketMessageError was not dispatched because this._socket in websocket-subject.ts is not handling error and complete functions in subscribe 

Issue Number: N/A

## Does this PR introduce a breaking change?
```
[ ] Yes
[x] No
```

PR #825, opened 2019-02-08
 feat(websocket-plugin): add WebSocketDisconnected, WebsocketMessageError actions

kwalski:
## PR Checklist

## PR Type

```
[x] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

* Dispatches WebSocketDisconnected action upon disconnect to allow user write a reducer/effect to track websocket connection state and retry if required. Note that this is plugin dispatched and hence different to DisconnectWebSocket which can be programatically dispatched
* WebsocketMessageError was not dispatched because this._socket in websocket-subject.ts is not handling error and complete functions in subscribe 

Issue Number: N/A

```
[ ] Yes
[x] No
```

## Other information
https://github.com/ngxs/store/pull/820


packages/websocket-plugin/src/symbols.ts
```diff
@@ -56,7 +56,7 @@ export class ConnectWebSocket {
 /**
  * Action triggered when a error ocurrs
  */
-export class WebsocketMessageError {
+export class WebSocketMessageError {
```
splincode: This is breaking changes!

kwalski: This action doesn't work before this PR, so can't be breaking anything

splincode: Seriously? Add test )))

kwalski: Please look at change on websocket-subject.ts line# 64, error was not handled before this PR

arturovt:
@eranshmil my opinion is we have to release the next version as soon as we can, because there a lot of features awaited by people. What do you think?

eranshmil:
> @eranshmil my opinion is we have to release the next version as soon as we can, because there a lot of features awaited by people. What do you think?

We could merge this PR without the `WebSocketMessageError`, and release the minor version.
For the major version, all this inconsistency should be resolved.

arturovt:
@eranshmil Yeah, probably, also could you please look with Mark at the edited markdown file, maybe there is some inconsistency in descriptions?

packages/websocket-plugin/src/symbols.ts
markwhitfeld: Although tempting to fix this because it didn't work before we need to be careful not to make breaking changes.
Please revert this rename and we will add it to the list of breaking changes we want to introduce in v4.

markwhitfeld:
I agree, this PR is good to go other than the rename.
So the `WebSocketMessageError` should be named back to `WebsocketMessageError`.
It is tempting to make a concession but we need to follow semantic versioning.
In the next major release we will rename all occurences of `Websocket` to `WebSocket` to make it consistent

kwalski:
> So the `WebSocketMessageError` should be named back to `WebsocketMessageError`.

@markwhitfeld  this is reverted

packages/websocket-plugin/src/symbols.ts
kwalski: reverted

packages/websocket-plugin/src/websocket-handler.ts
```diff
@@ -7,7 +7,8 @@ import {
   SendWebSocketMessage,
   NGXS_WEBSOCKET_OPTIONS,
   NgxsWebsocketPluginOptions,
-  WebsocketMessageError
+  WebsocketMessageError,
```
splincode: add WebsocketMessageError, WebSocketDisconnected in public_api

splincode:
image

packages/websocket-plugin/src/websocket-handler.ts
kwalski: @splincode where? In public_api.ts?? Why?

splincode: https://github.com/ngxs/store/blob/master/packages/websocket-plugin/src/public_api.ts

splincode: Add all

```
ConnectWebSocket,
  SendWebSocketMessage,
  DisconnectWebSocket,
  WebSocketDisconnected,
  WebsocketMessageError
```

splincode: Because it was necessary to do it initially, they forgot to add it last time.

```ts
constructor(private router: Router, private actions$: Actions) {
    this.actions$
      .pipe(ofActionDispatched(DisconnectWebSocket))
      .subscribe(({ payload }) =>console.log('disconnect');
  }
```

kwalski: There is already this line in public_api, not sure what you want to achieve and how is it relevant to this PR, I suggest create issue to track existing backlog
export * from './symbols';

PR #829, opened 2019-02-09
refactor(store): decomposition ngxs config

splincode:
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

now all services in one file

## What is the new behavior?

file decomposition and structuring

```
[ ] Yes
[x] No
```

### DIFF
--- a/packages/store/src/internal/config/ngxs-config.ts
+++ b/packages/store/src/internal/config/ngxs-config.ts
@@ -1,0 +1,48 @@
+import { Injectable, Type } from '@angular/core';
+
+import { ObjectKeyMap } from '../internals';
+import { NgxsCompatibility } from '../../symbols';
+import { NgxsExecutionStrategy } from '../../execution/symbols';
+import { DispatchOutsideZoneNgxsExecutionStrategy } from '../../execution/dispatchOutsideZoneNgxsExecutionStrategy';
+
+/**
+ * The NGXS config settings.
+ */
+@Injectable()
+export class NgxsConfig {
+  /**
+   * Run in development mode. This will add additional debugging features:
+   * - Object.freeze on the state and actions to guarantee immutability
+   * (default: false)
+   */
+  developmentMode: boolean;
+
+  /**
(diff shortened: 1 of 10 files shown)


packages/store/src/internal/config/ngxs-config.ts
```diff
@@ -0,0 +1,48 @@
...
+   * - Object.freeze on the state and actions to guarantee immutability
+   * (default: false)
+   */
+  developmentMode: boolean;
+
+  /**
+   * Support a strict Content Security Policy.
+   * This will circumvent some optimisations that violate a strict CSP through the use of `new Function(...)`.
+   * (default: { strictContentSecurityPolicy: false })
+   */
```
markwhitfeld: This description is now against the wrong thing, it used to be on the `strictContentSecurityPolicy` property inside the compatibility options. Please move this interface definition to be inline again and move the defaults back to the constructor.

packages/store/src/symbols.ts
```diff
@@ -14,50 +10,8 @@ export const SELECTOR_META_KEY = 'NGXS_SELECTOR_META';
...
-   */
-  defaultsState: ObjectKeyMap<any> = {};
-
-  constructor() {
-    this.compatibility = {
-      strictContentSecurityPolicy: false
-    };
-    this.executionStrategy = DispatchOutsideZoneNgxsExecutionStrategy;
-  }
+export interface NgxsCompatibility {
```
markwhitfeld: This should rather remain inline in the NgxsConfig class as it was before.

packages/store/src/internal/config/ngxs-config.ts
splincode: I do not agree with the fact that we have to transfer everything to the constructor, since this is an unnecessary gesture, we already specify the default value in advance, and here it is immediately obvious!

splincode:
no advances in question

PR #834, opened 2019-02-11
Feature/reset plugin

armanozak:
## PR Checklist

## PR Type
```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Currently, a meta reducer can be used in order to clear/reset a state. Meta reducers are great, but they have the following drawbacks:
- In order to use a meta reducer, one must prepare and inject it as a plugin to NGXS. This is not as straightforward as importing a module.
- I could not find any blueprint for developers who want to keep some states while clearing others. For instance, when one uses the storage plugin on some states, the obviously intention is to keep them for the next visit. On the other hand, the same person would probably want to clear everything else on logout. I believe it will be nice to offer them a reusable way to do so.
- As far as I know, there is no direct way of resetting states to their defaults. This is a requirement in many cases and I have seen many developers struggling with it. Also, let's not forget sometimes even multiple states may have to be reset.
- There are some internal utils in NGXS which are quite helpful in achieving these 

## What is the new behavior?
With the new reset plugin, there are basically 2 things a developer can do in a direct way that is easy to understand and follow:
- Reset states by dispatching a predefined action with a payload of state classes. Child states will not be a problem thanks to `getValue` and `setValue` utility functions.
- Erase whole state by dispatching a predefined action. If state classes are provided to the action as payload, they will be kept untouched. Again, child states are considered.

```
[ ] Yes
[x] No
```

### DIFF
--- a/docs/plugins/reset.md
+++ b/docs/plugins/reset.md
@@ -1,0 +1,80 @@
+# Reset Plugin
+
+Erase the whole state tree anytime you want while keeping the state(s) you need. Reset a single state or multiple states easily back to default value(s).
+
+## Installation
+
+```bash
+npm install @ngxs/reset-plugin --save
+
+# or if you are using yarn
+yarn add @ngxs/reset-plugin
+```
+
+## Usage
+
+Import `NgxsResetPluginModule` into your root module like:
+
+```TS
+import { NgxsModule } from '@ngxs/store';
+import { NgxsResetPluginModule } from '@ngxs/reset-plugin';
+
+@NgModule({
+  imports: [
+    NgxsModule.forRoot([ /* Your states here */ ]),
+    NgxsResetPluginModule.forRoot()
+  ]
+})
+export class AppModule {}
+```
+
+### Options
+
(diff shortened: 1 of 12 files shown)


mehmet-erim:
I am waiting this feature. I think this is awesome

splincode:
@armanozak I do not understand why we need this plugin and what benefits it brings, create an issue and write your proposal

arturovt:
@armanozak Yeah, I also would like to know what problem this is solving

armanozak:
> @armanozak Yeah, I also would like to know what problem this is solving

This is providing an easy way to do something that is moderately difficult when a plugin like this is not presented. I thought the use was clearly explained on the pull request, but I guess it is not. I will open an issue as suggested by @splincode and we can discuss.

armanozak:
> @armanozak I do not understand why we need this plugin and what benefits it brings, create an issue and write your proposal

I opened [issue #835](https://github.com/ngxs/store/issues/835)

Issue #838, opened 2019-02-12
Improve dispatch UPDATE_STATE


splincode:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[ ] Performance issue
[x] Feature request
[ ] Documentation issue or request
[ ] Support request => https://github.com/ngxs/store/blob/master/CONTRIBUTING.md
[ ] Other... Please describe:
</code></pre>

## Current behavior

![image](https://user-images.githubusercontent.com/12021443/52637923-c2aabf00-2ee1-11e9-9e13-20c19cfdbba9.png)

It is not clear which module has initialized the state

## Expected behavior

Need send payload

splincode:
Released in 3.5.0

Issue #827, opened 2019-02-09
🚀 [FEATURE]: List of Breaking Changes planned for V4


markwhitfeld:
NGXS follows semantic versioning. As a result we will change the major version number when breaking changes are introduced. Many developers seem to have the idea that upping the major version number is a good thing because it shows progress and is a bit of a marketing opportunity. When following a semantic versioning scheme increasing the version number means that breaking changes have happened, which is not such a positive thing.  
Obviously making breaking changes for critical issues should not be avoided but when there are smaller annoyances that need to be remedied then they should be collected together and released together in a single major version update.  
This is one such list:

- [x] An `@Selector` with parameters in a state class should not receive a default first parameter of the enclosing class state
  - See: https://github.com/ngxs/store/issues/386#issuecomment-395780734) (#2117)
  - Related: #541 #446 
- [x] The plugin interfaces should be part of a `@ngxs/store/plugin` sub-package
- [ ] Review plugin architecture, possibly include more extensibility points than just middleware
- [x] NGXS internals currently exposed as part of the public api should be moved to the `@ngxs/store/internals` sub-package
- [ ] The `StateContext.getState` should return a `DeepReadOnly<>` interface (#2110)
- [x] The value passed in the observable returned from a `dispatch` call should be `void`, matching its interface (#2109 )
- [x] `patchState` and `setState` should both return void to match their interfaces (#491) (#2114)
- [x] The web socket plugin should have consistent Title casing of the word WebSocket (#825 #820) (#2115)
- [x] The web socket plugin should not expose `NGXS_WEBSOCKET_OPTIONS` in its public api (#2115)
- [x] The storage plugin should not expose tokens in its public api
- [x] Improve public interface type safety
- [x] Disable suppress errors in selector (the default `suppressErrors` should be false) (#2111)
- [x] Default the developmentMode setting to align with the angular dev mode.
  - Also see https://github.com/ngxs/store/pull/1741#issuecomment-842953704
- [x] The ofActionErrored pipe should return the error as well as the action (change return value to the same signature which is returned by `ofActionCompleted`. Where `success` and `cancel` should be `false`) (#2112)
- [x] The `Actions` stream has to have `ActionContext` signature except of `any`.
- [x] Flatten the `.d.ts` file exports of the library (https://github.com/ngxs/store/issues/827#issuecomment-867911205) (#2131)
- [x] Deprecation notice for `@Select`. We would need a migration page in the recipes sections of the docs for both of these.
[...]

PR #845, opened 2019-02-16
fix(store): add state names to update state

eranshmil:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: #838 

## What is the new behavior?
State names array is being added to the `UpdateState` action, so they will be visible using the `logger` and `devtools` plugins.
The `states` parameter is optional in order to prevent a breaking change, in case some developer used this action directly. We might consider changing that to be required for version 4. #827 

image
image

```
[ ] Yes
[x] No
```

### DIFF
--- a/packages/store/src/actions/actions.ts
+++ b/packages/store/src/actions/actions.ts
@@ -1,3 +1,5 @@
+import { ObjectKeyMap } from '../internal/internals';
+
 /**
  * Init action
  */
@@ -16,4 +18,6 @@ export class UpdateState {
     // NOTE: Not necessary to declare the type in this way in your code. See https://github.com/ngxs/store/pull/644#issuecomment-436003138
     return '@@UPDATE_STATE';
   }
+
+  constructor(public payload?: ObjectKeyMap<any>) {}
 }
--- a/packages/store/tests/state.spec.ts
+++ b/packages/store/tests/state.spec.ts
@@ -140,6 +140,48 @@ describe('State', () => {
       expect(TestBed.get(Store).snapshot().foo).toEqual(['initState', 'onInit']);
     });
 
+    it('should call an UpdateState action handler with multiple states', () => {
+      const expectedStates = { foo: ['test'], bar: 'baz', qux: { key: 'value' } };
+
+      @State<any>({
+        name: 'eager',
+        defaults: []
+      })
+      class EagerState {
+        @Action(UpdateState)
+        updateState(ctx: StateContext<any[]>, action: UpdateState) {
+          ctx.setState({ ...ctx.getState(), ...action.payload });
+        }
+      }
+
+      @State<string[]>({
+        name: 'foo',
+        defaults: expectedStates.foo
+      })
+      class FooState {}
+
+      @State<string>({
+        name: 'bar',
+        defaults: expectedStates.bar
+      })
+      class BarState {}
+
+      @State<any>({
+        name: 'qux',
+        defaults: expectedStates.qux
+      })
+      class QuxState {}
+
+      TestBed.configureTestingModule({
+        imports: [
+          NgxsModule.forRoot([EagerState]),
+          NgxsModule.forFeature([FooState, BarState, QuxState])
+        ]
+      });
+
+      expect(TestBed.get(Store).snapshot().eager).toEqual(expectedStates);
+    });
+
     it('should call an UpdateState action handler before the ngxsOnInit method on feature module initialisation', () => {
       @State<string[]>({
         name: 'foo',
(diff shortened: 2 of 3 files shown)


deebloo:
im not sure I fully understand the purpose here.

splincode:
@deebloo https://github.com/ngxs/store/issues/838
when I have a lot of lazy modules, it is not clear which one was loaded

eranshmil:
Updated. Please review again.

packages/store/src/actions/actions.ts
```diff
@@ -16,4 +16,6 @@ export class UpdateState {
     // NOTE: Not necessary to declare the type in this way in your code. See https://github.com/ngxs/store/pull/644#issuecomment-436003138
     return '@@UPDATE_STATE';
   }
+
+  constructor(public payload?: any) {}
```
splincode: use `ObjectKeyMap<any>` instead any

eranshmil: I'll change that.

eranshmil:
> add also test to check payload
> 
> ```ts
> ofActionDispatched(UpdateState). subscribe(payload => expect(payload).toEqual(..))
> ```

The spy (@markwhitfeld, am I right now?) is a replacement for that check.

eranshmil:
I changed the payload type to `ObjectKeyMap<any>`.
@markwhitfeld, please review the updated code. (This is not the only change since your previous review)

packages/store/tests/state.spec.ts
```diff
@@ -140,6 +140,48 @@ describe('State', () => {
       expect(TestBed.get(Store).snapshot().foo).toEqual(['initState', 'onInit']);
     });
 
+    it('should call an UpdateState action handler with multiple states', () => {
```
splincode: I don't quite understand how this test covers what your  pull request does

I thought you would just add

```ts
it(' ... ', fakeAsync(() => {

  store.dispatch(new UpdateState([ 'MY_PAYLOAD' ]));
  tick(100);

  actions$
      .pipe(ofActionDispatched(UpdateState))
      .subscribe(({ payload }) => expect(payload).toEqual('MY_PAYLOAD'));

}))
```

eranshmil: This unit tests shouldn't be dependent on the actions stream.
Take a look at the other tests in the file.

splincode: You can always create a separate file with the test, but your code has not tested this case.

arturovt: @splincode wouldn't it be the same what he tests right now? I mean maybe I'm unaware of something in this discussion, but it seems like the result will be the same, no?

splincode: ok

splincode:
@eranshmil any news?

eranshmil:
> @eranshmil any news?

Waiting for @markwhitfeld review. :)

splincode:
@eranshmil fix conflict

splincode:
@markwhitfeld ping

markwhitfeld:
@splincode why did you change the title from `feat(store): add state names to update state` to `fix(store): add state names to update state`?
This is definitely a feature and not a bugfix.

splincode:
@markwhitfeld because we did not add a new class, but just modified it, I opened the task as a bug :)

https://github.com/ngxs/store/issues/838

markwhitfeld:
@splincode We added a payload to the `UpdateState` class. This is definitely new behaviour.
Do you feel like this PR fulfills the brief?
It had the state names at first and now returns the default values. I feel like it has lost some info about the state names. Is just the defaults sufficient info?

splincode:
@markwhitfeld okey :) I agree that this is a feature

splincode:
it seems to me that it is quite correct to display the payload in the form of the tree that we will use

markwhitfeld:
Just wondering if we should be using the `payload` name for the prop or if we should rather call it `defaults` (or something else that makes sense) to keep our options open for other info if needed in future.
@eranshmil what do you think?

eranshmil:
> Just wondering if we should be using the `payload` name for the prop or if we should rather call it `defaults` (or something else that makes sense... maybe `states` or `defaultsMerged` or `statesMerged` or `statesAdded`... suggestions please @ngxs/core ).
> The reason for this is to keep our options open for other info if needed in future (if we use the `payload` prop then we have boxed ourselves in).
> @eranshmil what do you think?

Agree, maybe `addedStateDefaults`.

markwhitfeld:
`addedStateDefaults` seems good to me. @splincode what do you think?

splincode:
@markwhitfeld too long and complicated

PR #846, opened 2019-02-16
fix(store): add static implements for store action

splincode:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

image

## What is the new behavior?

Ability to specify an interface for static fields for actions

```
[ ] Yes
[x] No
```

### DIFF
--- a/packages/form-plugin/src/actions.ts
+++ b/packages/form-plugin/src/actions.ts
@@ -1,22 +1,24 @@
+import { StaticAction } from '@ngxs/store';
+
+@StaticAction()
 export class UpdateFormStatus {
   static get type() {
-    // NOTE: Not necessary to declare the type in this way in your code. See https://github.com/ngxs/store/pull/644#issuecomment-436003138
     return '[Forms] Update Form Status';
   }
   constructor(public payload: { status: string | null; path: string }) {}
 }
 
+@StaticAction()
 export class UpdateFormValue {
   static get type() {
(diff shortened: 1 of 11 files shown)


splincode:
@markwhitfeld 

If you comment out this code

image

The code will not be compiled! Thus, we do not need to write meaningless comments, but simply use the desired decorator for service shots.

image

splincode:
@markwhitfeld In your case, even if people would write a comment, but confused and did not add a static property, the code would compile, but this would not work for users. You yourself care about the clean code, however, comments in code that can be easily changed or removed by someone will not protect you, unlike this type-safe decorator.

splincode:
@arturovt @eranshmil what do you think?

markwhitfeld:
The comment is detailing a stop gap measure which is in place while we are still supporting TS 2.x
PS. I am also investigating if we could possibly make the `type` property on actions optional so not to keen to introduce a new construct around that property.

deebloo:
I am getting concerned about API surface size. Is this absolutely necessary?

markwhitfeld:
Yeah, my concern as well.

eranshmil:
> @arturovt @eranshmil what do you think?

I think we should leave it for now.
I'm planning to check whether this is still necessary in order to support TS2.*.
No matter what, I guess we will have to drop support in the next year and then we could remove that.

Issue #446, opened 2018-06-15
Problem with state composition


foralobo:
```
* ngxs: 3.1.4
* @angular/core: 6.0.1
```

Hi, I have a request for you masters...

I tried to create a simple EntityAdpter class to use the Entity feature on ngxs. It works fine:
my store has the properties "entities" and "ids".

I have a problem: when I want return a composition of the two properties 
```
  @Selector()
  static orders(state: OrdersStateModel) { 
    return [...state.ids].map(id => state.entities[id]);
  }
```
it returns correctly but everytime I call a setState() on some other state property (different from ids and entities), the observable on this Selector intercepts a new value of "orders".

I think it is the right behavior of @Selector because I'am returning a completely new array!!
I'm resolving this problem so (for brevity I hide the entitydapter class):
```
// state
@Selector()
static ids(state: OrdersStateModel) {
  return state.ids
}

@Selector()
static entities(state: OrdersStateModel) {
  return state.entities
}

...

// container

@Select(OrdersState.ids) ids$: Observable<string[]>;
@Select(OrdersState.entities) entities$: Observable<any[]>;
// on init
this.elements = this.entities$.pipe(
  withLatestFrom(this.ids$),
  map(([ent, ids]) => {
    return ids.map(id => ent[id]);
  })
);
this.elements.subscribe(data => console.log('example', data));
```
I know it is very ugly, but I'am asking, is there a more elegant mode to workaround this?
I don't know, a mode to do the "withLatestFrom" inside the state...

Thank you very much..and thank you for your work.

foralobo:
Another possible solution is to use a child state to manage the entities operations. So we can divide the use of the two states.

I would like a decorator called @EntityState that it manages the state like an entities state.

markwhitfeld:
You could combine those selectors using a meta selector. 
See this: https://ngxs.gitbook.io/ngxs/concepts/select#meta-selectors 
The selector would need to be defined on a non state class.

foralobo:
I tested meta selector but it not returns values. It returns undefined (I used the example in the ngxs documentation)

markwhitfeld:
@foralobo is it possible for you to post a Stackblitz sample?

foralobo:
Ok It was a my error: It not worked because I was testing with an old version of store (3.0.1)
With 3.1.4 it works but it is not that i want. Every time I dispatch on the incriminated state, it send a value and my Select intercept it.

[...]

Issue #386, opened 2018-05-18
Discussion: Selector Composition and Query Classes


markwhitfeld:
## Discussion
The purpose of this issue is to facilitate a discussion around the possibilities for selectors and querying the application state. _Many ideas on this topic will be listed here._

### Versions

```
* ngxs: 3.0.1
* @angular/core: 3.0.0
```

### Items to agree on:
- [ ] Query Class concept
- [ ] State Class `@Selector(...)` method signatures
- [ ] Query Class `@Selector(...)` method signatures
- [ ] Allowed selector composition parameters
- [ ] Support for dynamic @Selector arguments
- [ ] @Selector options

I would like to try some rules to help facilitate voting on ideas ( because... why not! ):

- For a new idea, add a comment and give it a Title in bold at the top of the comment which people will use to reference to it. One idea per comment. If the idea has multiple options please split into multiple comments titled appropriately. 
- Vote on ideas using the Reactions to the comment. ie. 👍  for agreement, 👎 for disagreement.
- If you disagree with an idea and/or would like to add to it then add a comment referencing the comment with details as to why you disagree (unless someone has already stated a similar disagreement which you can agree to).

markwhitfeld:
**Query Class concept**
Regarding the case for the ability to have `@Selector(...)` on a non state class, here are some thoughts as a motivation in this regard...
This type of class could be called a Query Class. Your code would look something like this:
``` typescript
export class UserTodosQuery {
  @Selector([UserState.getCurrentUserId, TodoState])  
  getCurrentUserTodos(userId: string, todoState: TodoStateModel) { ... }

  @Selector([UserState.getCurrentUserId, TodoState.getInCompleteTodos])  
  getCurrentUserIncompleteTodos(userId: string, todos: Todo[]) { ... }
}
```
You would do this where you want to combine two selectors across two states that don't have a common route. I've also seen people mention that they want this sort of thing for feature modules so that they can combine states between 2 distinct feature modules. You wouldn't want to force a dependency inside the feature module to another for the sake of a selector that you only need when you are using both feature modules. This type of class promotes decoupling from the State.

The simplicity of the selectors and selector composition is very valuable when combining state across different parts of the app. Allowing for composed Selectors in non-state classes allows for a clean separation of selectors if someone prefers to keep them separate from the state as a part of their application style (command and query code separation).

[...]

PR #858, opened 2019-02-20
feat(store): add @Selector parameter to optimize updates on state class

sinedied:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

By default, the `Selector` decorator always inject the container state class as its first parameter, causing unwanted update when selectors are chained.

Here's an example use case:
```ts
@Selector()
static items(state: ItemStateModel): Item[] {  return state.items;  }

@Selector([ItemState.items])
static filteredItems(_state: ItemStateModel, items: Item[]): Item[] { // ... (filter items) }

@Selector([ItemState.filteredItems])
static sortedItems(_state: ItemStateModel, filteredItems: Item[]): Item[] { //  ... (sort filtered items)  }
```
There is actually no way to get rid of `_state: ItemStateModel` injection while keeping these selectors in the `ItemState` class. Worse, it degrades performance as all these 3 selectors gets recomputed when any of the properties on the `ItemState` changes.

Issue Number: N/A

## What is the new behavior?

This PR introduces a second argument to the `Selector` decorator, allowing to skip injection of container class, allowing desired behavior:
```ts
@Selector()
static items(state: ItemStateModel): Item[] {  return state.items;  }

@Selector([ItemState.items], true)
static filteredItems(items: Item[]): Item[] { // ... (filter items) }

@Selector([ItemState.filteredItems], true)
static sortedItems(filteredItems: Item[]): Item[] { //  ... (sort filtered items)  }
```
Now `filteredItems` and `sortedItems` only gets recomputed if `items` and `filteredItems` change respectively, not on any state property change.

```
[ ] Yes
[x] No
```

## Other information
Let me know if you prefer to change the boolean parameter by an explicit parameter object like `{ skipContainerInjection: true }`

### DIFF
--- a/packages/store/src/decorators/selector.ts
+++ b/packages/store/src/decorators/selector.ts
@@ -3,7 +3,7 @@ import { createSelector } from '../utils/selector-utils';
 /**
  * Decorator for memoizing a state selector.
  */
-export function Selector(selectors?: any[]) {
+export function Selector(selectors?: any[], skipContainerSelector: boolean = false) {
   return (target: any, methodName: string, descriptor: PropertyDescriptor) => {
     if (descriptor.value !== null) {
       const originalFn = descriptor.value;
@@ -11,7 +11,9 @@ export function Selector(selectors?: any[]) {
       const memoizedFn = createSelector(
         selectors,
         originalFn.bind(target),
-        { containerClass: target, selectorName: methodName }
+        skipContainerSelector
+          ? undefined
+          : { containerClass: target, selectorName: methodName }
       );
 
       return {
--- a/docs/concepts/select.md
+++ b/docs/concepts/select.md
@@ -268,6 +268,25 @@ When using the `Selector` decorator along with a state class, it will still
 inject the state class's state first followed by the other selectors in the order
 they were passed in the signature.
 
+#### Optimizing Selectors
+By default the state class's state is always injected first, causing your selector
+value to be updated on each state's change.
+
+You can change this behavior passing `true` as the second selector parameter, to
+only inject the selectors you need and therefore limit updates.
+
+```TS
+@State<string[]>({ ... })
+export class ZooState {
+
+  @Selector([ZooState.pandas('baby')], true)
+  static babyPandasCounter(babyPandas: string[]) {
+    return babyPandas.length;
+  }
+
+}
+```
+
 ### Meta Selectors
 By default selectors in NGXS are bound to a state. Sometimes you need the ability
 to join to un-related states in a high-performance re-usable fashion. A meta selector
--- a/packages/store/tests/selector.spec.ts
+++ b/packages/store/tests/selector.spec.ts
@@ -37,6 +37,11 @@ describe('Selector', () => {
     static fooBar(myState2: any, foo: any) {
       return foo + myState2.bar;
     }
+
+    @Selector([MyState2.foo], true)
+    static fooOnly(foo: any) {
+      return foo;
+    }
   }
 
   class MetaSelector {
@@ -96,6 +101,16 @@ describe('Selector', () => {
       const store: Store = TestBed.get(Store);
       const slice = store.selectSnapshot(MyState2.fooBar);
       expect(slice).toBe('HelloHelloWorld');
+    }));
+
+    it('should select only specified', async(() => {
+      TestBed.configureTestingModule({
+        imports: [NgxsModule.forRoot([MyState, MyState2])]
+      });
+
+      const store: Store = TestBed.get(Store);
+      const slice = store.selectSnapshot(MyState2.fooOnly);
+      expect(slice).toBe('HelloHello');
     }));
 
     it('context should be defined inside selector', () => {


splincode:
Can you add a test to the fact that the number of updates has decreased?

packages/store/src/decorators/selector.ts
```diff
@@ -3,15 +3,17 @@ import { createSelector } from '../utils/selector-utils';
 /**
  * Decorator for memoizing a state selector.
  */
-export function Selector(selectors?: any[]) {
+export function Selector(selectors?: any[], skipContainerSelector: boolean = false) {
```
arturovt: `:boolean` is not needed

packages/store/src/decorators/selector.ts
```diff
@@ -3,15 +3,17 @@ import { createSelector } from '../utils/selector-utils';
...
   return (target: any, methodName: string, descriptor: PropertyDescriptor) => {
     if (descriptor.value !== null) {
       const originalFn = descriptor.value;
 
       const memoizedFn = createSelector(
         selectors,
         originalFn.bind(target),
-        { containerClass: target, selectorName: methodName }
+        skipContainerSelector
```
arturovt: could you create a separate variable for that:
```TS
const creationMetadata = !skipContainerSelector && { containerClass: target, selectorName: methodName };
```

and pass it

```TS
createSelector(selectors, originalFn.bind(target), creationMetadata);
```

splincode:
```ts
interface SelectorOptions {
 skipContainer: boolean;
}
```

```ts
@Selector([ .. ], options: SelectorOptions )
```

markwhitfeld:
This is a known issue. 
See this https://github.com/ngxs/store/issues/386#issuecomment-395780734
Related: #446
As a workaround see here: https://ngxs.gitbook.io/ngxs/concepts/select#meta-selectors

This will be fixed in v4 (as a breaking change) see list here: https://github.com/ngxs/store/issues/827

I want to add a global config flag for to modify this behaviour in v3.5. Something like this:
```TS
@NgModule({
  imports: [
    NgxsModule.forRoot(states, {
      developmentMode: !environment.production,
      compatibility: {
        v4: true
      }
    })
  ]
})
export class AppModule {}
```
Then when v4 is released we would have this option to provide temporary backward compatibility:
```TS
@NgModule({
  imports: [
    NgxsModule.forRoot(states, {
      developmentMode: !environment.production,
      compatibility: {
        v3: true
      }
    })
  ]
})
export class AppModule {}
```

This is my current thinking, but this is an interesting option to provide granularity for this setting.
Not sure if I like the expansion of the `@Selector` decorator to include options though.

@ngxs/core team... what are your opinions?

splincode:
@markwhitfeld You mean you want to make a global config for Selector?

eranshmil:
@markwhitfeld What this flag will really do? The name sounds like it will turn on all v4 features in v3.

markwhitfeld:
The idea is that it will allow for the user to opt in for some v4 featrues before we release v4.
Maybe there is a better named option?

eranshmil:
> The idea is that it will allow for the user to opt in for some v4 features before we release v4.
> Maybe there is a better named option?

One flag for backward compatibility might be problematic if someone only wants to disable the selector breaking change but want the other features.
Is that flag will disable/enable new features, or just enable breaking changes in v3/disable breaking changes in v4?

markwhitfeld:
Changing the way that the first parameter is determined in the `@Selector` is a highly impactful change in v4, so I'm looking for a way to allow someone to take advantage of it early or to defer the breaking change once they have upgraded.
Maybe some granularity here would be better... something like this:
To take on the v4 behaviour early in v3.5:
```TS
{ 
  compatibility: { 
    selectors: { v4: true }
  }
}
```
To disable breaking change in v4 (temporarily - would display a console warning of deprecation)
```TS
{ 
  compatibility: { 
    selectors: { v3: true }
  }
}
```

sinedied:
@markwhitfeld I'm all for making it a default as you suggested in the list of breaking changes for v4.

I'm also willing to update this PR to remove the extra parameter and use a global feature flag instead.
However, as far as feature flag go it would be better in my opinion to have an explicit named flag that won't break through the v3 -> v4 update if it's set, letting only its default value change:
```
{ 
  compatibility: { 
    skipSelectorContainer: true
  }
}
```
This way, only its default value will change from v3 to v4, keeping your choice through the upgrade.

FYI, this is the same way adopted by Angular for breaking changes with backward compatibility (for example regarding the `preserveWhitespace` option).

markwhitfeld:
hmmm... you make a good point.
Maybe this would be a better naming:
```
{ 
  compatibility: { 
    injectSelectorContainer: true  // default changes to false in v4
  }
}
```
I think that we need to get that name right so that it makes the setting obvious. 
Do you like this name? What other options do we have?

sinedied:
Ok I have almost finished changing the implementation to match all your remarks and use the compatibility setting instead of a parameter, but I'm struggling to solve one problem: how to access the `NgxsConfig` within the `@Selector` decorator implementation, as it's applied on static methods that gets called before everything else (including DI setup)?

I've tried using a similar approach than the one used for the `@Select` decorator, but it's not working here as the selector is applied on static and not instance methods.

Any hints?

splincode:
@sinedied Unfortunately, this is impossible, I also tried

sinedied:
Awwww 😭 Which approach would you suggest then, keeping the second optional parameter?

sinedied:
What about using a `@ngxs/store/[next|compat]` path to import a variant of the `@Selector` annotation?

ie instead of `import { Selector } from '@ngxs/store';` we would use `import { Selector } from '@ngxs/store/next';` in v3, then when v4 is out the old one would be `import { Selector } from '@ngxs/store/compat';`.

What do you think?

markwhitfeld:
Hmm... interesting idea. Would still require quite comprehensive code file changes though.
It does give me an idea that it could be configured through a Module import as opposed to another option on the `forRoot`. That is if we decide that this setting is just a temporary transition flag and not a permanent feature flag.

I'm certain that I can get the global setting to work. We just need to fetch the config at a later stage.
I will have a look over the weekend.

splincode:
I like it ```@ngxs/store/next``` - new features

sinedied:
@markwhitfeld If you think it could work by delaying the config fetch, feel free to start from my branch that have all the other changes (doc, config, RecursivePartial<>, etc) already done, just the config fetch failing: https://github.com/sinedied/store/blob/config-compat/packages/store/src/decorators/selector.ts

FYI I tried to add a proxy function to delay the actual call, but could never make it work properly.

sinedied:
@markwhitfeld Did you had the chance to try accessing the config? Or should I give a try updating this PR with `@ngxs/store/next` approach?

markwhitfeld:
Hi @sinedied I have not had a chance to look at this yet because we are working on the v3.4 release and this has taken up all of my attention. I don't think we will use the `@ngxs/store/next` approach for now, but I need to discuss with the team first. 
I think that the first step though is to add the option and then to figure out how best to surface it to the user. We had a team meeting scheduled on the weekend but unfortunately it had to be postponed. I will discuss this with the team.

sinedied:
Ok, let me know once the choice is settled and I will update the PR consequently.

splincode:
now you have this opportunity

```ts
@NgModule({
 imports: [
    NgxsModule.forRoot(states, {
      developmentMode: true
      selectorOptions: { injectContainerState: false } // default false everywhere
    })
 ]
})
export class AppModule {}
```

Issue #871, opened 2019-02-26
logger-plugin should log payload for all variable names (not just 'payload')


pvamshi:
## I'm submitting a...
<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[ ] Performance issue
[ x ] Feature request
[ ] Documentation issue or request
[ ] Support request => https://github.com/ngxs/store/blob/master/CONTRIBUTING.md
[ ] Other... Please describe:
</code></pre>

## Current behavior
If my Action has a structure like this 
```typescript
export class AddUser {
    readonly type = '[User page] Add user'; 
    constructor(public user: User){}   // ---> This wont appear in log 
}
```

When I dispatch the action. Logger ignores the user data . It  works only if the payload is in 'payload' param like this : 
```typescript
export class AddUser {
    readonly type = '[User page] Add user'; 
    constructor(public payload: User){} // ---> This appears in log 
}
```

## Expected behavior
Logger should log payload for any params. I guess we should even allow multiple params.

## Minimal reproduction of the problem with instructions
Rename , payload key to anything other than 'payload' and check logs. 

## What is the motivation / use case for changing the behavior?
Having custom names make it self-explanatory on what the action contains and useful while handling in store. 

Also useful in destructuring the Action. 
```ts
@Action(AddUser)
addUser(
    { getState, setState, dispatch }: StateContext<ConfigStateModel>,
    { user }: AddUser,
) {
...
}
```

## Environment

<pre><code>
Libs:
- @angular/core version: any
- @ngxs/logger-plugin version: 3.3.4

Browser:
- [ x ] Chrome (desktop) version 72
- [ ] Chrome (Android) version XX
- [ ] Chrome (iOS) version XX
- [ ] Firefox version XX
- [ ] Safari (desktop) version XX
- [ ] Safari (iOS) version XX
- [ ] IE version XX
- [ ] Edge version XX

markwhitfeld:
The fix has been released in v3.4.3!

PR #879, opened 2019-02-28
fix(logger): print action properties

eranshmil:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

```ts
export class AddTodo {
  public static type = 'AddTodo';
  constructor(public todo: Todo) {}
}
```
The logger doesn't print action's payload if there's no property called `payload`.
image

Issue Number: #871 

## What is the new behavior?
If the action has at least one property with a value, it will be printed to the console.
image

```
[ ] Yes
[x] No
```

### DIFF
--- a/packages/logger-plugin/src/logger.plugin.ts
+++ b/packages/logger-plugin/src/logger.plugin.ts
@@ -35,8 +35,9 @@ export class NgxsLoggerPlugin implements NgxsPlugin {
       console.log(message);
     }
 
-    if (typeof event.payload !== 'undefined') {
-      this.log('payload', 'color: #9E9E9E; font-weight: bold', event.payload);
+    // print payload only if at least one property is supplied
+    if (this._hasPayload(event)) {
+      this.log('payload', 'color: #9E9E9E; font-weight: bold', { ...event });
     }
 
     this.log('prev state', 'color: #9E9E9E; font-weight: bold', state);
@@ -86,5 +87,11 @@ export class NgxsLoggerPlugin implements NgxsPlugin {
     }
 
     return ms_ie;
+  }
+
+  private _hasPayload(event: any) {
+    const nonEmptyProperties = Object.entries(event).filter(([, value]) => !!value);
+
+    return nonEmptyProperties.length > 0;
   }
 }
--- a/packages/logger-plugin/tests/helpers/utils.ts
+++ b/packages/logger-plugin/tests/helpers/utils.ts
@@ -4,7 +4,7 @@ export class FormatActionCallStackOptions {
   action: string;
   prevState: {};
   nextState?: {};
-  payload?: string;
+  payload?: any;
   error?: string;
   collapsed?: boolean;
 }
--- a/packages/logger-plugin/tests/logger.plugin.spec.ts
+++ b/packages/logger-plugin/tests/logger.plugin.spec.ts
@@ -15,7 +15,7 @@ describe('NgxsLoggerPlugin', () => {
   class UpdateBarAction {
     static type = 'UPDATE_BAR';
 
-    constructor(public payload?: string) {}
+    constructor(public bar?: string) {}
   }
 
   class ErrorAction {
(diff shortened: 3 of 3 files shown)


packages/logger-plugin/src/logger.plugin.ts
```diff
@@ -87,4 +88,10 @@ export class NgxsLoggerPlugin implements NgxsPlugin {
 
     return ms_ie;
   }
+
+  private _hasPayload(event: any) {
+    const nonEmptyProperties = Object.entries(event).filter(entry => !!entry[1]);
```
splincode: `entry[1]` ? 

can give a meaningful name

packages/logger-plugin/src/logger.plugin.ts
```diff
@@ -87,4 +88,10 @@ export class NgxsLoggerPlugin implements NgxsPlugin {
 
     return ms_ie;
   }
+
+  private _hasPayload(event: any) {
```
splincode: use `hasPayload` instead `_hasPayload`

eranshmil: Most of the core codebase is using the `_` prefix.

packages/logger-plugin/src/logger.plugin.ts
eranshmil: `property` maybe?

packages/logger-plugin/src/logger.plugin.ts
splincode: I do not respect this approach, unfortunately and very upset that everywhere it is used

Why then did the access modifier come up?

packages/logger-plugin/src/logger.plugin.ts
splincode: I just always get embarrassed when people go to the index, why the first index?

eranshmil: entry returns `[ propertyName, 'propertyValue' ]`

packages/logger-plugin/src/logger.plugin.ts
eranshmil: From the beginning ;)
https://github.com/ngxs/store/commit/372377ff20aac74e61750534bc8503cbf00d838d

splincode: typescript private access modifier most prefer 

the modifier is usually enough to understand that the method is private, I write in Java and absolutely do not understand why people come up with all these tricks

eranshmil: When reading code, it makes it easier to understand the access modifier from the name, instead of using the mouse to scroll up (in PRs) or to move the mouse over the method name (in the IDE).

deebloo: it was part of the original code design. If we want to make a design change we should do that as part of a separate task

deebloo: is the type of event really any?

packages/logger-plugin/src/logger.plugin.ts
splincode: prefer

```ts
const nonEmptyProperties = Object.entries(event).filter(([ value ]) => !!value);
```

packages/logger-plugin/src/logger.plugin.ts
eranshmil: The `event` is the name of the second parameter in the `handle` method of the plugin.
This is the name given for the parameter `action` in the `NgxsPlugin` interface:
```ts
handle(state: any, action: any, next: NgxsNextPluginFn): any;
```

packages/logger-plugin/src/logger.plugin.ts
arturovt: @splincode `[, value]`

splincode:
any news?

eranshmil:
> any news?

Waiting for @markwhitfeld and @arturovt review.

splincode:
@markwhitfeld ping

PR #884, opened 2019-03-02
check-integration-test-is-works

splincode:


splincode:
@eranshmil integration test works fine and checks regression

Issue #883, opened 2019-03-01
Action is called multiple times when a module is called by multiple routes (3.4.0 issues)


junimohano:
## I'm submitting a...

<pre><code>
[x] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[ ] Performance issue
[ ] Feature request
[ ] Documentation issue or request
[ ] Support request => https://github.com/ngxs/store/blob/master/CONTRIBUTING.md
[ ] Other... Please describe:
</code></pre>

## Current behavior
Action is called multiple times when same modules which contains a feature state are called by different routes on lazy load.

## Expected behavior
Action needs to be called only one time even though same modules which contains a feature state are called by different routes on lazy load.

NGXS 3.3.4 is fine, only 3.4.0 has a issue.

## Minimal reproduction of the problem with instructions

[https://github.com/junimohano/reproduce-ngxs-bugs](https://github.com/junimohano/reproduce-ngxs-bugs)

* Navigate to Add page
* Click Open Alert button then it will show alert message
* Navigate to Edit page
* Click Open Alert button then it will show alert message two times (should be called one time).
* Navigate to Detail page
* Click Open Alert button then it will show alert message three times (should be called one time).

It is happened if a module has a feature state which is called by different routes on lazy load.

NGXS 3.3.4 is fine but 3.4.0 has issues.

## What is the motivation / use case for changing the behavior?
I updated ngxs 3.4.0 on my company project and found a bug during QA testing.

## Environment

Libs:
- @angular/core version: 7.2.0
- @ngxs/store version: 3.4.0

Browser:
- [x] Chrome (desktop) version 72.0.3626.121
- [ ] Chrome (Android) version XX
- [ ] Chrome (iOS) version XX
- [ ] Firefox version XX
- [ ] Safari (desktop) version XX
- [ ] Safari (iOS) version XX
- [x] IE version 11.136
- [x] Edge version 44.17763.1.0

splincode:
@junimohano Please describe according to the issue template.

# Current behavior 

# Expected behavior

----

junimohano:
> @junimohano Please describe according to the issue template.
> 
> # Current behavior
> # Expected behavior

Updated

splincode:
@junimohano I fixed the bug, now I will add tests and make a pull request

we will release the fix for 3.4.1

splincode:
@junimohano try 3.4.1

junimohano:
@splincode works, thanks

[...]

PR #888, opened 2019-03-02
fix(store): duplication of deferred states

splincode:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

https://github.com/ngxs/store/issues/883

```
[ ] Yes
[x] No
```

### DIFF
--- a/packages/store/src/internal/state-factory.ts
+++ b/packages/store/src/internal/state-factory.ts
@@ -93,39 +93,38 @@ export class StateFactory {
    */
   add(stateClasses: StateClass[]): MappedStore[] {
     StateFactory.checkStatesAreValid(stateClasses);
-    const unmountedStateList: StateClass[] = this.getUnmountedStateList(stateClasses);
+    const { newStates } = this.addToStatesMap(stateClasses);
+    if (!newStates.length) return [];
+
+    const stateGraph: StateKeyGraph = buildGraph(newStates);
+    const sortedStates: string[] = topologicalSort(stateGraph);
+    const depths: ObjectKeyMap<string> = findFullParentPath(stateGraph);
+    const nameGraph: ObjectKeyMap<StateClass> = nameToState(newStates);
     const bootstrappedStores: MappedStore[] = [];
 
-    if (unmountedStateList.length) {
-      const stateGraph: StateKeyGraph = buildGraph(unmountedStateList);
-      const sortedStates: string[] = topologicalSort(stateGraph);
-      const depths: ObjectKeyMap<string> = findFullParentPath(stateGraph);
-      const nameGraph: ObjectKeyMap<StateClass> = nameToState(unmountedStateList);
+    for (const name of sortedStates) {
+      const stateClass: StateClass = nameGraph[name];
+      const depth: string = depths[name];
+      const meta: MetaDataModel = stateClass[META_KEY]!;
 
-      for (const name of sortedStates) {
-        const stateClass: StateClass = nameGraph[name];
-        const depth: string = depths[name];
-        const meta: MetaDataModel = stateClass[META_KEY]!;
+      this.addRuntimeInfoToMeta(meta, depth);
 
-        this.addRuntimeInfoToMeta(meta, depth);
+      const stateMap: MappedStore = {
+        name,
+        depth,
+        actions: meta.actions,
+        instance: this._injector.get(stateClass),
+        defaults: StateFactory.cloneDefaults(meta.defaults)
+      };
 
-        const stateMap: MappedStore = {
-          name,
-          depth,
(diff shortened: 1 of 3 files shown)


packages/store/src/internal/state-factory.ts
```diff
@@ -218,6 +215,20 @@ export class StateFactory {
     return forkJoin(results);
   }
 
+  private checkForDuplicateStateNames(stateClasses: StateClass[]): StateClass[] {
+    const filteredStates: StateClass[] = [];
+    for (const stateClass of stateClasses) {
+      const stateName = StoreValidators.checkStateNameIsUnique(stateClass, this.statesByName);
+      const notExistInTree: boolean = !this.statesByName[stateName];
```
arturovt: `doesntExistInTree`

splincode: done

packages/store/src/internal/state-factory.ts
```diff
@@ -218,6 +215,22 @@ export class StateFactory {
...
+  private getUnmountedStateList(stateClasses: StateClass[]): StateClass[] {
+    const unmountedStateList: StateClass[] = [];
+    const statesMap: StatesByName = this.statesByName;
+
+    for (const stateClass of stateClasses) {
+      const stateName: string = StoreValidators.checkStateNameIsUnique(stateClass, statesMap);
+      const unmountedState: boolean = !statesMap[stateName];
+      if (unmountedState) {
+        unmountedStateList.push(stateClass);
+        statesMap[stateName] = stateClass;
```
markwhitfeld: This method is confusing because it has side effects of modifying the `this.statesByName` but it is named `get...` which implies determinism but does not fulfill that promise.

packages/store/src/internal/state-factory.ts
```diff
@@ -56,10 +54,14 @@ export class StateFactory {
     private _internalStateOperations: InternalStateOperations
   ) {}
 
+  private _states: MappedStore[] = [];
```
markwhitfeld: We should stick to the established convention that private fields are declared at the top.

packages/store/src/internal/state-factory.ts
markwhitfeld: Fixed now by naming the method `addToStatesMap` and adding a descriptive return value.

packages/store/src/internal/state-factory.ts
markwhitfeld: Fixed in my commit

packages/store/src/internal/state-factory.ts
```diff
@@ -84,17 +84,22 @@ export class StateFactory {
...
+
   /**
    * Add a new state to the global defs.
    */
   add(stateClasses: StateClass[]): MappedStore[] {
-    this.checkStatesAreValid(stateClasses);
-    this.checkForDuplicateStateNames(stateClasses);
+    StateFactory.checkStatesAreValid(stateClasses);
+    const { newStates } = this.addToStatesMap(stateClasses);
+    if (!newStates.length) return [];
```
splincode: I stick to one rule that makes it easier to debug
`1 method = 1 return`

packages/store/src/internal/state-factory.ts
splincode: IDEA automatically formats, then configure tslint

packages/store/src/internal/state-factory.ts
splincode: Goog catch

packages/store/src/internal/state-factory.ts
markwhitfeld: This is a guard clause that reduces the nesting within the method.
It is a tradeoff between the two, and reduced nesting is the better one to opt for in my opinion.
(we have had this conversation before... guard clauses are the one exception to the 1 mthod 1 return rule from Code Complete)

packages/store/src/internal/state-factory.ts
markwhitfeld: Yeah, your auto formatting keeps doing things it shouldn't 😕

packages/store/src/internal/state-factory.ts
```diff
@@ -218,6 +212,22 @@ export class StateFactory {
     return forkJoin(results);
   }
 
+  private addToStatesMap(stateClasses: StateClass[]): { newStates: StateClass[] } {
```
splincode: { newStates: StateClass[] } - create interface

markwhitfeld: Not needed. This is only part of this one method's signature.

Issue #893, opened 2019-03-04
[Router plugin] - Feature request - state selector generic


beagleknight:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[ ] Performance issue
[x] Feature request
[ ] Documentation issue or request
[ ] Support request => https://github.com/ngxs/store/blob/master/CONTRIBUTING.md
[ ] Other... Please describe:
</code></pre>

## Current behavior

I am using a custom `RouterStateSerializer` that includes `url`, `params` and `queryParams`. After upgrading from ngxs 3.3.4 to 3.4.1 both `select` and `selectSnapshot` returns a `RouterState` type like this:

```typescript
// This is an error in 3.4.1 but works fine in 3.3.4
const { queryParams } = this.store.selectSnapshot(RouterState.state);
```

The reason behind this is the current implementation of the `RouterState.state` returns a `RouterStateModel` type and by default it includes a `state` member of type `RouterStateSnapshot` that doesn't include the `queryParams`.

## Expected behavior

If I have a custom `RouterStateSerializer` that serializes the state to this:

```typescript
export interface RouterStateParams {
  url: string;
  params: Params;
  queryParams: Params;
}
```

I expect the following code to work:
```typescript
const { queryParams } = this.store.selectSnapshot(state => RouterState.state<RouterStateParams>(state));
```

## What is the motivation / use case for changing the behavior?

Using a custom `RouterStateSerializer` I don't have the advantage of the type system when selecting the router's state. The current work around is using this ugly code:

```typescript
const { queryParams } = this.store.selectSnapshot(RouterState.state) as unknown as RouterStateParams;
```

## Environment

<pre><code>
Libs:
- @angular/core version: X.Y.Z
- @ngxs/store version: X.Y.Z

Browser:
- [ ] Chrome (desktop) version XX
- [ ] Chrome (Android) version XX
- [ ] Chrome (iOS) version XX
- [ ] Firefox version XX
- [ ] Safari (desktop) version XX
- [ ] Safari (iOS) version XX
- [ ] IE version XX
- [ ] Edge version XX
 
For Tooling issues:
- Node version: XX  
- Platform:  

Others:

</code></pre>

beagleknight:
I am trying to implement this myself here: https://github.com/codegram/store/tree/feat/state-selector-generic

I wrote the necessary code and an additional test but I am unable to make the test pass 😞 

```typescript
[...]

PR #894, opened 2019-03-04
fix(router-plugin): add state selector with generic

beagleknight:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: #893 

## What is the new behavior?

Now is possible to select the router state correctly typed when using the `state` selector like this:

```
store
  .select(state => RouterState.state<RouterStateParams>(state.router))
  .subscribe(routerState => {
    expect(routerState!.url).toEqual('/a-path?foo=bar');
    expect(routerState!.queryParams.foo).toEqual('bar');
  });
```

```
[ ] Yes
[x] No
```

### DIFF
--- a/packages/router-plugin/tests/router.plugin.spec.ts
+++ b/packages/router-plugin/tests/router.plugin.spec.ts
@@ -1,10 +1,10 @@
 import { Component, Provider } from '@angular/core';
 import { async, fakeAsync, TestBed, tick } from '@angular/core/testing';
-import { Router } from '@angular/router';
+import { Router, Params, RouterStateSnapshot } from '@angular/router';
 import { RouterTestingModule } from '@angular/router/testing';
 
 import { NgxsModule, Store } from '@ngxs/store';
-import { NgxsRouterPluginModule, RouterState } from '../';
+import { NgxsRouterPluginModule, RouterState, RouterStateSerializer } from '../';
 import { Navigate } from '@ngxs/router-plugin/src/router.actions';
 
 describe('NgxsRouterPlugin', () => {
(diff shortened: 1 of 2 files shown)


packages/router-plugin/tests/router.plugin.spec.ts
```diff
@@ -74,6 +74,39 @@ describe('NgxsRouterPlugin', () => {
       expect(routerState!.url).toEqual('/a-path');
     });
   }));
+
+  it('should select custom router state ', fakeAsync(async () => {
```
eranshmil: Is the `async` really required?

eranshmil: Is the `async` really required?

beagleknight: @eranshmil Regarding the [official documentation](https://ngxs.gitbook.io/ngxs/recipes/unit-testing) you need to wrap the test using Angular's `async` method. I tried without the `async` and it also works so maybe the `fakeAsync` method is also creating an asynchronous test zone? 🤔

eranshmil: Yeah, I wrote that in the documentation. ;)
Since then, I've figured out that it's not mandatory and not all the time you need to wrap it with angular's async().

And aside from that, you used the async-await modifier, and not the angular's utility.

beagleknight: haha that's embarassing 😅! I removed the `async` keyword since it's not necessary. I reused it from another test and didn't notice it was the modifier, my bad!

eranshmil: Thanks, I'm trying to figure out this mystery for some time, when we need to use it.

Issue #481, opened 2018-07-05
Resolved data not in store


piernik:
I'm using current version of `ngxs` with router plugin.

I've created my own router serializer:

	import {RouterStateSerializer} from '@ngxs/router-plugin/src/serializer';
	import {ActivatedRouteSnapshot, Params, RouterStateSnapshot} from '@angular/router';

	export interface RouterStateParsed {
	  url: string;
	  queryParams: Params;
	  params: Params;
	  data: any;
	}

	export class AppRouterSerializer implements RouterStateSerializer<RouterStateParsed> {
	  serialize(routerState: RouterStateSnapshot): RouterStateParsed {
		const {url} = routerState;
		const {queryParams} = routerState.root;

		let state: ActivatedRouteSnapshot = routerState.root;
		while (state.firstChild) {
		  state = state.firstChild;
		}
		const {params, data} = state;

		return {url, queryParams, params, data};
	  }
	}

I think that `serialize` is triggered before router `resolver` is resolved and that's why I don't have resolved data in store.

Is it a bug?

amcdnl:
@leon - Any thoughts? You wrote this one.

piernik:
@leon any news on that one?

hallowatcher:
Any news? This is a pretty vital feature.

eranshmil:
> Any news? This is a pretty vital feature.

Could you create a repro stackblitz using the following template?
https://stackblitz.com/fork/ngxs-simple

hallowatcher:
> > Any news? This is a pretty vital feature.
> 
> Could you create a repro stackblitz using the following template?
> https://stackblitz.com/fork/ngxs-simple

sure thing
https://stackblitz.com/edit/ngxs-router-resolver

markwhitfeld:
@piernik and @hallowatcher do you think that the current behavior would be desirable in any circumstance? We have a fix ready in PR #895 but just want to make sure that we don't need to provide a way of falling back to the current behavior.

hallowatcher:
I don't think the current behavior would be desirable in any case, as one expects their resolved data to show up in the state. That's the case for me, at least.

It would be opening doors rather than closing them

piernik:
I agreed with @hallowatcher In store should be up-to-date state.

markwhitfeld:
The fix has been released in v3.4.3!
Please test and confirm that it is working as expected.

hallowatcher:
Yes! It's working as expected.

[...]

PR #895, opened 2019-03-05
fix(router-plugin): serialize after `ResolveEnd`

arturovt:
## PR Checklist

## PR Type
```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: #481 

## What is the new behavior?
Right now `serialize` is triggered in the `beforePreactivation` hook. Before preactivation is handled before running guards and resolvers. I've moved it to the `ResolveEnd` hook

```
[ ] Yes
[x] No
```

### DIFF
--- a/integration/app/list/list.component.html
+++ b/integration/app/list/list.component.html
@@ -2,3 +2,7 @@
   {{ list$ | async }},
   {{ foo | async }}
 </div>
+
+<div *ngIf="router$ | async as router">
+  animals were resolved {{ router.state.root.data.list }}
+</div>
--- a/integration/app/list/list.component.html
+++ b/integration/app/list/list.component.html
@@ -4,5 +4,5 @@
 </div>
 
 <div *ngIf="router$ | async as router">
-  animals were resolved {{ router.root.firstChild.data.list }}
+  animals were resolved {{ router.root.firstChild.firstChild.data.list }}
 </div>
--- a/integration/app/list/list.component.html
+++ b/integration/app/list/list.component.html
@@ -4,5 +4,5 @@
 </div>
 
 <div *ngIf="router$ | async as router">
-  animals were resolved {{ router.state.root.data.list }}
+  animals were resolved {{ router.root.firstChild.data.list }}
 </div>
--- a/integration/app/list/list.component.ts
+++ b/integration/app/list/list.component.ts
@@ -1,5 +1,6 @@
 import { Component } from '@angular/core';
 import { Select } from '@ngxs/store';
+import { RouterState } from '@ngxs/router-plugin';
 import { Observable } from 'rxjs';
 
 import { ListState } from '@integration/list/list.state';
@@ -11,4 +12,5 @@ import { ListState } from '@integration/list/list.state';
 export class ListComponent {
   @Select(ListState) public list$: Observable<string[]>;
   @Select(ListState.hello) public foo: Observable<string>;
+  @Select(RouterState) public router$: Observable<RouterState>;
 }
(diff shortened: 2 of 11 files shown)


arturovt:
@splincode @markwhitfeld @eranshmil 
The current implementation uses `RouterStateSnapshot`. After investigation I've found that it's basically impossible to access resolved data, because the `data` property can be inside 4 nested `firstChild` objects.

eranshmil:
This is a breaking change. :(

arturovt:
@eranshmil okay, I'm gonna try to minimize changes

arturovt:
@eranshmil @splincode ping

also guys what do you think about adding a new selector to the `RouterState` that will resolve the `ActivatedRouteSnapshot` linked to provided component? Basic example:

```TS
// pseudocode
static findActivatedRouteThatLinkedWithComponent<T>(component: Type<T>, snapshot: ActivatedRouteSnapshot) {
  let route = snapshot;

  while (route && route.component.constructor !== component) {
    route = route.firstChild;
  }

  return route;
}

static activatedRoute<T>(component: Type<T>) {
  return createSelector([RouterState], (state: RouterStateModel) => {
    return this.findActivatedRouteThatLinkedWithComponent(component, state.state.root);
  });
}
```

So people can use it like:

```
export class TodosComponent {
  @Select(RouterState.activatedRoute(TodosComponent)) public route$;
}
```

arturovt:
@eranshmil @splincode @markwhitfeld

splincode:
@eranshmil why is it a breaking changes?

markwhitfeld:
@splincode why did you delete @eranshmil 's comment?

splincode:
@eranshmil in the chat he replied that this is not a breaking change

markwhitfeld:
Ok, rather add a new comment to the thread than delete.
Delete is really only reserved for if there is some profane or abusive comment.

arturovt:
@splincode I had a breaking change in the first commit

splincode:
@markwhitfeld sorry :(

markwhitfeld:
@splincode no worries 👍 
@arturovt I saw that you fixed it after Eran's comment
Could I get your feedback on this
> @arturovt Just one question, it does make sense to me that the snapshot should happen on `ResolveEnd`, but do you think that there could have been any reason for someone to expect it to happen in `beforePreactivation` instead?
> Just checking if this could be a breaking change for a reasonable use case...

arturovt:
@markwhitfeld umm, I'm not sure if it's a breaking change... I guess I could add more tests there but it's hard to test route snapshot as they dont work with `RouterTestingModule`, so I was able to add only SSR tests :fearful:

markwhitfeld:
Yeah, not worried about the tests, those look fine.
Was just wondering if there could be some wierd case where someone would actually want the previous behaviour.

arturovt:
@markwhitfeld I've heard from @eranshmil that he actively used this plugin on some project(s). Eran what do you think?

markwhitfeld:
I was just thinking, is there any difference between if this was dispatched on `NavigationEnd` as opposed to `ResolveEnd`? Not particularly wanting to change, but just interested in what more the later event could offer...

arturovt:
`NavigationEnd` instance has only string information about the last navigation - `id` and `urlAfterRedirects`

eranshmil:
> Looks good to me.
> @arturovt Just one question, it does make sense to me that the snapshot should happen on `ResolveEnd`, but do you think that there could have been any reason for someone to expect it to happen in `beforePreactivation` instead?
> Just checking if this could be a breaking change for a reasonable use case...

Sorry for the late answer.
I don't remember what I wrote in here, but this is what **might** be a breaking change. In ngrx, they added a new option to the config.
But, just like you, I also think it should be that way. 😆

Issue #902, opened 2019-03-07
The state created with undefined initial state even when defaults are specified


catchshyam:
## I'm submitting a...

<pre><code>
[x ] Regression (a behavior that used to work and stopped working in a new release)
[ x] Bug report  
[ ] Performance issue
[ ] Feature request
[ ] Documentation issue or request
[ ] Support request => https://github.com/ngxs/store/blob/master/CONTRIBUTING.md
[ ] Other... Please describe:
</code></pre>

## Current behavior
The feature state object is created with initial state as undefined (even when the defaults is provided). 

## Expected behavior
The feature state object must use the values given in defaults object definition as initial state.

## Minimal reproduction of the problem with instructions
https://stackblitz.com/edit/ngxs-no-initial-state

Launch the application and click on the hyperlink in home page. Then on the console, you will see the debug message with initial state object.

## Environment

<pre><code>
Libs:
- @angular/core version: 7.2.5
- @ngxs/store version: 3.4.1
- @ngxs-labs/emitter: 1.9.0
- @ngxs-labs/immer-adapter: 1.1.2

splincode:
@markwhitfeld very strange

![image](https://user-images.githubusercontent.com/12021443/53945065-cf31ca00-40d1-11e9-916c-332b3e6a6187.png)

![image](https://user-images.githubusercontent.com/12021443/53945078-d527ab00-40d1-11e9-9e9a-86fdd3661060.png)

splincode:
@arturovt problem with emitter plugin?

kuhnroyal:
I can confirm this happening with 3.4.0/3.4.1

markwhitfeld:
This only affects Emitter plugin users. This is because of the way that Emitter inherits from Store to create an `EmitStore`.
I have found the point at which the state is reset. See this stackblitz:
https://stackblitz.com/edit/ngxs-no-initial-state-debug

This bug is caused by this change in PR #791 
https://github.com/ngxs/store/pull/791/files#diff-084c7b85b6562e3126b58451d5f8943aR20

Ngxs does not expect anyone to inherit from `Store` and essentially run this code twice. Emitter does some things with Ngxs internals that are not usual plugin behaviours.

splincode:
fixed 3.4.2

xmlking:
when we have `default` state and partial state coming from ` NgxsStoragePluginModule`
default state is still coming as undefined

Store
[...]

PR #904, opened 2019-03-07
fix(store): initial state should not be overwritten by defaults

markwhitfeld:
## PR Checklist

## PR Type

```
[X] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: #902
There was a regression caused by PR #791 in this line: 
https://github.com/ngxs/store/pull/791/files#diff-084c7b85b6562e3126b58451d5f8943aR20

This only affects Emitter plugin users. This is because of the way that Emitter inherits from Store to create an EmitStore. I found the point at which the state was being overwritten. See this stackblitz:
https://stackblitz.com/edit/ngxs-no-initial-state-debug

Ngxs does not expect a plugin to inherit from Store and essentially run this code twice. Emitter does some things with Ngxs internals that are not usual plugin behaviours.

## What is the new behavior?

This code is now protected from being run twice. The default values are only written into the store if there is no current value in the store.

```
[ ] Yes
[X] No
```

### DIFF
--- a/packages/store/src/store.ts
+++ b/packages/store/src/store.ts
@@ -18,7 +18,11 @@ export class Store {
     private _config: NgxsConfig,
     private _internalExecutionStrategy: InternalNgxsExecutionStrategy
   ) {
-    this._stateStream.next(this._config.defaultsState);
+    const value = this._stateStream.value;
+    if (!value || Object.keys(value).length === 0) {
+      // Only set the defaults if the state stream is empty.
+      this._stateStream.next(this._config.defaultsState);
+    }
   }
 
   /**
--- a/packages/store/src/store.ts
+++ b/packages/store/src/store.ts
@@ -9,6 +9,7 @@ import { StateStream } from './internal/state-stream';
 import { NgxsConfig } from './symbols';
 import { InternalNgxsExecutionStrategy } from './execution/internal-ngxs-execution-strategy';
 import { leaveNgxs } from './operators/leave-ngxs';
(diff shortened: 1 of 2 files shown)


packages/store/src/store.ts
```diff
@@ -18,7 +18,11 @@ export class Store {
     private _config: NgxsConfig,
     private _internalExecutionStrategy: InternalNgxsExecutionStrategy
   ) {
-    this._stateStream.next(this._config.defaultsState);
+    const value = this._stateStream.value;
```
splincode:
```ts
const value: ObjectKeyMap<any> = this._stateStream.value;
const storeIsEmpty: boolean = !value || Object.keys(value).length === 0;
if (storeIsEmpty) {
 this._stateStream.next(this._config.defaultsState);
}
```

packages/store/tests/default-states.spec.ts
```diff
@@ -41,6 +44,46 @@ describe('Reusable States', () => {
...
+        MyStateSettingModule,
+        NgxsModule.forRoot([FooState], {
+          defaultsState: {
+            foo: [4, 5, 6]
+          }
+        })
+      ]
+    });
+
+    // const stateStream: StateStream = TestBed.get(StateStream);
```
splincode: remove comments

markwhitfeld:
I have tested HMR in the integration app and can confirm that it is working great.

Issue #922, opened 2019-03-18
queryParams changes doesn't trigger RouterNavigation action


goldenass:
##I'm submitting a...

<pre><code>
[x ] Regression
[ ] Bug report  
[ ] Performance issue
[ ] Feature request
[ ] Documentation issue or request
[ ] Support request 
[ ] Other... Please describe:
</code></pre>

## Current behavior

Navigating to the same path changing only the queryParams doesn't trigger a RouteNavigation action anymore.

this.store.dispatch(new Navigate('/search',{query:'a'})); //--> triggers RouterNavigation action
this.store.dispatch(new Navigate('/search',{query:'ab'})); //--> doesen't trigger RouterNavigation action

## Expected behavior

Changing anything in the route a RouterNavigation action must be triggered.

this.store.dispatch(new Navigate('/search',{query:'a'})); //--> triggers RouterNavigation action
this.store.dispatch(new Navigate('/search',{query:'ab'})); //--> also triggers RouterNavigation action

## What is the motivation / use case for changing the behavior?

that breaks code written for previous versions

## Environment

<pre><code>
Libs:
- @angular/core version: 7.2.9
- @ngxs/store version: 3.4.3

Browser:
- [ x] Chrome (desktop) version XX
- [ x] Chrome (Android) version XX
- [x ] Chrome (iOS) version XX
- [ x] Firefox version XX
- [ x] Safari (desktop) version XX
- [x ] Safari (iOS) version XX
- [ x] IE version XX
- [ x] Edge version XX

arturovt:
@goldenass oh, that's interesting! Would be great if you could create a repro on the stackblitz

goldenass:
@arturovt 

Here you are:
https://stackblitz.com/edit/angular-hacwyr

If you change the search value, routing happens, but the eventCount remains 1

mucbits:
Facing the same problem.
This was working on 3.4.2: [stackblitz](https://stackblitz.com/edit/ngxs-simple-iwdk8f)
Updating to 3.4.3 breaks it: [stackblitz](https://stackblitz.com/edit/ngxs-simple-s4h8qh)

arturovt:
@goldenass thank you, I will investigate!

arturovt:
@goldenass @mucbits 
Guys, this might be fixed by #924. Could you please check it and say if everything alright?
You will need to install the `@dev` version from npm to test.

mucbits:
Not sure how to install the new dev version of the router-plugin.
[...]

PR #939, opened 2019-03-24
fix(router-plugin): dispatch `RouterNavigation` right away after the necessary event

arturovt:
## PR Checklist

## PR Type
```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: #922 #946 

```
[ ] Yes
[x] No
```

### DIFF
--- a/packages/router-plugin/src/router.state.ts
+++ b/packages/router-plugin/src/router.state.ts
@@ -6,7 +6,6 @@ import {
   RouterStateSnapshot,
   RoutesRecognized,
   ResolveEnd,
-  NavigationEnd,
   GuardsCheckEnd
 } from '@angular/router';
 import { Location } from '@angular/common';
@@ -42,7 +41,6 @@ export class RouterState {
   private lastRoutesRecognized: RoutesRecognized;
   private dispatchTriggeredByRouter = false; // used only in dev mode in combination with routerReducer
   private navigationTriggeredByDispatch = false; // used only in dev mode in combination with routerReducer
-  private lastResolvedStateSnapshot: RouterStateSnapshot = null!;
 
   /**
    * Selectors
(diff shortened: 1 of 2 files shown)


eranshmil:
Does it mean that the same state will be triggered twice? (In case of `GuardsCheckEnd` + `ResolveEnd`)
And please fix the two failing tests, caused by the same error:
```ts
ERROR: /home/travis/build/ngxs/store/packages/router-plugin/tests/router.plugin.spec.ts:183:27 - Shadowed name: 'store'
```

arturovt:
This is a vicious circle which I'm stuck in.
The `ResolveEnd` event is triggered only in cases if some route has some linked resolvers. The `GuardsCheckEnd` event is triggered always.
If I dispatch `RouterNavigation` event after the `NavigationEnd` is triggered - than we cannot access `RouterState` inside selectors.

markwhitfeld:
@arturovt, not considering the required data, where do you think would be the ideal place to dispatch the action?

arturovt:
@markwhitfeld To prevent breaking changes for our users, I'd say the ideal place is in `NavigationEnd`. The problem is users want to access the `router` state on the initial load, and it's not possible when the `RouterNavigation` action is dispatched after `NavigationEnd`.
Paradoxically :joy: :joy: I'd like to hear your thoughts.

markwhitfeld:
After discussing, we decided to add a new `RouterDataResolved` Action that represents the event when the data has been resolved. The `RouterNavigation` Action will be dispatched on GuardCheckEnd because this event is always executed and is the last guaranteed event before the component's constructor is called. The `RouterDataResolved` Action will also update the state, but this will only happen if there are resolvers attached to the route.
This work has been moved to PR #939 and this PR has been closed without merging.

Issue #946, opened 2019-03-27
[Regression] No route snapshot provided by the Router Plugin in version 3.4.3


catchshyam:
## I'm submitting a...

<pre><code>
[x ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[ ] Performance issue
[ ] Feature request
[ ] Documentation issue or request
[ ] Support request => https://github.com/ngxs/store/blob/master/CONTRIBUTING.md
[ ] Other... Please describe:
</code></pre>

## Current behavior
I am using the Router Plugin and here is the selectorI have to get the active route,

>   @Selector([RouterState.state])
>     static getActiveRoute(stateModel: AppStateModel, route: RouterStateSnapshot): ActivatedRouteSnapshot {
>         console.log('Snapshot provided by the router state  ', route);
>         let state: ActivatedRouteSnapshot = route.root;
>         while (state.firstChild) {
>             state = state.firstChild;
>         }
>         return state;
>     }
> 

Until 3.4.1, I would get the route object in the console statement above. In 3.4.3, the route object injected by RouterState is undefined.

## Expected behavior
The RouterState must provide the actual snapshot.

## Minimal reproduction of the problem with instructions

https://stackblitz.com/edit/ngxs-router-plugin-bug
 - Launch the application and click on the hyperlink "Click me and See the console". 
 - See the debug message in the console "Snapshot provided by the router state
undefined"
 - If you change the NGXS and router plugin version to 3.4.1 in package.json, application outputs the correct object in the debug message.

## Environment

<pre><code>
Libs:
- @angular/core version: ^7.2.5
- @ngxs/store version: 3.4.3
- @ngxs/router-plugin version: 3.4.3

Browser:
- [ x] Chrome (desktop) version XX
- [ ] Chrome (Android) version XX
- [ ] Chrome (iOS) version XX
- [ ] Firefox version XX
- [ ] Safari (desktop) version XX
- [ ] Safari (iOS) version XX
- [ ] IE version XX
- [ ] Edge version XX

</code></pre>

arturovt:
@catchshyam 
Hello. Yes, we're working on that.

markwhitfeld:
This should be fixed in the latest `@dev` version.
Please could you test it out and let us know if it fixes your problem?

arturovt:
@catchshyam 
We would really appreciate your regression feedback

splincode:
Released in 3.5.0

PR #952, opened 2019-04-01
feat(router-plugin): implement `RouterDataResolved` action

arturovt:
## PR Checklist

## PR Type
```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: #946 #922 #1049 

```
[ ] Yes
[x] No
```

### DIFF
--- a/packages/router-plugin/src/router.state.ts
+++ b/packages/router-plugin/src/router.state.ts
@@ -118,7 +118,7 @@ export class RouterState {
 
   private guardsCheckEnd(routerState: RouterStateSnapshot): void {
     if (this.routerStateSnapshot) {
-      routerState = { ...routerState, ...this.routerStateSnapshot };
+      routerState = { ...this.routerStateSnapshot, ...routerState };
     }
 
     this.routerStateSnapshot = this._serializer.serialize(routerState);
--- a/packages/router-plugin/src/router.state.ts
+++ b/packages/router-plugin/src/router.state.ts
@@ -6,7 +6,6 @@ import {
   RouterStateSnapshot,
   RoutesRecognized,
   ResolveEnd,
-  NavigationEnd,
   GuardsCheckEnd
 } from '@angular/router';
 import { Location } from '@angular/common';
@@ -19,7 +18,8 @@ import {
   RouterAction,
   RouterCancel,
   RouterError,
-  RouterNavigation
+  RouterNavigation,
+  RouterDataResolved
 } from './router.actions';
 import { RouterStateSerializer } from './serializer';
 
@@ -42,7 +42,6 @@ export class RouterState {
   private lastRoutesRecognized: RoutesRecognized;
   private dispatchTriggeredByRouter = false; // used only in dev mode in combination with routerReducer
   private navigationTriggeredByDispatch = false; // used only in dev mode in combination with routerReducer
-  private lastResolvedStateSnapshot: RouterStateSnapshot = null!;
 
   /**
    * Selectors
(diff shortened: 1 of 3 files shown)


markwhitfeld:
Looks good so far. Just a failing unit test to fix.

splincode:
@arturovt any progress?

arturovt:
Curious that circle tests pass, but travis not :joy:

packages/router-plugin/src/router.state.ts
```diff
@@ -118,8 +116,13 @@ export class RouterState {
     });
   }
 
-  private navigationEnd(): void {
-    this.routerStateSnapshot = this._serializer.serialize(this.lastResolvedStateSnapshot);
+  private guardsCheckEnd(routerState: RouterStateSnapshot): void {
+    if (this.routerStateSnapshot) {
+      routerState = { ...this.routerStateSnapshot, ...routerState };
```
markwhitfeld: A problem I forsee here is that this code is relying on the fact that the `this._serializer.serialize(routerState)` returns data that is compatible with the `routerState`.
Maybe we should create a separate snapshot of the data (and any other specific properties that we are trying to resolve this for) directly off the `routerState` and then do this same technique here with that data snapshot.

packages/router-plugin/tests/router.plugin.spec.ts
```diff
@@ -133,45 +155,348 @@ describe('NgxsRouterPlugin', () => {
...
+          take(1)
+        )
+        .subscribe(counter => {
+          // Assert
+          expect(selectorCalledTimes).toEqual(3);
+          expect(selectorCalledTimes).toEqual(counter);
+        });
+
+      await router.navigateByUrl('/a/b/c');
+      await router.navigateByUrl('/a/b');
```
markwhitfeld: Probably better to have:
```
      const subscription = store
        .select(CounterState.counter)
        .subscribe(counter => { });

      await router.navigateByUrl('/a/b/c');
      await router.navigateByUrl('/a/b');
      subscription.unsubscribe();

      // Assert
      expect(selectorCalledTimes).toEqual(3);
      expect(selectorCalledTimes).toEqual(counter);
```
Thus the expects would be guaranteed to be called.
In your code it may silently fail if the selectors are called less than three times and would pass if called more than three times.

packages/router-plugin/tests/router.plugin.spec.ts
```diff
@@ -133,45 +155,348 @@ describe('NgxsRouterPlugin', () => {
...
+      // The very first `ResolveEnd` event is triggered during root module bootstrapping
+      actions$
+        .pipe(
+          ofActionSuccessful(RouterDataResolved),
+          first()
+        )
+        .subscribe(({ routerState }: RouterDataResolved) => {
+          // Assert
+          const dataFromTheEvent = routerState.root.firstChild!.data;
+          expect(dataFromTheEvent).toEqual({ test });
```
markwhitfeld: If we just capture the data in the subscribe into a variable then we can move this `expect` out of the subscribe and into the main body of the test. This makes for a test that can detect failure more reliably.

packages/router-plugin/tests/router.plugin.spec.ts
```diff
@@ -103,18 +126,17 @@ describe('NgxsRouterPlugin', () => {
 
     const store: Store = TestBed.get(Store);
 
-    store.dispatch(new Navigate(['a-path'], { foo: 'bar' }));
-    tick();
+    await store.dispatch(new Navigate(['a-path'], { foo: 'bar' })).toPromise();
```
splincode: why await instead tick?

markwhitfeld: I think code that relies on js behavior is better than code that relies in zone.js
This makes the test more reliable IMO

splincode: Well, then it will be necessary to rewrite a lot of code anyway from zone.js?)

markwhitfeld:
@arturovt I couldn't push to your repo, so I pushed to the main repo so that we can collaborate on the PR.
This PR will now be tracked in PR #1059 
Closing this one

PR #954, opened 2019-04-02
feat: add selector options to metadata

markwhitfeld:
## PR Checklist

## PR Type

```
[ ] Bugfix
[X] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

There is currently an issue where the state of the container class of a selector is always injected as the first parameter. This is a problem because it makes the memoisation of every selector defined in a state class dependent on the entire state class, and therefore they are recalculated on any change within the state.
(I will populate the list of realted issues later) 
Issue Number: N/A

## What is the new behavior?
This PR introduces some internal representations of selector options, configurable at a method, class, or application level. This will be used as part of the `compat` mechanisms to ease the change of the above behaviour in NGXS v4.

```
[ ] Yes
[X] No
```

## Other information
This PR will be followed by another PR that creates a `compat` mechanism for configuring selector behaviour.

### DIFF
--- a/packages/store/src/internal/internals.ts
+++ b/packages/store/src/internal/internals.ts
@@ -40,7 +40,7 @@ export interface MetaDataModel {
   selectFromAppState: SelectFromState | null;
   children?: StateClass[];
   instance: any;
-  sharedSelectorOptions?: InternalSelectorOptions;
+  internalSelectorOptions?: InternalSelectorOptions;
 }
 
 export type SelectFromState = (state: any) => any;
@@ -114,7 +114,7 @@ export function ensureSelectorMetadata(target: Function): SelectorMetaDataModel 
       containerClass: null,
       selectorName: null,
       selectorOptions: {
-        injectContainerState: true // default is true in v3, will change in v4
+        injectContainerState: true // TODO: default is true in v3, will change in v4
       }
     };
 
--- a/packages/store/src/internal/internals.ts
+++ b/packages/store/src/internal/internals.ts
@@ -40,15 +40,21 @@ export interface MetaDataModel {
   selectFromAppState: SelectFromState | null;
   children?: StateClass[];
   instance: any;
+  sharedSelectorOptions?: InternalSelectorOptions;
 }
 
 export type SelectFromState = (state: any) => any;
+
+export interface InternalSelectorOptions {
+  injectContainerState?: boolean;
+}
 
 export interface SelectorMetaDataModel {
   selectFromAppState: SelectFromState | null;
   originalFn: Function | null;
   containerClass: any;
   selectorName: string | null;
+  selectorOptions: InternalSelectorOptions;
 }
 
 export interface MappedStore {
@@ -106,7 +112,10 @@ export function ensureSelectorMetadata(target: Function): SelectorMetaDataModel 
       selectFromAppState: null,
       originalFn: null,
       containerClass: null,
-      selectorName: null
+      selectorName: null,
+      selectorOptions: {
+        injectContainerState: true // default is true in v3, will change in v4
+      }
     };
 
     Object.defineProperty(target, SELECTOR_META_KEY, { value: defaultMetadata });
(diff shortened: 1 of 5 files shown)


markwhitfeld:
You are welcome to review.
I still need to add tests and also add a mechanism to specify application level overrides.

markwhitfeld:
Although this PR is labelled as a feature it does not modify any public interface or any behavior.
This could be merged now and be part of the next bugfix release. This would allow for experimentation with what an API could look like for the `compat` subpackage to enable v4 style selectors.

markwhitfeld:
@deebloo @eranshmil @arturovt Any other takers for a review? I would like at least 2 reviews.
@splincode Please have another look because I have added a mechanism for application level configuration.

packages/store/src/internal/internals.ts
```diff
@@ -40,15 +40,21 @@ export interface MetaDataModel {
   selectFromAppState: SelectFromState | null;
   children?: StateClass[];
   instance: any;
+  sharedSelectorOptions?: InternalSelectorOptions;
 }
 
 export type SelectFromState = (state: any) => any;
 
+export interface InternalSelectorOptions {
```
splincode: why not give a different name?

`SharedSelectorOptions`

packages/store/src/internal/internals.ts
```diff
@@ -106,7 +112,10 @@ export function ensureSelectorMetadata(target: Function): SelectorMetaDataModel
       selectFromAppState: null,
       originalFn: null,
       containerClass: null,
-      selectorName: null
+      selectorName: null,
+      selectorOptions: {
+        injectContainerState: true // default is true in v3, will change in v4
```
splincode: add prefix
// TODO:

splincode:
I want you to expand support for State decorator and add test

```
@NgModule({
 imports: [
    NgxsModule.forRoot(states, {
      developmentMode: true
      sharedSelectorOptions: { injectContainerState: false } // default false everywhere
    })
 ]
})
export class AppModule {}
```

my-state.state.ts

```
@State({
  name: 'myState',
  default: {},
  selectorOptions: {
    injectContainerState: true // default true for all selectors in current state
  }
})
export class MyState {
  @Selector() public mySelector = (state) => state;
}
```

packages/store/src/internal/internals.ts
markwhitfeld: Will do

markwhitfeld:
@splincode The purpose of this PR is to make the setting available internally (no change to public API).
Then, in another PR we will be adding the API for manipulating this setting.
I don't want to expose this as an option on the normal public api because it is only available for the purpose of transitioning from v3 to v4. Through both an early opt in upgrade before v4 arrives, or an opt in downgrade to v3 style after upgrading to v4.

cc @sinedied - this would form the basis for an API that we were discussing in PR #858

packages/store/src/internal/internals.ts
markwhitfeld: I have kept the name `InternalSelectorOptions` to keep things consistent. They should not be part of the public API and this needs to be explicit.

splincode:
@markwhitfeld 
I think we should make this a public API. Because in the future, we will want to come up with a new behavior or something else, or otherwise expand the selectors.

markwhitfeld:
When we need the public API we will create it. As mentioned this PR is to create the internal mechanisms and not to introduce a public API.

packages/store/src/internal/state-factory.ts
```diff
@@ -231,6 +232,12 @@ export class StateFactory {
   private addRuntimeInfoToMeta(meta: MetaDataModel, depth: string): void {
     meta.path = depth;
     meta.selectFromAppState = propGetter(depth.split('.'), this._config);
+    const globalSelectorOptions: InternalSelectorOptions = (<any>this._config)
```
eranshmil: Why do you need to cast to `any`?

markwhitfeld: Because this is a hidden internal field and not part of the public interface

PR #956, opened 2019-04-04
feat(store): add state defaults to UpdateState

eranshmil:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

The comeback of PR #845, with one change - `payload` parameter has been changed to `addedStats`.
Issue Number: #838 

## What is the new behavior?
State defaults array is being added to the `UpdateState` action, so they will be visible using the `logger` and `devtools` plugins.
The `payload` parameter is optional in order to prevent a breaking change, in case some developer used this action directly. We might consider changing that to be required for version 4. #827 

image
image

```
[ ] Yes
[x] No
```

### DIFF
--- a/packages/store/src/actions/actions.ts
+++ b/packages/store/src/actions/actions.ts
@@ -1,3 +1,5 @@
+import { ObjectKeyMap } from '../internal/internals';
+
 /**
  * Init action
  */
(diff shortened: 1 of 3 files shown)


eranshmil:
> Just a thought, do you think that the `addedStates` property should be directly on the action or under a `payload` property?
> I personally prefer putting properties directly on action but most of our docs have them under a `payload` property, so maybe we should do that for consistency?
> I see that most of the plugins follow this convention except for the router plugin Actions.
> I feel like the `payload` property is just another layer to get through, so that is why I opt to put the properties straight on the action.
> What do you think?

Yeah, you're right.
We need to change all of the examples in the future. ;)

Issue #843, opened 2019-02-15
Errors are swallowed in selectors


ogix:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[x] Bug report  
[ ] Performance issue
[ ] Feature request
[ ] Documentation issue or request
[ ] Support request => https://github.com/ngxs/store/blob/master/CONTRIBUTING.md
[ ] Other... Please describe:
</code></pre>

## Current behavior

When error occurs in a selector - it is swallowed and not shown in the console.

## Expected behavior

Error should be shown in the console.

## Minimal reproduction of the problem with instructions

```
  @Selector() 
  static pandas(state: string[]) {
    const foo: any = null;
    console.log(foo.bar);
    return state.filter(s => s.indexOf('panda') > -1);
  }
```

## What is the motivation / use case for changing the behavior?

Finding bugs easier in selectors.

## Environment

<pre><code>
Libs:
- @angular/core version: 7.1.4
- @ngxs/store version: 3.3.3

Browser:
- [ ] Chrome (desktop) version XX
- [ ] Chrome (Android) version XX
- [ ] Chrome (iOS) version XX
- [ ] Firefox version XX
- [ ] Safari (desktop) version XX
- [ ] Safari (iOS) version XX
- [ ] IE version XX
- [ ] Edge version XX
 
For Tooling issues:
- Node version: XX  
- Platform:  

Others:

</code></pre>

arturovt:
@ogix selectors calls are wrapped in `try-catch`, this decision was made in order to avoid exceptions. E.g. you've got some state that stores a `null` value or some object, and you want to create a selector that returns `token` value

```TS
@State<Token | null>({
  name: 'token',
  defaults: null
})
class TokenState {
  @Selector()
  public static getAccessToken(state: Token) {
    return state.accessToken;
  }
}
```
this won't throw and will return `undefined`.

@splincode what do you think?

eranshmil:
@arturovt A `console.log` could be added to the `catch`, if `developmentMode` is turned on.

dominicglenn:
`console.error` perhaps instead when dev mode?

ogix:
Isn't it possible to check if state is null before applying a selector function? And then replace try/catch with that check.

sroettering:
The swallowing is pretty annoying for testing, too. Most of the time I have to guess which properties I forgot to add to the initial state setup.

I think at least logging in dev mode should be added.

[...]

PR #1015, opened 2019-04-21
fix(store): disable suppressing errors in selector

splincode:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

https://github.com/ngxs/store/issues/843

## What is the new behavior?

```ts
@NgModule({
 imports: [
   NgxsModule.forRoot([TasksState], {
       developmentMode: true,
       selectorOptions: { suppressErrors: false }
    })
 ]
})
```

```
[ ] Yes
[x] No
```

### DIFF
--- a/packages/store/src/internal/internals.ts
+++ b/packages/store/src/internal/internals.ts
@@ -40,13 +40,14 @@ export interface MetaDataModel {
   selectFromAppState: SelectFromState | null;
   children?: StateClass[];
   instance: any;
-  internalSelectorOptions?: InternalSelectorOptions;
+  selectorOptions?: SharedSelectorOptions;
 }
 
 export type SelectFromState = (state: any) => any;
 
-export interface InternalSelectorOptions {
+export interface SharedSelectorOptions {
   injectContainerState?: boolean;
+  suppressErrors?: boolean;
 }
 
 export interface SelectorMetaDataModel {
@@ -54,7 +55,7 @@ export interface SelectorMetaDataModel {
   originalFn: Function | null;
   containerClass: any;
   selectorName: string | null;
-  selectorOptions: InternalSelectorOptions;
+  selectorOptions: SharedSelectorOptions;
 }
 
 export interface MappedStore {
--- a/packages/store/src/internal/state-factory.ts
+++ b/packages/store/src/internal/state-factory.ts
@@ -25,7 +25,7 @@ import {
   StatesAndDefaults,
   StatesByName,
   topologicalSort,
-  InternalSelectorOptions
+  SharedSelectorOptions
 } from './internals';
 import { getActionTypeFromInstance, getValue, setValue } from '../utils/utils';
 import { ofActionDispatched } from '../operators/of-action';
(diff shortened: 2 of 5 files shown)


splincode:
@markwhitfeld 
What are the advantages of this interface `internalSelectorOptions`? We still need to use this opportunity anyway.

@NgModule({
 imports: [
   NgxsModule.forRoot([TasksState], {
       developmentMode: true,
       selectorOptions: { injectContainerState: false }
    })
 ]
})

markwhitfeld:
The `injectContainerState` setting will not be used as a permanent feature, but rather used internally as a transitionary step in upgrading to v4, so I don't want to add it to the public interface.
It will be used internally by a `compat` decorator/plugin. 
For example a decorator that could be applied to a selector or class. For Example `@UpgradeSelectorToV4()`. This would be provided as part of a `compat` subpackage. I will be doing a PR for this soon.

splincode:
@markwhitfeld

Do not you think that this way you complicate the transition?  I think we need the whole team to discuss this moment!  because the decorator will be much harder to support and why people will need to have extra dependencies in their code.

To upgrade to version 4, people can be asked to do so.

```ts
@NgModule({
 imports: [
   NgxsModule.forRoot([TasksState], {
       developmentMode: true,
       selectorOptions: { injectContainerState: false }
    })
 ]
})

```

If nothing works for them, it will be able for the 4th version to leave this injectContainerState flag to true as global settings.

markwhitfeld:
I have always thought of the transition using the decorators which come from the `compat` library so that we can drop then after the transition, but keeping this as an option is growing on me...
What would you propose the syntax to be when specifying it at a state class level or for an individual selector?

splincode:
@markwhitfeld 

ngxs.config.ts

```ts
export const NGXS_CONFIG = {
     developmentMode: true,
     selectorOptions: { 
           // use global options
           injectContainerState: false
     }
};
```

app.module.ts

```ts
@NgModule({
 imports: [
   NgxsModule.forRoot([TasksState], NGXS_CONFIG)
 ]
})
export class AppModule {}
```

app.state.ts

```ts
@State({
 name: 'appState',
 default: {
   ... 
 },
 selectorOptions: { 
     // use options for local any selector in state
     injectContainerState: true
 }
})
export class AppState {
   
   @Selector() 
   public static selectorA(state) {
     return state;
   }

   @Selector() 
   public static selectorB(state) {
     return state;
   }
 
    // use custom selector options for current selector
   @Selector([], { injectContainerState: false }) 
   public static selectorC(state) {
     return state;
   }

}
```

markwhitfeld:
@sinedied what do you think of this proposal?

markwhitfeld:
@splincode I think that we would need to rather have a `@SelectorOptions(...)` decorator to apply at the class level because you can define selectors outside of states.

sinedied:
My opinion on this is that a global setting for the whole app would be enough, a configuration per state seems overkill to me.
The simple reason is that instead of an option, it's easier to have a distinct selector decorator with a different name for the transition time (that could be bundled in a compat package) if you need to keed the old behavior at specific places.

We already went this way even before the v4 option is available, by making our own `OptimizedSelector` decorator that behaves just like v4 will. Once v4 is release, a simple search/replace on our selector to put back the one from NGXS will be enough to make the migration 🎉 

For people wanting to delay the migration, the easiest way for them to migrate would be one of this in my opinion:
- either set compatibility option to v3 globally and then selectively upgrade the selectors using a differently named one and move on once the full migration is done (like we plan to)
- set compatibility option to v4, and use a differently named selector to downgrade behavior to v3 (opposite of previous solution), can be a global search/replace at the beginning then progressing in the migration you replace the compat selector with the normal one. At the end, nothing left to do, just to remove the compat package

I think compatibility selectors to either upgrade or downgrade with a distinct name is simpler and better than an option, as it's easier to track in the app during the migration (and easier to implement for you).

Issue #917, opened 2019-03-14
Life-cycle events not triggered in root state


Koslun:
## I'm submitting a...

<pre><code>
[ x ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[ ] Performance issue
[ ] Feature request
[ ] Documentation issue or request
[ ] Support request => https://github.com/ngxs/store/blob/master/CONTRIBUTING.md
[ ] Other... Please describe:
</code></pre>

## Current behavior

1. Implement `NgxsOnInit` and `NgxsAfterBootstrap` in a state that is imported with root. In our case the state was `AuthState`:

```
 NgxsModule.forRoot(
      [
        AuthState,
        // follwed by three other states
      ],
      {
        developmentMode: !environment.production,
      }
    ),
// importing other plugins
// importing module with a NgxsModule.forFeature
```
2. Load the app

3. **Neither** life-cycle events are triggered.

## Expected behavior

Expecting life-cycle events to be triggered when loading the app at the specified timings described in the [documentation](https://ngxs.gitbook.io/ngxs/advanced/life-cycle).

This is still working in versions prior to `3.4.0` such as `3.3.4`. 

## Minimal reproduction of the problem with instructions
Was unable to reproduce in a minimal reproduction. So assuming it must be some edge-case to do with import order of NGXS modules similar to other issues: #539 and #375. The difference with this bug is that the life-cycle is called in a state initialized by `forRoot`.

Here is a more advanced version of the minimal demo that still emits life-cycle events: https://stackblitz.com/edit/ngxs-simple-k8i8y6

## What is the motivation / use case for changing the behavior?

Want to use life-cycle hooks to set authentication state.

## Environment

<pre><code>
Libs:
- @angular/core version: 7.2.7
- @ngxs/store version: 3.4.0/3.4.2
Also these plugins in-use:
```
    "@ngxs/devtools-plugin": "3.4.2",
    "@ngxs/logger-plugin": "3.4.2",
    "@ngxs/router-plugin": "3.4.2",
    "@ngxs/storage-plugin": "3.4.2",
```

Browser:
- [ x ] Chrome (desktop) version 73
- [ ] Chrome (Android) version XX
- [ ] Chrome (iOS) version XX
- [ ] Firefox version XX
- [ ] Safari (desktop) version XX
- [ ] Safari (iOS) version XX
- [ ] IE version XX
- [ ] Edge version XX
 
For Tooling issues:
- Node version: 8.11.3  
[...]

Issue #1007, opened 2019-04-17
How to mock dispatch


uiii:
Hi, I'm creating unit tests and have a problem when testing actions. I would like to mock `dispatch` function in a way it really dispatches only a tested action and any additional dispatch call are only spied. 

Something like this:
```ts
    it('should login with valid credentials', async (done) => {
        const action = new Login(validCredentials)

        const dispatchSpy = spyOn(store, 'dispatch').withArgs(action).and.callThrough()

        store.dispatch(action)

        await expect(store.selectSnapshot(AuthState.getError)).toBeFalsy()
        await expect(store.selectSnapshot(AuthState.getPending)).toBeTruthy()

        authTrigger.next()

        const dispatchedLoginSuccessAction = [].concat(...dispatchSpy.calls.allArgs()).find(a => a instanceof LoginSuccess)

        await expect(dispatchedLoginSuccessAction).toBeTruthy()
        await expect(dispatchedLoginSuccessAction.payload).toEqual(token)
    })
```

**BUT!** This doesn't work, because as I investigate, the `store.dispatch` is different from the `dispatch` function in the action context. I know I can use the construction like this without mocking (and it works):
```ts
        actions$.pipe(ofActionDispatched(LoginSuccess)).subscribe(async (action) => {
            await expect(action).toBeTruthy()
            done()
        })
```

**BUT!** I don't want to actually dispatch additional actions because of side effects. Consider the tested action dispatches an action from another module, so I would have to mock all services which causes side effects in that module.

I've found out the actual dispatch to be mocked is the one in the `InternalStateOperations` object, but I don't know how to mock it.

**QUESTION** So what is the proper way to make tests like this?

splincode:
@uiii hi, could you please help me with your idea.

uiii:
@splincode Hi, perhaps, what do you need to know?

uiii:
My intention to report this issue was I started creating unit tests. But I'm not skilled at it and work with ngxs only few weeks. As I studied, I've found out unit test should really test only one unit of code and all dependencies shloud be mocked as much as possible. This make sense to me. What I didn't like at action testing as described here: https://ngxs.gitbook.io/ngxs/recipes/unit-testing is that the store is set up and the action is actually dispatched which may cause side-effects so it is more like integration testing. Few days ago I've found new way how to accomplish clean unit testing and it is mocking the action context and run the action handler directly without dispatching the action:

```ts
let indexState: IndexState
[...]

Issue #950, opened 2019-03-30
Mock return of store.select in component unit test


AngeloETH:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[x] Performance issue
[ ] Feature request
[ ] Documentation issue or request
[ ] Support request => https://github.com/ngxs/store/blob/master/CONTRIBUTING.md
[ ] Other... Please describe:
</code></pre>

## Current behavior

component.spec.ts

```
describe('MyComponent', () => {
  let store: Store;
  let component: MyComponent;
  let fixture: ComponentFixture<MyComponent>;

  @State<MyStateModel>({
    name: 'MyState',
    defaults: {
      myStates: {
        [TESTING_ID]: <SingleMyState>{
          status: 'status',
        }
      },
    }
  })
  class MyState {
    static getStatus(id: string): () => { } {
      const selector = createSelector([MyState], (state: MyStateModel) => {
           return state.MyState.myStates[TESTING_ID].status ;
     })
    }
  }
  
   beforeEach(async(() => {
    TestBed.configureTestingModule({
      imports: [ MatCardModule ],
      declarations: [MyComponent ],
      providers: [ Store ],
    })
    .compileComponents();
  }));

  beforeEach(() => {
    store = TestBed.get(Store);
    const selected = MyState.getStatus(TESTING_ID);
    console.log(selected);
    fixture = TestBed.createComponent(MyComponent);
    component = fixture.componentInstance;
    fixture.detectChanges();
  });
  
  it('should create store', () => {
    expect(store).toBeTruthy();
  });
})
```

## Expected behavior

Expected log print 'status',  which is assigned to `state.MyState.myStates[TESTING_ID].status`.
In th ngOnInit of MyComponent, Jest executes ngOnInit and get error of following snapshot `this.store.select(MyState.getStatus(TESTING_ID)).pipe(...)`
 gets undefined of pipe. 
## What is the motivation / use case for changing the behavior?

Mocking the return observable of  `this.store.select(MyState.getStatus(TESTING_ID))` so that I can test my component

## Environment

<pre><code>
Libs:
- @angular/core version: X.Y.Z
- @ngxs/store version: X.Y.Z

Browser:
- [ ] Chrome (desktop) version XX
- [ ] Chrome (Android) version XX
- [ ] Chrome (iOS) version XX
- [ ] Firefox version XX
- [ ] Safari (desktop) version XX
[...]

Issue #833, opened 2019-02-10
Documentation for unit testing Actions 


mohan-m-u:
I am having some issues with testing an Action which makes a API request and on success it dispatch an action which transform that response into a particular format .Is there any way to test  API Action using Jest and check if the store is mutating based on API response.

splincode:
This problem is now considered here: ngxs-labs/ngxs-testing#2

splincode:
Released in 3.5.0

PR #1034, opened 2019-05-04
test(store): add @ngxs/store/internal/testing

splincode:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

## What is the new behavior?

own ability to manage testing more convenient

```TS
import { NgxsTestBed } from '@ngxs/store/testing';

describe('Zoo', () => {

  it('it toggles feed', (done) => {
    const { store } = NgxsTestBed.configureTestingState([ ZooState ]);

    store.dispatch(new FeedAnimals());
    store.selectOnce(state => state.zoo.feed).subscribe(feed => {
      expect(feed).toBe(true);
      done();
    });
  });

});
```

```
[ ] Yes
[x] No
```


markwhitfeld:
@splincode What does this provide to the user that could not just be done simply through `TestBed`?

splincode:
- First, at least we can get rid of the use of scattered setup functions in our code.

image

- Secondly, using the usual TestBed users can not catch the method call `ngxsAfterBootrstap` in their state. Thanks to our wrapper, users can always test the full life cycle.

- Third, we can now implement related issues:

https://github.com/ngxs/store/issues/1007
https://github.com/ngxs/store/issues/950
https://github.com/ngxs/store/issues/843
https://github.com/ngxs/store/issues/833
https://github.com/ngxs/store/issues/482
https://github.com/ngxs/store/issues/533

- Improved testing

```
describe('Zoo', () => {
  let store: Store;

  beforeEach(() => {
     const describer = NgxsTestBed.configureTestingState(
            [ ZooState ],
            { developmentMode: false  },
           [ 
              FeatureModule,
              NgxsStoragePluginModule.forRoot() 
           ]
     );

     store = describer.store;
  })

  it('it toggles feed', (done) => {
     // use store
  });

});
```

markwhitfeld:
I believe that the creation of a `@ngxs/store/testing` package is essential and a high priority but I think that we need to take small steps to improve this. We need to create a coherent API that covers all testing needs.

IMO, the feature introduced by this particular PR is more focused on testing the `ngxsAfterBootstrap` in their state, not so much the general setup method. I also would not want to add a TestBed that does all the additional work for triggering the bootstrap but adds an additional cost to each test. Not ideal for unit testing.

I propose the following:
- We create a `@ngxs-labs/testing` labs project so that we can quickly experiment, iterate and get feedback
- We create helpers that directly solve each of the problems listed in the issues above
- We elicit as much feeback/contribution as possible from the community
- Then get it into the `@ngxs/store/testing` lib ASAP!

markwhitfeld:
PS. Down the line if we do end up creating our own `NgxsTestBed.configureTestingState` I would think it would be better to have a method signature that is compatible with the Angular `TestBed.configureTestingState`, but ours just adds additional configuration options and returns a describer that contains `store` and other objects convenient for ngxs testing.

splincode:
@markwhitfeld One of the main problems, why I don’t want to do this in a lab project, is because it’s not convenient to test on our internal tests.  Having translated internal tests for our development we will be able to work out the API ourselves.

markwhitfeld:
Ok, could we then make these test helpers internal to our tests for now then?
We can move them to the sub package later

splincode:
`@ngxs/store/experimental/testing`

uiii:
Hi, my intention to report the issue #1007  was I started creating unit tests. But I'm not skilled at it and work with ngxs only few weeks. As I studied, I've found out unit test should really test only one unit of code and all dependencies shloud be mocked as much as possible. This make sense to me. What I didn't like at action testing as described here: https://ngxs.gitbook.io/ngxs/recipes/unit-testing is that the store is set up and the action is actually dispatched which may cause side-effects so it is more like integration testing. Few days ago I've found new way how to accomplish clean unit testing and it is mocking the action context and run the action handler directly without dispatching the action:

```ts
let indexState: IndexState
let indexStateContextMock: StateContextMock<IndexStateModel>

beforeEach(() => {
    indexState = new IndexState(indexServiceMock)
    indexStateContextMock = mockStateContext<IndexStateModel>(IndexState) 
})

it('should fetch modules', async (done) => {
    indexState.fetchModules(indexStateContextMock, new FetchModules).subscribe(() => {
        await expect(IndexState.getPending(indexStateContextMock.getState())).toBeTruthy()
        await expect(indexStateContextMock.dispatch).toHaveBeenCalledWith(new ModulesFetched(fetchedModulesFixture))
        done()
    })
})
```

the mock helpers are this:

```ts
export type StateContextMock<T> = jasmine.SpyObj<StateContext<T>>

export function mockStateContext<T>(stateClass: any): StateContextMock<T> {
    let values = stateClass['NGXS_OPTIONS_META'].defaults

    return {
        getState: jasmine.createSpy('getState').and.callFake(() => values),
        setState: jasmine.createSpy('setState').and.callFake((val: T) => values = val),
        patchState: jasmine.createSpy('setState').and.callFake((val: Partial<T>) => values = { ...values, ...val }),
        dispatch: jasmine.createSpy('dispatch').and.callFake(() => of())
    }
}
```

I'm quite happy with it. There is one problem that I have to use the `'NGXS_OPTIONS_META'` constant directly because I can't import it as it is in internal package.

uiii:
My colleague mentioned that with this I didn't test if the action decorator is correctly set with the appropriest action handler. I solved it with checking the action handler metadata. This is not possibly ideal becaouse it relies on internal stuff but i works.

```ts
export function getActionFunctions(stateClass: any, actionClass: any): Function[] {
    return stateClass['NGXS_META'].actions[actionClass.type].map(meta => stateClass.prototype[meta.fn])
}

...

it('should have fetch modules action function', async () => {
    await expect(getActionFunctions(IndexState, FetchModules)).toContain(indexState.fetchModules)
})
```

splincode:
@markwhitfeld ping

splincode:
@markwhitfeld I add this PR for internal testing and fixes
https://github.com/ngxs/store/issues/917

markwhitfeld:
@splincode sorry I didn't get back to you. I have had a very busy week.
It's internal, so we can iterate in this a bit. Can always ship it to a labs package later if we choose to go that way as the first public api of this.

Issue #541, opened 2018-08-13
🐞[BUG]: Incorrect behavior when using @Selector


abalad:
### Versions

```
* ngxs: 3.1.4
* @angular/core: 6.0.0
```

### Repro steps

Stackblitz / Github link: https://stackblitz.com/edit/ngxs-simple-selector-issue

* Create a @Selector that uses map to change some state value.
* Use this @selector with @Select in any component

### Observed behavior

```
Select is triggering changes to the Selector created with map even though it does not change the value it uses.
```
![image](https://user-images.githubusercontent.com/5459282/44063521-28febcd8-9f37-11e8-9dbc-ac704080de94.png)

### Desired behavior

The correct behavior would be that it does not fire the changes unless it is changed in the state the value that it is using

markwhitfeld:
Hi @abalad 
This is how memoization works. The returned value is cached based on the same input value.
In this case modifying the `counter` property results in a change to the container state which is then a new value passed to the `issueWithMap` selector which in turn results in a new calculated result.

You can get the behavior you are expecting by breaking down the `issueWithMap` selector's dependency to only be the `other` prop. This can be done using a composite selector but unfortunately (in ngxs v3) any composite selectors defined within the state class automatically take a dependency on the container class. So, to get this working you would need to move this selector to an unrelated class (this will be fixed in v4 and not necessary) so that it only takes a dependency on the provided selector.

Here is an example evolved from yours in a stackblitz:
https://stackblitz.com/edit/ngxs-simple-selector-issue-yrfhnp

abalad:
@markwhitfeld Thank you. What I was left in doubt is because the:

```
@Selector () static others (state: CounterStateModel) {
     return state.other;
}
```
Does not issue changes even though I change the `state.counter`. 
Already when I realize a `map() `in him it is shot. 

That's what I do not understand.

markwhitfeld:
The `others` selector re-evaluates when the state changes but because it returns the same value as before the subscription does not fire. The observable has a `distinctUntilChanged` operator on it so the same returned value is ignored.

The `Object.keys(...)` call is returning a new value whenever the selector is re-evaluated which gets through the `distinctUntilChanged`. This is why it is different to the `others` selector.

Hope this answers all your questions ;-)

abalad:
@markwhitfeld Thank you so much. Now I understand.

abalad:
To finalize the solution then final.

[...]

PR #1047, opened 2019-05-10
feat: expose SelectorOptions decorator

markwhitfeld:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Selector options is only exposed as a global setting. 
The new `@SelectorOptions` decorator was left internal by PR #1029.

Issue Numbers...
See: #386 (comment)
Related: #541 #446
Related PRs: #858, #954, #1015, #1029

## What is the new behavior?

Expose the new `@SelectorOptions` decorator and add the relative docs for it.

```
[ ] Yes
[X] No
```

### DIFF
--- a/packages/store/src/decorators/selector.ts
+++ b/packages/store/src/decorators/selector.ts
@@ -26,7 +26,8 @@ export function Selector(selectors?: any[]) {
               }
             );
           return memoizedFn;
-        }
+        },
+        originalFn
       };
     } else {
       throw new Error('Selectors only work on methods');
--- a/packages/store/src/decorators/selector.ts
+++ b/packages/store/src/decorators/selector.ts
@@ -8,7 +8,7 @@ export function Selector(selectors?: any[]) {
     if (descriptor.value !== null) {
       const originalFn = descriptor.value;
       let memoizedFn: any = null;
-      return {
+      const newDescriptor = {
         configurable: true,
         get() {
           // Selector initialisation defered to here so that it is at runtime, not decorator parse time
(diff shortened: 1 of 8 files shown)


packages/store/types/tests/selector-options.lint.ts
```diff
@@ -0,0 +1,95 @@
...
+      }
+    }
+  });
+
+  it('should not be a valid parameter decorator', () => {
+    @State<string>({ name: 'my-state' })
+    class MyState {
+      @Selector([MyState])
+      getMyState(
+        @SelectorOptions({ injectContainerState: false }) // $ExpectError
```
splincode: decorated field?

splincode:
build failed

packages/store/types/tests/selector-options.lint.ts
markwhitfeld: This is just to make it explicit that it cannot be used to decorate a parameter.
There is no typings restriction that we can do for fields or accessors because both of these match the method decorator signature.

splincode: ok

splincode:
Great job, you just have to fix the pipeline!

packages/store/src/decorators/selector.ts
```diff
@@ -26,9 +26,11 @@ export function Selector(selectors?: any[]) {
               }
             );
           return memoizedFn;
-        },
-        originalFn
+        }
       };
+      // Add hidden property to descriptor
```
splincode: Why add hidden?

markwhitfeld: Because it was an internal trick I had to do to make the selector work independent of order.
I would not want this exposed in the public type information.

Issue #1049, opened 2019-05-12
[Router Plugin] does not change if the navigation is on query parameters


ThibaudAV:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[x] Bug report  
[ ] Performance issue
[ ] Feature request
[ ] Documentation issue or request
[ ] Support request => https://github.com/ngxs/store/blob/master/CONTRIBUTING.md
[ ] Other... Please describe:
</code></pre>

## Current behavior

When I navigate with the next and previous buttons of the browser the RouterNavigation does not evolve if only the query parameters change 

Ex : 
First search: /todo/123/search?q=toto
-> [Router] RouterNavigation OK (with `dispatch(new Navigate())`)
Second research: /todo/123/search?q=titi
-> [Router] RouterNavigation OK (with `dispatch(new Navigate())`)
Go back:  /todo/123/search?q=toto
-> no [Router] RouterNavigation :/

## Expected behavior
The router plugin should detect the change and update with RouterNavigation its store

## Minimal reproduction of the problem with instructions

https://stackblitz.com/edit/ngxs-simple-4w3qej

* Step 1 : Click on `Click Me`
* Step 2 : Click on `Click Me`
* Step 3 : Uses the browser go back
not easy to reproduce here because stackblitz reloads the page, but in the console the event RouterNavigation is not sent

## What is the motivation / use case for changing the behavior?
 
🤷‍♂ Have the plugin router store with the right data

## Environment

<pre><code>
Libs:
- @angular/core version: 7.2.9
- @ngxs/store version: 7.3.0
- @ngxs/router-store version: 7.3.0

Browser:
- [x] Chrome (desktop) version XX
- [ ] Chrome (Android) version XX
- [ ] Chrome (iOS) version XX
- [x] Firefox version XX
- [ ] Safari (desktop) version XX
- [ ] Safari (iOS) version XX
- [ ] IE version XX
- [ ] Edge version XX
 
For Tooling issues:
- Node version: v11.10.1
- Platform:  Mac

Others:
N/A
</code></pre>

arturovt:
Dup of #922.

markwhitfeld:
This should be fixed in the latest @dev version.
Please could you test it out and let us know if it fixes your problem?

splincode:
Released in 3.5.0

PR #1059, opened 2019-05-16
feat(router-plugin): add `RouterDataResolved` action

markwhitfeld:
## PR Checklist

## PR Type
```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: #946 #922 #1049 
Progression of PR #952 

```
[ ] Yes
[x] No
```


markwhitfeld:
@arturovt turns out that the test failure was caused by a race condition encountered if jasmine ran the tests in a specific order. The race condition was caused by the case if testA was run before testB and both tests used the same route. As a result, testB had code that was waiting for the route to change after navigating to the url, but this was never triggered because the document was already located at that url!

packages/router-plugin/src/router.state.ts
```diff
@@ -118,8 +116,13 @@ export class RouterState {
     });
   }
 
-  private navigationEnd(): void {
-    this.routerStateSnapshot = this._serializer.serialize(this.lastResolvedStateSnapshot);
+  private guardsCheckEnd(routerState: RouterStateSnapshot): void {
+    if (this.routerStateSnapshot) {
+      // routerState = { ...this.routerStateSnapshot, ...routerState };
```
splincode: Is it really worth leaving comments in the code?

markwhitfeld: Ah, that was Artur's fix, but it turned out that it was not needed. I'll remove it.

szalonna:
Shouldn't the [`RouterAction`](https://github.com/ngxs/store/blob/master/packages/router-plugin/src/router.actions.ts#L79) type contain the `RouterDataResolved` action?

PR #1042, opened 2019-05-08
feat(store): expose StateClass for @ngxs/store/testing

splincode:
## PR Checklist
Please check if your PR fulfills the following requirements:

- [ ] The commit message follows our guidelines: https://github.com/ngxs/store/blob/master/CONTRIBUTING.md#commit
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

```
[ ] Bugfix
[ ] Feature
[ ] Code style update (formatting, local variables)
[x] Refactoring (no functional changes, no api changes)
[ ] Build related changes
[ ] CI related changes
[ ] Documentation content changes
[ ] Other... Please describe:
```

PR #1070, opened 2019-05-18
fix(store): trim down StateClass interface and move it to internals subpackage

markwhitfeld:
Reverts ngxs/store#1042
This PR should never have been merged.
It exposes too much of the internal NGXS implementation details.
Hopefully the new extensibility model allows for the hooks that this was intended to expose.


markwhitfeld:
ok, I have opted to expose a slimmed down interface called `StateClass` from `@ngxs/store/internals`.
There is another fuller interface for internal use called `StateClassInternal`.

PR #1112, opened 2019-06-02
feat(store): support Angular 8, TypeScript 3.4

splincode:
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[x] Build related changes
[ ] CI related changes
[x] Other... Please describe:
```

Only support Angular 7, TypeScript 3.1

## What is the new behavior?

Support Angular 8, TypeScript 3.4

After upgrading to Angular 8, TypeScript 3.4 all types were broken and tests did not pass!
In order to avoid breaking change, the types were set to `any`.
In the next pull request, we need to apply a new typing to support state operators.

```
[ ] Yes
[x] No
```


fetters5:
I see you've referenced jest 8.0.1, you may want to wait for PR #[330](https://github.com/meltedspark/angular-builders/pull/330) to be merged or you may not be able to run tests

whichwit:
That referenced PR actually bumped jest builder to 8.0.2, not sure how much it matters for this PR.

splincode:
@arturovt What do you think about state operators and their typing?

markwhitfeld:
Not too happy about the `any` typing of the state operators. 
Could you explain the errors that were causing this to fail?

arturovt:
@splincode 
I'm not a fan of it. As I know you've been fighting for the good typings support of this library. I wouldn't like state operators to support `any` type.

splincode:
@arturovt Um, by no means, I didn’t want to worsen the types, I just tried to at least pass the tests and that the assembly worked. As you understand these operators you have developed, and typing does not work for your operators in TS 3.4, a compilation error occurs.

packages/store/operators/src/compose.ts
```diff
@@ -1,7 +1,16 @@
 import { StateOperator } from '@ngxs/store';
 
-export function compose<T>(...operators: StateOperator<T>[]) {
```
arturovt: image

packages/store/operators/src/append.ts
```diff
@@ -1,13 +1,22 @@
...
+ * In the next pull request, we need to apply a new typing to support state operators.
+ * TODO: Need to fix types
+ */
+
 /**
  * @param items - Specific items to append to the end of an array
  */
-export function append<T>(items: T[]) {
-  return function appendOperator(existing: Readonly<T[]>): T[] {
+export function append<T>(items: T[]): StateOperator<any> {
```
arturovt: image

arturovt:
If tests don't pass - this doesn't mean that we have to change everything to `any` type, just because "to make tests happy". This means we're doing something wrong.

packages/router-plugin/package.json
```diff
@@ -6,8 +6,8 @@
   "sideEffects": true,
   "peerDependencies": {
     "@ngxs/store": "^0.0.0",
-    "@angular/core": ">=5.0.0 <8.0.0",
-    "@angular/router": ">=5.0.0 <8.0.0",
+    "@angular/core": ">=5.0.0 <9.0.0",
+    "@angular/router": ">=5.0.0 <9.0.0",
```
kuncevic: As per this twit https://twitter.com/stephenfluin/status/1129429898047643648 suggested to set peer dependency to ^8.0.0 || ^9.0.0-0.

packages/store/operators/src/compose.ts
splincode: Yeah, this build is correct for me also, but not working for test

packages/store/operators/src/append.ts
splincode: Yeah

packages/router-plugin/package.json
splincode: Good catch

arturovt:
@splincode What do you think if support of Angular 8 would be a part of NGXS v4?

splincode:
@arturovt My company switched to Angular 8 and our team  switched also projects, but if NGXS does not have Angular 8 support (our company has set up a common CI, our build fall if warnings are shown when installing packages), then our architect will decide to switch to NGRX 8  to integrate with Angular 8. And then I will have to switch to NGRX in my projects due to poor support for NGXS.  I still hope that the NGXS will not keep up with current trends.  NGXS for a long time did not release anything, no new version and we are very worried about this topic.

arturovt:
I don't mind switching to Angular 8 but without breaking typings and setting them to `any`.

splincode:
I hope in the future NGXS will support Angular 8 and Typescript 3.4. I did everything I could.

markwhitfeld:
@arturovt and @splincode I am currently working on the typings to hopefully resolve this.

Issue #1118, opened 2019-06-06
Storage Plugin SSR localStorage is not defined


Dav1dde:
## I'm submitting a...
<pre><code>[ ] Regression (a behavior that used to work and stopped working in a new release)
[X] Bug report  
[ ] Performance issue
[ ] Feature request
[ ] Documentation issue or request
[ ] Support request => https://github.com/ngxs/store/blob/master/CONTRIBUTING.md
[ ] Other... Please describe:
</code></pre>

## Current behavior

SSR fails with
```
ReferenceError: localStorage is not defined
    at engineFactory (app/dist/server.js:152303:9)
    at _callFactory (app/dist/server.js:24259:20)
    at _createProviderInstance (app/dist/server.js:24215:26)
    at initNgModule (app/dist/server.js:24145:32)
    at new NgModuleRef_ (app/dist/server.js:24872:9)
    at Object.createNgModuleRef (app/dist/server.js:24861:12)
    at NgModuleFactory_.create (app/dist/server.js:27396:25)
    at app/dist/server.js:20742:43
    at ZoneDelegate.invoke (app/dist/server.js:546:26)
    at Object.onInvoke (app/dist/server.js:20276:33)
```

I have seen #465 but this is either a regression or was never really fixed. I am using the `STORAGE_ENGINE` workaround but this does **not** solve the issue, localStorage is still undefined and throws an exception.

Relevant code from `server.js`:
```js
function engineFactory(options) {
          if (options.storage === 0 /* LocalStorage */) {
              // todo: remove any here
              return (/** @type {?} */ (localStorage));
          }
          else if (options.storage === 1 /* SessionStorage */) {
              return (/** @type {?} */ (sessionStorage));
          }
          return null;
}
```

Identical to: https://github.com/ngxs/store/blob/master/packages/storage-plugin/src/storage.module.ts#L23

## Expected behavior

SSR does not fail and storage is either ignored or stubbed.

## Minimal reproduction of the problem with instructions

Angular Universal example, add NGXS and NGXS Storage modules.

## Environment

<pre><code>Libs:
- @angular/core version: 7.2.0
- @ngxs/store version: 3.4.3 (also happens in current master)

 
For Tooling issues:
- Node version: 11.15.0  
- Platform:  Linux
</code></pre>

splincode:
You can send pull request

Dav1dde:
I will look into it when I get home.

arturovt:
@Dav1dde I will investigate today

[...]

PR #1119, opened 2019-06-06
fix(storage-plugin): fix undefined localStorage error during SSR (#1118)

Dav1dde:
## PR Checklist

## PR Type

```
[X] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: #1118 

## What is the new behavior?

Storage during SSR is emulated with an empty in memory storage engine, which can still be modified by the user by providing a different storage engine.

```
[ ] Yes
[X] No
```

### DIFF
--- a/packages/storage-plugin/src/storage.module.ts
+++ b/packages/storage-plugin/src/storage.module.ts
@@ -1,4 +1,5 @@
-import { NgModule, ModuleWithProviders, InjectionToken } from '@angular/core';
+import { NgModule, ModuleWithProviders, PLATFORM_ID, InjectionToken } from '@angular/core';
+import { isPlatformServer } from '@angular/common';
 import { NGXS_PLUGINS } from '@ngxs/store';
 
 import { NgxsStoragePlugin } from './storage.plugin';
@@ -20,7 +21,14 @@ export function storageOptionsFactory(options: NgxsStoragePluginOptions) {
   };
 }
 
-export function engineFactory(options: NgxsStoragePluginOptions): StorageEngine | null {
+export function engineFactory(
+  options: NgxsStoragePluginOptions,
+  platformId: any
+): StorageEngine | null {
+  if (isPlatformServer(platformId)) {
+    return null;
+  }
+
   if (options.storage === StorageOption.LocalStorage) {
     // todo: remove any here
     return <any>localStorage;
@@ -56,7 +64,7 @@ export class NgxsStoragePluginModule {
         {
           provide: STORAGE_ENGINE,
           useFactory: engineFactory,
-          deps: [NGXS_STORAGE_PLUGIN_OPTIONS]
+          deps: [NGXS_STORAGE_PLUGIN_OPTIONS, PLATFORM_ID]
         }
       ]
     };
--- a/packages/storage-plugin/src/storage.plugin.ts
+++ b/packages/storage-plugin/src/storage.plugin.ts
@@ -1,4 +1,5 @@
-import { Inject, Injectable } from '@angular/core';
+import { PLATFORM_ID, Inject, Injectable } from '@angular/core';
+import { isPlatformServer } from '@angular/common';
 import {
   NgxsPlugin,
   setValue,
@@ -21,10 +22,15 @@ import { tap } from 'rxjs/operators';
 export class NgxsStoragePlugin implements NgxsPlugin {
   constructor(
     @Inject(NGXS_STORAGE_PLUGIN_OPTIONS) private _options: NgxsStoragePluginOptions,
-    @Inject(STORAGE_ENGINE) private _engine: StorageEngine
+    @Inject(STORAGE_ENGINE) private _engine: StorageEngine,
+    @Inject(PLATFORM_ID) private _platformId: any
   ) {}
 
   handle(state: any, event: any, next: NgxsNextPluginFn) {
+    if (isPlatformServer(this._platformId) && this._engine === null) {
+      return next(state, event);
+    }
+
     const options = this._options || <any>{};
     const matches = actionMatcher(event);
     const isInitAction = matches(InitState) || matches(UpdateState);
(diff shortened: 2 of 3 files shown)


Dav1dde:
This fixes the mentioned issue, I am just not sure if this is a solution you guys are happy with. Please give me a short review if you have other ideas of fixing the issue.

@splincode @arturovt

Dav1dde:
Regarding tests I'd also welcome your input, adding the storage module to [app.server.module.ts](https://github.com/ngxs/store/blob/master/integration/app/app.server.module.ts) or moving it from [app.browser.module.ts](https://github.com/ngxs/store/blob/master/integration/app/app.browser.module.ts) to [app.module.ts](https://github.com/ngxs/store/blob/master/integration/app/app.module.ts) should be enough?

I unfortunately cannot run the SSR tests locally, so it's hard for me to just try it out.

packages/storage-plugin/src/storage.module.ts
```diff
@@ -20,7 +22,14 @@ export function storageOptionsFactory(options: NgxsStoragePluginOptions) {
   };
 }
 
-export function engineFactory(options: NgxsStoragePluginOptions): StorageEngine | null {
+export function engineFactory(
+  options: NgxsStoragePluginOptions,
+  platformId: any
+): StorageEngine | null {
+  if (isPlatformServer(platformId)) {
```
arturovt: Did you check if it's working?

Dav1dde: Yes, it works. I'll check in my test project so you can easily try it.

arturovt: Hm, I just copied your code, built package, copied bundle and it prints `undefined`:
```TS
console.log(`platformId = ${platformId}`);
if (isPlatformServer(platformId)) {
  return new InMemoryStorageEngine();
}
```

arturovt: Also the problem is that you can easily just return `null` as `InMemoryStorageEngine` seems meaningless for me. What's the purpose of it? You can also inject `PLATFORM_ID` into plugin and check if it's executed on the server and just return `next(...)`:
```TS
  if (isPlatformServer(this.platformId)) {
    return next(state, event);
  }

  ...
}
```

arturovt: if `InMemoryStorageEngine` is created on the server side which is bootstrapped only once - this means that the instance will be the same for all requests right?

Dav1dde: image
So this should work.

If you return `null` instead of the `InMemoryStorageEngine` the build will fail because no storage engine is selected (so you would have to manually provide one via `app.server.module.ts`

Injecting it into plugin and nexting is a different behaviour, in the long run I want to use the storage plugin server side and provice a custom storage engine to pre-initialize the store for server side rendering with defaults (e.g. theme, language etc.) so just skipping the plugin would not be ideal.

About being created only once, in my setup every request re-creates the storage engine (checking with debugger), but maybe there is something about SSR I am not aware about.

Dav1dde:
@arturovt  Test Repository: https://github.com/Dav1dde/ngxs-pr-1119-test

arturovt:
@Dav1dde I believe you and can't check right now, but the `InMemoryStorageEngine` should be removed. Many users will receive different states because the storage will be the same for all consumers. Also add a check inside `handle` method.

Dav1dde:
So we would return null in `engineFactory`:
```js
if (isPlatformServer(platformId)) {
  return null: // new InMemoryStorageEngine();
}
```
And additionally skip the plugin when platform = server?

--- 

Out of curiosity, I was under the impression that angular universal has to bootstrap the application for every request, simply because stuff like the URL changes and maybe app initializer or other services may inject the request or response. In which case this wouldn't be problematic.

arturovt:
@Dav1dde You're right but this in memory engine  seems meaningless, I don't see a purpose of it. The user reloads page - the new in memory engine is created right? Or I'm missing something.

Dav1dde:
You're right it's basically useless if the user does not inject `STORAGE_ENGINE` anywhere else and does not use it otherwise and rely on it - I am not sure if that's a usecase.

packages/storage-plugin/src/storage.plugin.ts
```diff
@@ -21,10 +22,15 @@ import { tap } from 'rxjs/operators';
 export class NgxsStoragePlugin implements NgxsPlugin {
   constructor(
     @Inject(NGXS_STORAGE_PLUGIN_OPTIONS) private _options: NgxsStoragePluginOptions,
-    @Inject(STORAGE_ENGINE) private _engine: StorageEngine
+    @Inject(STORAGE_ENGINE) private _engine: StorageEngine,
+    @Inject(PLATFORM_ID) private _platformId: any
   ) {}
 
   handle(state: any, event: any, next: NgxsNextPluginFn) {
+    if (isPlatformServer(this._platformId)) {
```
Dav1dde: @arturovt Does it make sense to modify this check into `if (isPlatformServer(this._platformId) && this._engine === null)`.

In case the user supplies it's own storage engine it will still be used?

arturovt: Sure, good idea.

Dav1dde: Done

Dav1dde:
Awesome, thanks for your help.

Let me know if you want the commits squashed.

arturovt:
@Dav1dde We squash, no worries :D

splincode:
@Dav1dde I will happy if you add plugin

```ts
NgxsStoragePluginModule.forRoot({ key: [TODOS_STORAGE_KEY] })
```

to app.server.module.ts

https://github.com/ngxs/store/blob/master/integration/app/app.server.module.ts

If the tests pass, then the SSR works with storage-plugin!

Dav1dde:
@splincode done and tests seem to pass.

PR #1128, opened 2019-06-13
refactor: improve state operator typings

arturovt:
## PR Checklist

## PR Type
```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

The current typings are not explicit

## What is the new behavior?
To add explicit return type typings for the state operators.

```
[ ] Yes
[x] No
```

### DIFF
--- a/integration/app/store/todos/todos.state.ts
+++ b/integration/app/store/todos/todos.state.ts
@@ -25,7 +25,7 @@ export class TodosState {
   @Action(SetPrefix)
   public setPrefix({ setState }: StateContext<TodoStateModel>) {
     setState(
-      patch({
+      patch<TodoStateModel>({
         pizza: patch({
           model: patch({
             toppings: (topping: any) => 'Mr. ' + topping
--- a/integration/server.ts
+++ b/integration/server.ts
@@ -40,13 +40,10 @@ app.use((req, res, next) => {
   next();
 });
 
-app.engine(
-  'html',
-  ngExpressEngine({
-    bootstrap: AppServerModuleNgFactory,
-    providers: [provideModuleMap(LAZY_MODULE_MAP)]
-  })
-);
+app.engine('html', ngExpressEngine({
+  bootstrap: AppServerModuleNgFactory,
+  providers: [provideModuleMap(LAZY_MODULE_MAP)]
+}) as any);
 
 app.set('view engine', 'html');
 app.set('views', '.');
--- a/package.json
+++ b/package.json
@@ -117,7 +117,7 @@
     "codelyzer": "^4.5.0",
     "core-js": "^2.6.0",
     "cross-env": "^5.2.0",
-    "dtslint": "^0.7.6",
+    "dtslint": "^0.8.0",
     "express": "^4.17.1",
     "fs-extra": "^7.0.1",
     "husky": "^1.1.0",
--- a/packages/store/operators/src/append.ts
+++ b/packages/store/operators/src/append.ts
@@ -1,13 +1,15 @@
+import { StateOperator } from '@ngxs/store';
+
 /**
  * @param items - Specific items to append to the end of an array
  */
-export function append<T>(items: T[]) {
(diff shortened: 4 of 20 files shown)


splincode:
https://github.com/markwhitfeld/store/blob/8b92796ba9c8c815bafed40c3c2fb3fe5b50ff56/packages/store/types/tests/state-operators.iif.lint.ts

arturovt:
Yeah, I forgot, I will paste today

splincode:
I almost fixed the problem and want to add changes

arturovt:
Need to notify Mark because he's also working on the fix

splincode:
@markwhitfeld @arturovt Guys, I suggest to make release of version 4.0.0. Because 3.5.0 includes a huge number of changes, and it can break everything.

- Where we will do support Angular 8 and go to support only TypeScript 3+.
- It will also explain why we delayed the release so long.
- And also we can make a number of changes that we wanted to make as part of the task: https://github.com/ngxs/store/issues/827

arturovt:
@splincode 
```TS
(async () => await Mark())()
```
:joy:

arturovt:
I guess that we could release 3.5 with those changes that we ALREADY HAVE, like router/web-socket plugins bug fixes and blah blah.
But I'm afraid that operators and Angular 8 could be really a breaking change so it should be v4 release.
Also the problem is that Angular 8 uses typescript 3.4 and it has performance issues that were fixed in 3.5, I'm just wondering why Angular team doesn't want to upgrade the TS version.

splincode:
We can make backports with bugfixes for 3.4.4 , but I would send new features to version 4 and make such a release.

arturovt:
We have a lot of bug fixes so minor patch to the v3.5 is very ok and can be satisfied by semver philosophy, like if we fixed only 1 bug - then 3.4.4 would be ok.

My point is:
`v3.5 (with a lot of bug fixes) --> v4 (Angular 8 / TS 3.4(or 3.5) + typings improvement`

splincode:
@arturovt Okey, up to you

splincode:
@arturovt But why we waited for release v3.5 so long? More people want integrated Angular 8. I think we have to drop the shackles of legacy

arturovt:
The problem is that now only a small part of the projects is being transferred to the new version of Angular. That's just statistics, not my subjective opinion. I've got only 1 project that uses Angular 8 in my company - mine :joy: Other projects still use Angular 6-7. 

v3.5 is for those people who cannot migrate to the new Angular, but they will receive our bug fixes, e.g. those who use router plugin. And v4 is for those who migrated to the Angular 8.

If we move those bug fixes to the v4 - so people are not able to get them (because they don't use Angular 8).

> But why we waited for release v3.5 so long?

I cannot do releases w/o Mark :(

splincode:
In general, this is a big question that is difficult to answer. But I’m completely upset that we don’t have release schedules.

markwhitfeld:
Unfortunately some of these changes to the state operator typings are indeed breaking changes.
We still need to get NGXS v3.5 out before NGXS v4 to help with the transition. Also if we start bundling these things into a v4 release it will delay it even further!

It would also be my preference if we could keep the implicit nature of the state operators... this has become an even bigger challenge with TS 3.4.

I have realized though that the issue with the typings has been introduced by breaking changes in TS and not in NGXS, so I will have a look to see what the minimum is that we can do to soften the TS3.5 transition without breaking compatibility for those on earlier versions of TS. If someone upgrades to Angular v4 and TS3.5 they will be fixing types because of that upgrade, not because of NGXS.

I will be getting deep into this in the next few days.

PS. @splincode those tweaks to the patch operator typings look amazing, I need to get my head around them because it was such a balancing act to get that right in the first place (then maybe they can be lined up for v4).

markwhitfeld:
I'm closing this PR because I would not want to lose the implicit state operator functionality.
... and if the changes on master work then we don't need these changes.

PR #1135, opened 2019-06-22
fix: typescript 3.4 has issues seeing true / false as boolean

markwhitfeld:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

After upgrading to the more recent versions of Typescript (3.4 for example) the types narrow too far and do not recognize true or false as compatible with boolean.

## What is the new behavior?
Introduced a `RepairType` conditional type to repair the type specificity.

```
[ ] Yes
[x] No
```

### DIFF
--- a/packages/store/operators/src/append.ts
+++ b/packages/store/operators/src/append.ts
@@ -1,13 +1,15 @@
+import { RepairType } from './utils';
+
 /**
  * @param items - Specific items to append to the end of an array
  */
 export function append<T>(items: T[]) {
-  return function appendOperator(existing: Readonly<T[]>): T[] {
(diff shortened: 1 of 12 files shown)


markwhitfeld:
@splincode @arturovt Please could you review this. I would like to solve the TS issues in small safe incremental typings improvements like this one.
**_PS. This is actually a fix that I had on the cards for a while because it was an issue with the typings that came up before the Angular 8 upgrade dance._**

markwhitfeld:
Merged after one review so that we can move further forward with this.

Issue #1125, opened 2019-06-12
Angular 8 Support


internalsystemerror:
## I'm submitting a...
<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[ ] Performance issue
[ ] Feature request
[x] Documentation issue or request
[ ] Support request => https://github.com/ngxs/store/blob/master/CONTRIBUTING.md
[ ] Other... Please describe:
</code></pre>

A statement as to the current status regarding Angular 8 and Universal would be appreciated as I can't find one if this exists. I can see various dependency update chores however I can see the benefit of having this ticket closed when full support is reached for anyone else finding this.

splincode:
Thank you for reporting

arturovt:
@internalsystemerror 
We know about that and we're working on that, actually @splincode already has a skeleton for upgrading to Angular 8. It's currently working with Angular 8 (I checked), but not with Ivy (it's related to `@angular/cli`, not our issue, will be fixed in 2-3 weeks as Google team says)

internalsystemerror:
@arturovt thanks for the feedback. Just planning our own upgrades and getting the status of various libraries. A ticket like this I feel is just good for reference.

jtomek:
> @internalsystemerror
> We know about that and we're working on that, actually @splincode already has a skeleton for upgrading to Angular 8. It's currently working with Angular 8 (I checked), but not with Ivy (it's related to `@angular/cli`, not our issue, will be fixed in 2-3 weeks as Google team says)

Can you please publish the version of ngxs that works with Angular 8 (without Ivy support)? I just went through a lengthy upgrade from 7 -> 8 and just now noticed that ngxs isn't supporting Angular 8.

arturovt:
@jtomek 
This is the third place where you’re telling that something doesn’t work. Create an issue with reproduce!

jtomek:
> @jtomek
> This is the third place where you’re telling that something doesn’t work. Create an issue with reproduce, please!

As people have reported, it's not working with Angular 8. I don't think it's necessary to create another issue regarding this topic. 

Maybe I should create one about the future of ngxs. I don't want to sound bitter, but I think it's a big problem that so many people are left hanging while everybody package had such a long runway to prepare for a new version. I've implemented ngxs as an integral part of my app because it's been so easy and reduced the boilerplate. But as pointed out in a different discussion here, ngxs has been stale for some time and this problem with Angular 8 is just not helping at all. 

[...]

PR #1136, opened 2019-06-23
feat(store): support Angular 8, TypeScript 3.4

splincode:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: https://github.com/ngxs/store/issues/1125

## What is the new behavior?

- Angular 8
- TypeScript 3.4.5
- Improved typings for state operators
- Full Backward compatibility for state operator without breaking changes

```
[ ] Yes
[x] No
```

## Other information

@markwhitfeld I suggest to make changes to this pull request, since there was a complete integration with Angular 8. One of the main problems I encountered is that our router plugin does not support SSR and we need to further investigate the cause.


packages/store/types/tests/state-operators.remove-item.lint.ts
```diff
@@ -120,54 +120,58 @@ describe('[TEST]: the removeItem State Operator', () => {
...
+
+    patch<Original>({ arrs: removeItem(item => item === {}) })(original); // $ExpectType Original
+    patch<Original>({ arrs: removeItem(item => item === 2) })(original); // $ExpectType Original
+    patch<Original>({ arrs: removeItem(item => item === '2') })(original); // $ExpectType Original
+    patch<Original>({ arrs: removeItem(item => item === true) })(original); // $ExpectType Original
+
+    patch<Original>({ objs: removeItem(item => item === 2) })(original); // $ExpectType Original
+    patch<Original>({ objs: removeItem(item => item === '2') })(original); // $ExpectType Original
+    patch<Original>({ objs: removeItem(item => item === true) })(original); // $ExpectType Original
+    patch<Original>({ objs: removeItem(item => item === []) })(original); // $ExpectType Original
```
markwhitfeld: None of these should be valid! That is why they were explicitly listed as expected errors below.
It seems that this PR has broadened the typings too much.

splincode: I can try to fix it, but for now it’s not quite clear how to do it without breaking another logic

packages/store/types/tests/state-operators.append.lint.ts
```diff
@@ -15,23 +15,22 @@ describe('[TEST]: the append State Operator', () => {
...
     };
 
-    patch<Original>({ nums: append<number>([]) })(original); // $ExpectType { nums: number[]; strs: string[]; bools: boolean[]; }
-    patch<Original>({ nums: append(null!) })(original); // $ExpectType { nums: number[]; strs: string[]; bools: boolean[]; }
-    patch<Original>({ nums: append(undefined!) })(original); // $ExpectType { nums: number[]; strs: string[]; bools: boolean[]; }
-    patch<Original>({ nums: append([1, 2]) })(original); // $ExpectType { nums: number[]; strs: string[]; bools: boolean[]; }
+    patch<Original>({ nums: append<number>([]) })(original); // $ExpectType Original
+    patch<Original>({ nums: append(null!) })(original); // $ExpectType Original
+    patch<Original>({ nums: append(undefined!) })(original); // $ExpectType Original
+    patch<Original>({ nums: append([1, 2]) })(original); // $ExpectType Original
```
markwhitfeld: Changing the returned expected type here may be a breaking change on patch. We need explicit dtslint tests for the `patch` operator if we are going to make this change. I will add the tests for the `patch` and `compose` operators tonight.
I have been very strict with myself following the rule that we discussed a few months back. Any typings change needs to have thorough typings tests documenting all aspects of the existing behavior before any change is done to the typings in a subsequent PR. This is why I have been solving these typings issue step by step and not all in one go.

packages/store/types/tests/state-operators.remove-item.lint.ts
splincode: Anyway, I suggest working in this pull request, since we can track that the build is successful with Angular 8

packages/store/types/tests/state-operators.append.lint.ts
splincode: I could not do it in a less complex way, however, the return type, the correct one, is quite logical.

patch<T> => return T

markwhitfeld: This may interfere with the implicit operators. I will investigate if my suspicion is true. I hope I am wrong because what you have above does feel better. :+1:

packages/store/types/tests/state-operators.remove-item.lint.ts
markwhitfeld: I disagree, we need to move forward in small steps. This is why I am breaking down my changes into small PRs.

packages/store/operators/src/internals.ts
```diff
@@ -1 +1,33 @@
...
+// RepairType<false> -> boolean
+// RepairType<'abc'> -> string
+// RepairType<{}> -> object
+export type RepairType<T> = T extends boolean
+  ? IncludeType<boolean>
+  : T extends Primitive
+  ? IncludeType<T>
+  : T extends object
+  ? IncludeType<T>
+  : never;
```
markwhitfeld: Your improvement on my `RepairType` is next level awesome by the way!!! Nicely done!

splincode:
@markwhitfeld What do you propose to do next?

splincode:
@markwhitfeld I'm a little tired of working on this, I think you should continue to do the Angular 8 update.

markwhitfeld:
You have put so much effort into this. Thank you so much for all the time you have put into this. 
This is just really hard and can have unexpected side effects.

Issue #1153, opened 2019-07-11
Router plugin leads to erased query params on load


fabioscala:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[x] Bug report  
[ ] Performance issue
[ ] Feature request
[ ] Documentation issue or request
[ ] Support request => https://github.com/ngxs/store/blob/master/CONTRIBUTING.md
[ ] Other... Please describe:
</code></pre>

## Current behavior
When the user enters an URL containing query parameters, the router plugin will redirect to the same path without query parameters. I was able to trace it down to this line:

https://github.com/ngxs/store/blob/7b0d9951db4f614b26b162633c968cbe62796441/packages/router-plugin/src/router.state.ts#L210

Without reading too much of the code, it seems that if statement compares the URL from the angular router (which contains query params) with the URL without it. The `path()` method used currently includes a boolean flag though:

```
    // TODO: vsavkin. Remove the boolean flag and always include hash once the deprecated router is
    // removed.
    path(includeHash = false)
```

## Expected behavior
No to be redirected and erase all the original query parameters

## Environment

<pre><code>
Libs:
- @angular/core version: 8.1.0
- "@ngxs/devtools-plugin": "3.4.3-dev.master-e7f64b9",
- "@ngxs/logger-plugin": "3.4.3-dev.master-7b0d995",
- "@ngxs/router-plugin": "3.4.3-dev.master-e7f64b9",
- "@ngxs/storage-plugin": "3.4.3-dev.master-e7f64b9",
- "@ngxs/store": "3.4.3-dev.master-7b0d995",

Browser:
- [x] Chrome (desktop) version XX
- [ ] Chrome (Android) version XX
- [ ] Chrome (iOS) version XX
- [ ] Firefox version XX
- [ ] Safari (desktop) version XX
- [ ] Safari (iOS) version XX
- [ ] IE version XX
- [ ] Edge version XX
 
For Tooling issues:
- Node version: XX  
- Platform:  

Others:

</code></pre>

splincode:
Please check dev version:

```
$ npm I @ngxs/{store,router-plugin}@dev
```

is there still a problem?

fabioscala:
I am already using the latest/nightly version of all, updated as of yesterday

splincode:
add stackblitz example for reproduce

fabioscala:
Here you go: https://stackblitz.com/edit/ngxs-simple-c6zex5

[...]

PR #1159, opened 2019-07-17
fix(router-plugin): serialize entered URL and compare with the recognized one

arturovt:
## PR Checklist

## PR Type
```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: #1153

## What is the new behavior?
`location.pathname` is concatenated with `location.search` as it has to be.

```
[ ] Yes
[x] No
```

### DIFF
--- a/packages/router-plugin/src/router.state.ts
+++ b/packages/router-plugin/src/router.state.ts
@@ -209,10 +209,10 @@ export class RouterState {
         // in order to redirect user to the manually entered URL if it differs from the recognized one.
         // Also `location.search` is used here in case of putting query parameters in the URL,
         // e.g. `/foo?query=some_cool_query`
-        if (url !== this._location.path()) {
-          this._router.navigateByUrl(
-            `${this._platformLocation.pathname}${this._platformLocation.search}`
-          );
+        const currentUrl = `${this._platformLocation.pathname}${this._platformLocation.search}`;
+
+        if (url !== currentUrl) {
+          this._router.navigateByUrl(currentUrl);
         }
       });
   }
--- a/packages/router-plugin/src/router.state.ts
+++ b/packages/router-plugin/src/router.state.ts
@@ -212,7 +212,9 @@ export class RouterState {
         // e.g. `/test?redirect=https://google.com/`
 
         // `url` is a recognized URL by the Angular's router, while `currentUrl` is an actual URL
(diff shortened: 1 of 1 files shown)


arturovt:
@splincode @markwhitfeld 

Just a small bugfix.

arturovt:
> Test?)

```ts
if (isAngularInTestMode()) {
  return;
}
```

?

splincode:
I mean, can you write tests?

markwhitfeld:
Thanks @arturovt!
Just a thought, could this be done using this method instead?
https://angular.io/api/common/Location#iscurrentpathequalto
This could make it more resilient around urls that are the same but differ in small aspects like some url encoded characters. Really unfamiliar with this API though, so this is a guess.

markwhitfeld:
Oh, and can you remember what this code was a workaround for?
```
if (isAngularInTestMode()) {
  return;
}
```
If it was around protecting against no `window.location` in nodejs (jest) then we have fixed that shortfall by using PlatformLocation.

arturovt:
> Thanks @arturovt!
> Just a thought, could this be done using this method instead?
> https://angular.io/api/common/Location#iscurrentpathequalto
> This could make it more resilient around urls that are the same but differ in small aspects like some url encoded characters. Really unfamiliar with this API though, so this is a guess.

Let me check tomorrow.

arturovt:
@markwhitfeld 

The problem of `isCurrentPathEqualTo` is that we basically compare the entered URL with itself, like:

```ts
// entered in the search bar
const url = '/path';
if (url === url) {
  ... do stuff
}
```

So this code:

```ts
 const currentUrl = `${this._platformLocation.pathname}${this._platformLocation.search}`;
if (this._location.isCurrentPathEqualTo(currentUrl)) {
  // ...
}
```

Is not valid, because Angular compares `pathname + search` with `pathname + search`, because `isCurrentPathEqualTo` uses `PlatformLocation` under the hood.

By using the `url` from the `RoutesRecognized` event - we're comparing the regognized one URL.

markwhitfeld:
@arturovt Are you saying that we could write it as:
```
if (this._location.isCurrentPathEqualTo(url)) {
  // ...
}
```

PR #1163, opened 2019-07-19
chore: decouple build infrastructure from root package.json

markwhitfeld:
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[x] CI related changes
[ ] Other... Please describe:
```

Currently, if we build with the Angular 8 compiler and Typescript 3.4 it creates d.ts files that are only compatible with TS3.4 upwards. This breaks compatibility.

## What is the new behavior?
Created a build infrastructure that allows for us to build against various versions of angular (not just the one specified in the root package.json).
The default build is set to use Angular 7 so that it creates backwards compatible definitions.

In future we could use these multiple builds to generate multiple variants of the `.d.ts` files for the different versions of typescript.

```
[ ] Yes
[x] No
```


tools/set-metadata.ts
```diff
@@ -18,7 +27,9 @@ async function main() {
     // set all the packages peerDependencies to be the same as root package.json version
     for (const p of packages) {
```
splincode: can immediately rename p to package?

markwhitfeld: `package` is a reserved word that is not valid in strict mode. I could call it something else, but this isn't code that I deliberately changed in this PR so I am not going to worry about this.
Prettier got excited when I opened the file and reformatted the lines! 🤕

splincode: rename to packageName?)

build/package.json
```diff
@@ -0,0 +1,12 @@
+{
+  "name": "build",
+  "version": "1.0.0",
+  "description": "",
+  "scripts": {
+    "build": "yarn build:ng7",
```
splincode: What is the advantage now to generate for Angular 7? How we detect that's correct build for Angular 8 in CI?

tools/set-metadata.ts
markwhitfeld: Ok, I have renamed it. But this is unrelated to my PR!
As I said, prettier made the line format change.

markwhitfeld: `packageInfo` reflects a better name for this btw

build/package.json
markwhitfeld: The fact is that if we gen using Angular 8 then it breaks compatibility of the lib to all previous versions of angular because of a TS 3.4 specific change.
Doing the build with Angular 7 ensures that we maintain compatibility backwards and with Angular 8. Please read the description of the PR for context.

splincode: Ok

markwhitfeld:
Merging this so that I can continue testing against the dev build

PR #1168, opened 2019-07-21
docs: make some tweaks to the recipes and code readability

arturovt:
## PR Checklist

## PR Type
```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: N/A

```
[ ] Yes
[x] No
```


docs/recipes/authentication.md
```diff
@@ -6,10 +6,10 @@ at how we would implement this in NGXS.
 First, let's define our state model and our actions:
 
 ```TS
-export class AuthStateModel {
-  token?: string;
-  username?: string;
-}
+export type AuthStateModel = null | {
```
markwhitfeld: Probably better as an interface instead of a type.

docs/recipes/authentication.md
```diff
@@ -22,38 +22,52 @@ export class Logout {
 ```
 
 In our state model, we want to track our token and the username. The token
-represents a JWT token that was issued for the session.
+represents a JWT token that was issued for the session. If the user is not authorized,
+the state value is `null`, which corresponds to the semantics. It is very easy to work
+with the `null` value and make the appropriate checks.
```
markwhitfeld: My personal preference is to try to stay away from a state being null because it makes selectors more complex.
In this case the `AuthState.token` selector would now need to check for null;
I would have preferred to check that the token is not null for `AuthState.isAuthenticated`.

docs/recipes/authentication.md
arturovt: Nah, because of discriminated union :D

docs/recipes/authentication.md
arturovt: Okay

arturovt: @markwhitfeld 

Should I change `token` property from optional to `null`? Like:

```ts
defaults: {
  token: null,
  username: null
}
```
Except of plain object `{}`

docs/recipes/cache.md
```diff
@@ -7,37 +7,87 @@ using the store's current values and returning them instead of calling the HTTP
...
-      return this.animalService.getZebra(action.id).pipe(
-        map(resp => dispatch(new GetZebraSuccess(resp)))
-      );
+      const zebra = state.zebras[index];
+      return ctx.dispatch(new GetZebraSuccess(zebra));
     }
+
+    return this.animalService
+      .getZebra(action.id)
+      .pipe(mergeMap(zebra => ctx.dispatch(new GetZebraSuccess(zebra))));
```
markwhitfeld: Given the fact that this is a cache pattern surely it makes sense to also store the zebra in the state for next call?

docs/recipes/cache.md
```diff
@@ -7,37 +7,87 @@ using the store's current values and returning them instead of calling the HTTP
...
+      (getZebra)="getZebra($event)"
+    ></app-zebras-list>
+  `
+})
+export class ZebrasComponent {
+
+  @Select(ZooState.getZebras) zebras$: Observable<Zebra[]>;
+
+  zebra$ = this.actions$.pipe(
+    ofActionDispatched(GetZebraSuccess),
```
markwhitfeld: Maybe we should be filtering this by the last requested zebra id to make sure it corresponds to the requested id call?
What do you think?

docs/recipes/authentication.md
```diff
@@ -30,30 +30,43 @@ service.
...
+    username: null
+  }
 })
 export class AuthState {
 
   @Selector()
-  static token(state: AuthStateModel) { 
-    return state.token;
+  static isAuthenticated(state: AuthStateModel): boolean {
+    return state.token !== null;
```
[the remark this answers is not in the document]
markwhitfeld: Regarding the code change, I'm happy with what you have too. Your choice in terms of what is more readable.

arturovt:
@splincode have you got any ideas about some "clean" and "teachable" examples of caching? I understand that it could be anything (like `tickets` state + `ticketsInfo` state with the whole info of any ticket). But I'm just looking for a way that our users can easily understand

arturovt:
We can't use the snapshot. Yes it's easier then the resolver approach.

arturovt:
@splincode do you think it's a good example using `ActivatedRoute.params + switchMap`?

docs/recipes/cache.md
```diff
@@ -8,37 +8,125 @@ service.
...
+  template: `
+    <h1>{{ novel?.title }}</h1>
+    <span>{{ novel?.author }}</span>
+    <p>
+      {{ novel?.content }}
+      <del datetime="{{ novel?.publishedAt }}"></del>
+    </p>
+  `
+})
+export class NovelComponent implements OnDestroy {
```
arturovt: @splincode do you think it's a good example using ActivatedRoute.params + switchMap?

splincode: why not?

Issue #912, opened 2019-03-11
Route selector for the router plugin


arturovt:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[ ] Performance issue
[x] Feature request
[ ] Documentation issue or request
[ ] Support request => https://github.com/ngxs/store/blob/master/CONTRIBUTING.md
[ ] Other... Please describe:
</code></pre>

## Current behavior
```typescript
import { RouterState } from '@ngxs/router-plugin';

@Component({ ... })
export class SomeNestedComponent {
  @Select(RouterState.state) public router$: Observable<RouterStateSnapshot>;

  constructor() {
    this.router$.subscribe((state) => {
      console.log(state.root.firstChild.firstChild.firstChild);
    });
  }
}
```

## Expected behavior
```typescript
import { RouterState } from '@ngxs/router-plugin';

@Component({ ... })
export class SomeNestedComponent {
  @Select(RouterState.route(SomeNestedComponent)) public route$: Observable<ActivatedRoute>;

  constructor() {
    this.route$.subscribe((route) => {
      console.log(route.data.someResolvedData);
    });
  }
}
```

markwhitfeld:
Why wouldn't you just inject the `ActivatedRoute` into your component?

arturovt:
@markwhitfeld 
Sure, what if I would like to "watch" resolved data change? Basic example:
```ts
{
  path: 'cat/:name',
  resolve: { cat: CatResolver },
  runGuardsAndResolvers: 'paramsChange'
}
```
Inside the component:
```ts
public cat$ = this.store.select(RouterState.route(SomeNestedComponent)).pipe(
  map(({ data }) => data.cat)
);
```
Inside the template:
```html
<app-cat-information [cat]="cat$ | async"></app-cat-information>
```

I understand that we shouldn't increase the `@ngxs/router-plugin` functionality. What about a subpackage like `@ngxs/router-plugin/resolve` where we could expose a selector?

NicolasLiveris:
Hi guys !
To make this selection, I use a joining selector.
https://ngxs.gitbook.io/ngxs/concepts/select#joining-selectors
```typescript
  @Selector([RouterState])
  static activityId(state: OptionsState, router: any) {
    return router.state.root.firstChild.firstChild.params.id;
  }
```
It works but it may affect performance.

arturovt:
@NicolasLiveris 
[...]

PR #1169, opened 2019-07-21
feat(router-plugin): implement route snapshot selector

arturovt:
## PR Checklist

## PR Type
```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: #912 

```
[ ] Yes
[x] No
```


arturovt:
Brr, ng 5 integration tests are down because of `import("@angular/router").RouterStateSnapshot` in the `.d.ts` files...

integration/app/list/list.component.html
```diff
@@ -4,5 +4,9 @@
 </div>
 
 <div *ngIf="router$ | async as router">
-  animals were resolved {{ router.root.firstChild.firstChild.data.list }}
+  {{ join('router$', router.root.firstChild.firstChild.data.list) }}
+</div>
+
+<div *ngIf="route$ | async as route">
+  {{ join('route$', route.data.list) }}
```
splincode: Why function instead pipe? Usage function in template is bad practice

arturovt: This is a div just to check that the list was resolved, I don't see any practical benefit to create a pipe specially in this case.

splincode: As you know, but such code in the template makes me confused

arturovt: I'm not sure what exactly makes you confused and I'm also not sure if the method should be a blocker in this case :wink: I'm not using this code within our library, this code is used only for SSR tests, what would be the difference between method and pipe? Would like to hear your thoughts.

markwhitfeld: Maybe a better method name would do, like `getListDisplayText`. I was confused by `join`, thinking that it was `string.join`.

arturovt: @markwhitfeld Okay, NP

integration/tests-ssr/request-service.js
```diff
@@ -4,9 +4,10 @@ module.exports = async value =>
   new Promise((resolve, reject) => {
     request.get(value, { proxy: '' }, (error, response, data) => {
       if (error) {
-       console.log(error);
-        reject(error);
+        console.log(error);
+        return reject(error);
```
markwhitfeld: What does it mean to return something from a promise callback?
If it is just to avoid the else then maybe this would be clearer:
```suggestion
        reject(error);
        return;
```

packages/router-plugin/src/serializer.ts
```diff
@@ -27,7 +27,7 @@ export class DefaultRouterStateSerializer
       fragment: route.fragment,
       data: route.data,
       outlet: route.outlet,
-      component: null,
+      component: route.component,
```
markwhitfeld: This is a problem! We should never add a component to the state because this would cause the component to be made readonly by the dev mode's deepFreeze and result in runtime errors.
Maybe we could have a `componentType` property instead (set to the constructor ref) that is used for comparison.

packages/router-plugin/src/get-route-snapshot.ts
```diff
@@ -0,0 +1,47 @@
...
+  let i = parent.children.length;
+
+  while (i--) {
+    const child = parent.children[i];
+
+    if (child.component === component) {
+      return child;
+    }
+
+    return searchRouteAmongChildren<T>(child, component);
```
markwhitfeld: Both statements return here so it would never check the other children

packages/router-plugin/src/get-route-snapshot.ts
```diff
@@ -0,0 +1,47 @@
...
+  return searchRouteAmongChildren<T>(root, component);
+}
+
+function searchRouteAmongChildren<T>(
+  parent: ActivatedRouteSnapshot,
+  component: Type<T>
+): ActivatedRouteSnapshot | null {
+  let i = parent.children.length;
+
+  while (i--) {
```
markwhitfeld: This code is confusing. I found myself needing to remember the subtle difference between postfix and prefix.
I was also asking myself if this would run for child 0.
Better to move to `i>=0` and do an `i = i-1;` on the last line of the loop.

markwhitfeld: Also, is there any reason why we traverse in reverse order as opposed to starting at 0?
Not worth it if it is just to save a variable declaration IMO.

packages/router-plugin/src/get-route-snapshot.ts
```diff
@@ -0,0 +1,47 @@
+import { Type } from '@angular/core';
+import { ActivatedRouteSnapshot } from '@angular/router';
+import { createSelector } from '@ngxs/store';
+
+import { RouterState, RouterStateModel } from './router.state';
+
+export function getRouteSnapshot<T>(component: Type<T>) {
```
markwhitfeld: We could move this onto `RouterState` to make it more discoverable.

packages/router-plugin/src/serializer.ts
markwhitfeld: Hmmm... could the same type of component appear in multiple places in the RouterState?
If yes, then this is a much bigger problem to solve than just checking the type.

packages/router-plugin/src/get-route-snapshot.ts
markwhitfeld: It could be as `RouterState.getRouteSnapshot` or `RouterState.getSnapshot` or `RouterState.getComponentSnapshot`.

packages/router-plugin/src/serializer.ts
markwhitfeld: Oh wait, I see that route.component has the type `component: Type<any> | string | null`.
This would not be the instance of the component but rather the Type.
Still would not want to freeze a type though. Maybe we could resolve this down to a string in the state if there is s type there?

arturovt:
@splincode @markwhitfeld cc

I'm a little stuck with one question. As we know our selectors do not run if the input doesn't change. So such selector `getRouteSnapshot(Component)` will not be invoked because `Component` doesn't change. Do you have any proposes?

arturovt:
@splincode could you review?

splincode:
@arturovt @markwhitfeld any updates?

markwhitfeld:
See my last comment above

splincode:
@markwhitfeld @arturovt any progress?

splincode:
Why closed?((

Issue #1145, opened 2019-07-03
🐞[BUG]: Handling errors in NGXS (angular) with global error handler


cmaart:
Basically this question here:

https://stackoverflow.com/questions/56872587/handling-errors-in-ngxs-angular-with-global-error-handler

I can't find a way to use a global error handler as fallback only when an error is not handled by the dispatching component. Other good ways of handling errors appreciated.

joaqcid:
I think the way you are supposed to do this is handle the error in the action, if an unhandled exception happens, that's when your fallback global error handler would handle it.

cmaart:
If I'd do that there would be still the problem of the global error handler being called twice when subscribing on the store when calling dispatch. I've written the global error handler to detect duplicates but certainly I don't find this ideal

fikkatra:
For those of you, like me, still running into the issue of errors being processed by the `ErrorHandler` even when they are caught, or errors that are **processed twice** by the `ErrorHandler` when they are not caught, we found a solution to bypass this behaviour.

To summarize:

Desired behaviour:
- errors that aren't caught within the action handler, but are caught later (e.g. by the dispatcher), should not be handled by `ErrorHandler`
- errors that aren't caught anywhere and bubble up, should be handled by `ErrorHandler`, but only once

Observed behaviour:
- errors that aren't caught within the action handler, but are caught later (e.g. by the dispatcher), ARE handled by `ErrorHandler`
- errors that aren't caught anywhere and bubble up, are be handled by `ErrorHandler`, but they are handled TWICE (once because NGXS calls `ErrorHandler`, once because Angular calls `ErrorHandler`)

**Solution:**
Unfortunately, NGXS ALWAYS calls `ErrorHandler` when an error leaves the action handler, there's no way to prevent this (see [this comment](https://github.com/ngxs/store/issues/803#issuecomment-565170651)). The only thing you can do, is customize the logic in `ErrorHandler` so it can detect whether it is called by ngxs (rather than by Angular for uncaught errors), and have a duplicate detection system to ignore errors.

To do this, write an NgxsPlugin that hooks into the ngxs pipe:
```
import { Injectable } from '@angular/core';
import { catchError } from 'rxjs/operators';
import { NgxsPlugin } from '@ngxs/store';
import { ErrorConstants } from './models/error-constants';

@Injectable()
export class NgxsErrorPlugin implements NgxsPlugin {
  public handle(state: any, action: any, next: any): void {
    return next(state, action).pipe(
[...]

Issue #995, opened 2019-04-13
🐞[BUG]: When using @select, If the passed method throws an Exception, the subscription stops and do not emits any values after that.


shivamd20:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ x ] Bug report  
[ ] Performance issue
[ ] Feature request
[ ] Documentation issue or request
[ ] Support request => https://github.com/ngxs/store/blob/master/CONTRIBUTING.md
[ ] Other... Please describe:
</code></pre>

## Current behavior

When using @Select, If the passed method throws an Exception the subscription stops and do not emits any values after that.

## Expected behavior
Selector should ignore the particular  case where the passed method throws an exception and it should print the error message  on browser console.
We should get subsequent updates even when we get error on the selector method.

## Minimal reproduction of the problem with instructions

> Lets say this is our data model
```  
export class AuthStateModel {
    profile: Object;
}
  ```
> and this is our state 

```  

@State<AuthStateModel>({
    name: 'auth',
    defaults: {
        profile: {},
    }
})
export class AuthState {


    constructor(private authService: AuthService) { }

    @Action(ClearAuth)
    clearAuth(ctx: StateContext<AuthStateModel>) {
        ctx.setState({
            profile: undefined
        });
    }

    @Action(SaveAuthData)
    saveAuthData(ctx: StateContext<AuthStateModel>, action: SaveAuthData) {
        const state = ctx.getState();
        const nextState = produce(state, (draftState) => Object.assign(draftState, { ...action.payload }));
        ctx.setState(nextState);
    }

// more actions 
}
```
> Now if we write a selector like this

` @Select(state =>  state.auth.profile.picture )  `
And then if we dispatch two actions,   @Action(ClearAuth) and @Action(SaveAuthData), We will not get the data  from second action because the first action will make the profile object undefined and when ngxs calls select's passed method, It will throw ` cannot read property length of undefined ` exception.

## Environment

<pre><code>
Libs:
- @angular/core version: 7.0.4
- @ngxs/store version: 3.3.2

Browser:
- [x] Chrome (desktop) version XX
- [ ] Chrome (Android) version XX
- [ ] Chrome (iOS) version XX
- [ ] Firefox version XX
- [ ] Safari (desktop) version XX
- [ ] Safari (iOS) version XX
- [ ] IE version XX
- [ ] Edge version XX

</code></pre>

splincode:
[...]

PR #1170, opened 2019-07-21
feat(store): add selection global strategy

splincode:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: https://github.com/ngxs/store/issues/995, https://github.com/ngxs/store/issues/1145, https://github.com/ngxs/store/issues/541

## What is the new behavior?

Support override the standard NGXS behavior for flexible state management.

```
[ ] Yes
[x] No
```


docs/SUMMARY.md
```diff
@@ -32,6 +32,7 @@
   * [Life-cycle](advanced/life-cycle.md)
   * [Meta Reducers](advanced/meta-reducer.md)
   * [Shared State](advanced/shared-state.md)
+  * [Selection strategy](advanced/shared-state.md)
```
arturovt: `selection-strategy` I guess :D

splincode: Good catch :)

markwhitfeld:
I'm battling to understand the use case that this feature is trying to address?
Please explain.

arturovt:
@splincode You have done a big and great job in that PR! I haven't made a review as I support Mark with his question, what problem does it solve?

splincode:
@markwhitfeld @arturovt 

When using `@select` or `this.store.select`, If the passed method throws an Exception the subscription stops and do not emits any values after that.

If we try to do this:

```ts
@State({
  name: "manager",
  defaults: {
    selectedId: null,
    entities: {
        1: { .. },
        2: { .. } 
    }
  }
})
class MyState {
   public static selectedEntity(): MyEntity {
      return state!.entities[state!.selectedId];
   }

   @Action(SetIdAction)
   public setId({ setState }, { id }: SetIdAction) {
     setState((state) => ({ ...state, selectedId: id });
   }

}
```

```ts
class AppComponent {
  
  constuctor(store: Store) {}
 
  public ngOnInit() {
      this.store.select(MyState.selectedEntity).subscribe((entity) => {
         console.log('entity', entity):
     });
  }

}
```

But if `selectedId` is Null, then we can get unexpected consequences, our subscription will be automatically unsubscribed.  In normal situations, I wanted to do so:

```ts
class AppComponent {
  
  constuctor(store: Store) {}
 
  public ngOnInit() {
      this.store.select(MyState.selectedEntity).pipe(retryWhen((errors) => errors.pipe(
         tap((err) => {
            console.log('check error null pointer exception', err);
            // retry subscription 
        })
     ))).subscribe((entity) => {
         console.log('entity', entity):
     });
  }

}
```

But custom pipe with retryWhen doesn't work. Why? Because we use catchError in `return this._stateStream.pipe( catchError ( 'OUR LOGIC' ) )` in Store. 

Thus, we can no longer use user our own RxJS operators (catchError, retryWhen, retry) to work with a subscription after subscribe.

packages/store/src/selection/selection-global-strategy.ts
```diff
@@ -0,0 +1,36 @@
...
+
+  /**
+   * The default strategy defines the standard behavior of the NGXS:
+   * unsubscribe at the first error in the stream.
+   * @param errors
+   */
+  public retryWhenHandler(errors: Observable<Error | any>): Observable<Error | any> {
+    return errors.pipe(
+      tap(error => {
+        throw error;
```
arturovt: P.S. - I'm not a fan of using `tap` in that place. To avoid any such effects it's much better to map to the error stream:
```ts
return errors.pipe(
  mergeMap(error => throwError(error))
);
```
Using `throwError` - we primarily insure our users.
```

File: docs/advanced/selection-strategy.md [hunk 3]
```diff
@@ -0,0 +1,172 @@
...
+...
+
+@Injectable()
+class CustomSelectionStrategy extends SelectionGlobalStrategy {
+ private readonly delay: number = 500;
+ private readonly maxRetry: number = 2;
+
+ public retryWhenHandler(errors: Observable<Error>): Observable<number> {
+   return errors.pipe(
+     scan((count: number, err: Error): number => {
```
arturovt: `scan` definitely shouldn't be in our docs, let's think of such simpler example, I understand that you understand this code :D but our users might not...

packages/store/src/selection/selection-global-strategy.ts
```diff
@@ -0,0 +1,36 @@
+import { Observable, of, throwError } from 'rxjs';
+import { Injectable } from '@angular/core';
+import { tap } from 'rxjs/operators';
+
+import { NgxsConfig } from '../symbols';
+import { SelectorStrategy } from './symbols';
+
+@Injectable()
+export class SelectionGlobalStrategy implements SelectorStrategy {
```
arturovt: should be `GlobalSelectionStrategy`

packages/store/src/selection/symbols.ts
```diff
@@ -0,0 +1,6 @@
+import { Observable } from 'rxjs';
+
+export interface SelectorStrategy {
```
arturovt: should be `SelectionStrategy` ("Selector" is not semantic)

packages/store/tests/select.spec.ts
```diff
@@ -325,24 +329,78 @@ describe('Select', () => {
       }
     }
 
+    @Injectable()
+    class CustomSelectionStrategy extends SelectionGlobalStrategy {
```
arturovt: should it extend? maybe `implements SelectionStrategy`? as you don't use `super` keyword here

packages/store/src/selection/selection-global-strategy.ts
splincode: Good

packages/store/tests/select.spec.ts
splincode: the problem is that then you need to implement all the methods in the service

markwhitfeld:
I agree with @splincode, you have done an amazing job on this PR. I almost want to stick it on the wall and frame it!

Unfortunately I think that it would have been better to discuss it as a team before putting all this effort in because you have clearly invested a lot of time in this...

I still have questions:
- Where exactly should we be receiving selector errors?
  - In the code that selects?
  - or in the code that does the state change that causes the error?
  - I think that it would be in the code that causes the change (if `supressErrors` is false)
  - If `supressErrors` is true then there is no possibility of the error closing the subscription.
- I think that the observable returned by `select` etc. should be resilient to errors anyway. If you want to end the subscription early then use `takeWhile` or similar.
- The retry logic is very edge case and I don't think a problem that needs to be addressed if the selectors are resilient.

Issue #1108, opened 2019-05-30
@Selector() not working when inheriting from a parent @State


joaqcid:
## I'm submitting a...
<pre><code>[ X ] Bug report</code></pre>

## Current behavior
When inheriting from a Parent @State class with a @Selector, using that selector on child @State class doesnt return the store value

## Expected behavior
Child @State should return the value of the selected value via the @Selector method in its parent class

## Minimal reproduction of the problem with instructions

https://stackblitz.com/edit/ngxs-inherit-selector-issue-1108

* Step 1
* Step 2
* Step 3

## What is the motivation / use case for changing the behavior?
N/A

## Environment

<pre><code>
Libs:
N/A

Browser:
N/A
 
For Tooling issues:
N/A

Others:
N/A
</code></pre>

joaqcid:
One thing i noticed is that if i add the parent `@State` to the `AppModule` it works, but the behavior is not the desired. In this case both parent and child state have the same state.
I wonder if i what i want to do is possible or not…

arturovt:
I'm sorry, maybe I'm missing something, but what behavior do you expect if you don't add the `CountState` when calling `forRoot`? We can't manually add this state to the root DI providers.

joaqcid:
The behaviour expected is to able to inherit selectors from base abstract classes, to avoid duplicated selectors, for instance for "loading" @selector

splincode:
Yes, I noticed this problem and it needs to be solved

arturovt:
@joaqcid 
You use a selector from the non-imported class:
```ts
@Select(CountState.count) count$: Observable<number>;
```

joaqcid:
yeah, i think i was testing with both states, just updated it, you can see it doesnt work, even if use `ChildCountState.count`

arturovt:
You could just write `ChildCountState` except of `CountState` cause I got confused haha :joy:

MuhammadRizwanBashir:
@arturovt  Hi,
Is it going to be resolved? If yes then please mark it as bug or feature as you see it. So we can follow this.
Thanks

joaqcid:
Hi!

After reading more the documentation, found i can actually do this with dynamic selectors

```
export abstract class CountState {
  static get count() {
    return createSelector([this], (state: CountStateModel) => {
      return state.count
    })
  }
}
```

Updated the stackblitz with working code

splincode:
You can send us a pull request for updating documentation.

splincode:
[...]

PR #1171, opened 2019-07-21
docs: inherit selectors from base class

joaqcid:
## PR Checklist

## PR Type
```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ X ] Documentation content changes
[ ] Other... Please describe:
```

## Other information

addresses https://github.com/ngxs/store/issues/1108


markwhitfeld:
@joaqcid This is great! Please just rebase on master to resolve your failing tests issue. Unfortunately you branched when there was an issue on master.

docs/concepts/select.md
```diff
@@ -272,6 +272,61 @@ export class ZooComponent {
 }
 ```
 
+_Inheriting selectors_
```
splincode: ### Inheriting selectors

joaqcid:
sure thing

docs/concepts/select.md
```diff
@@ -353,6 +353,61 @@ export class CityService {
...
 
+### Inheriting selectors  
+
+Another way to share selectors across states with similar structure could be to create dynamic selectors in a base class. Using the same previous example we could:
+
+```TS
+export class EntitiesState {
+
+  static entities() {
+    return createSelector([this], (state: { entities: any[] }) => {
```
arturovt: should be it be `state: string[]`? because below states has the `string[]` model :D

joaqcid: right, it was after the dynamics selectors section before

joaqcid:
@markwhitfeld @arturovt @splincode what do you think now?

docs/concepts/select.md
```diff
@@ -353,6 +353,61 @@ export class CityService {
...
+
+And extend the `EntitiesState` class on each `@State` like this:
+
+```TS
+
+@State<string[]>({
+  name: 'animals',
+  defaults: []
+})
+export class ZooState extends EntitiesState {
```
markwhitfeld: This would need to be a state that has an `{ entities: any[] }` model type.
I would recommend that we don't redefine what a ZooState is for consistency.
Maybe we could have UsersState and RolesState as entity states?

docs/concepts/select.md
```diff
@@ -353,6 +353,61 @@ export class CityService {
...
 now we can use this `zooThemeParks` selector anywhere in our application.
 
+### Inheriting selectors  
+
+When we have states that share similar structure, we can extract the shared selectors into a base class which we can later extend from. If we have an `entities` field on multiple states, we can create a base class containing a dynamic `@Selector()` for that field, and extend from it on the `@State` classes like this.
+
+```TS
+export class EntitiesState {
+
+  static entities() {
```
markwhitfeld: I may be wrong so please check, but I think that we could make this a getter and then we won't need to invoke the method to get the selector in the component... ?
```suggestion
  static get entities() {
```

joaqcid: yup, i actually have it this way on my code :), updated

docs/concepts/select.md
joaqcid: ok, i finally went with UsersState and ProductsState, i also added the interfaces for UsersStateModel and ProductsStateModel and update the `name` in the `@State` options

docs/concepts/select.md
splincode: Maybe we can use generic here?

For example
```export class EntitiesState<T>```

File: docs/concepts/select.md [hunk 5]
```diff
@@ -375,19 +375,27 @@ And extend the `EntitiesState` class on each `@State` like this:
...
 }
 
-@State<string[]>({
-  name: 'animals',
+export interface ProductsStateModel {
+  entities: Product[]
+}
+
+@State<ProductsStateModel[]>({
+  name: 'roles',
```
markwhitfeld: ProductsState stored on 'roles' path? ;-)

docs/concepts/select.md
```diff
@@ -353,6 +353,69 @@ export class CityService {
 
 now we can use this `zooThemeParks` selector anywhere in our application.
 
+### Inheriting selectors  
+
+When we have states that share similar structure, we can extract the shared selectors into a base class which we can later extend from. If we have an `entities` field on multiple states, we can create a base class containing a dynamic `@Selector()` for that field, and extend from it on the `@State` classes like this.
+
+```TS
+export class EntitiesState {
```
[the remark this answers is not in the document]
joaqcid: It doesnt allow me to use generics on static members

image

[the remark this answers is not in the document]
arturovt: Yeah it would require to declare `entities` not as property, but as a method:
```ts
static entities<T>() { ... }
```

So we could use it like:
```ts
@Select(State.entities) --> @Select(State.entities<Model>())
```
@splincode @markwhitfeld cc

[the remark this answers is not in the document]
splincode: Good point 

```ts
static entities<T>() { ... }
```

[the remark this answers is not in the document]
joaqcid: I prefer typing it on the Component, since the @Select doesn't provide the Observable typing by default. But if you guys want i can update it.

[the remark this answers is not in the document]
joaqcid: what about this 

```
export interface EntitiesStateModel<T> {
  entities: T[]
}

export class EntitiesState {

  static entities<T>() {
    return createSelector([this], (state: EntitiesStateModel<T>) => {
      return state.entities as T[];
    });
  }

}

export interface User {
  name: string
}

export interface UsersStateModel extends EntitiesStateModel<User> {

}
```

joaqcid:
or as @arturovt suggested

```
export class EntitiesState {

  static entities<T>() {
    return createSelector([this], (state: { entities: T[] })  => {
      return state.entities;
    });
  }
}

export interface UsersStateModel {
   entities: User[]
}

@State<UsersStateModel>({
  name: 'users'
  //...
})
export class UsersState extends EntitiesState {
  //...
}


this.store.select(UsersState.entities<User>())
```

docs/concepts/select.md
```diff
@@ -353,6 +353,69 @@ export class CityService {
...
+
+```TS
+
+export interface UsersStateModel {
+  entities: User[]
+}
+
+@State<UsersStateModel>({
+  name: 'users',
+  defaults: []
```
arturovt: Should it be
```ts
{
  entities: []
}
```
?

docs/concepts/select.md
```diff
@@ -353,6 +353,69 @@ export class CityService {
...
+  //...
+}
+
+export interface ProductsStateModel {
+  entities: Product[]
+}
+
+@State<ProductsStateModel>({
+  name: 'products',
+  defaults: []
```
arturovt: Should it be
```ts
{
  entities: []
}
```
?

docs/concepts/select.md
```diff
@@ -353,6 +353,69 @@ export class CityService {
...
 
+### Inheriting selectors  
+
+When we have states that share similar structure, we can extract the shared selectors into a base class which we can later extend from. If we have an `entities` field on multiple states, we can create a base class containing a dynamic `@Selector()` for that field, and extend from it on the `@State` classes like this.
+
+```TS
+export class EntitiesState {
+
+  static get entities() {
+    return createSelector([this], (state: { entities: any[] }) => {
```
arturovt: Will you do it generic? As we discussed previously `static entities<T>() { ... }`

docs/concepts/select.md
```diff
@@ -353,6 +353,69 @@ export class CityService {
...
+@State<ProductsStateModel>({
+  name: 'products',
+  defaults: []
+})
+export class ProductsState extends EntitiesState {  
+  //...
+}
+```
+
+Then you, can use them
```
arturovt: `Then you could use them as follows:`

docs/concepts/select.md
```diff
@@ -353,6 +353,81 @@ export class CityService {
...
+  @Select(UsersState.entities<User>())
+  users$: Observable<User[]>;
+
+  @Select(ProductsState.entities<Product>())
+  products$: Observable<Product[]>;
+
+}
+```
+
+or
```
arturovt: `Or:`

Just add a colon and I'll merge :tada:

docs/concepts/select.md
```diff
@@ -353,6 +353,81 @@ export class CityService {
...
+  defaults: {
+    entities: []
+  }
+})
+export class UsersState extends EntitiesState {  
+  //...
+}
+
+export interface ProductsStateModel {
+  entities: Product[]
```
arturovt: `Product[];`

docs/concepts/select.md
```diff
@@ -353,6 +353,81 @@ export class CityService {
...
+
+}
+```
+
+And extend the `EntitiesState` class on each `@State` like this:
+
+```TS
+
+export interface UsersStateModel {
+  entities: User[]
```
arturovt: `User[];`

arturovt:
@joaqcid Looks amazing, thank you!

PR #1205, opened 2019-08-03
fix(hmr-plugin): remove @angularclass/hmr peer dependency

splincode:
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

image

We depend on a package that has not been updated for 2 years, while using only 2 functions from there.

## What is the new behavior?

Now we use our own functionality and it works

```
[ ] Yes
[x] No
```


splincode:
@markwhitfeld @arturovt ping

packages/store/internals/src/index.ts
```diff
@@ -3,4 +3,4 @@ export { NgxsBootstrapper } from './ngxs-bootstrapper';
 export { memoize } from './memoize';
 export { ObjectUtils } from './object-utils';
 export { INITIAL_STATE_TOKEN, InitialState } from './initial-state';
-export { ObjectKeyMap, StateClass } from './symbols';
+export { ObjectKeyMap, StateClass, Callback } from './symbols';
```
markwhitfeld: It is really not worth exposing a generic utility type like this.
Although this is an 'internal' public API it should only be used when there is no other option than to expose the internals.
Let's keep it internal and private.

splincode: done

packages/store/src/store.ts
```diff
@@ -41,19 +41,17 @@ export class Store {
     const selectorFn = getSelectorFn(selector);
     return this._stateStream.pipe(
       map(selectorFn),
-      catchError(
-        (err: Error): Observable<never> | Observable<undefined> => {
-          // if error is TypeError we swallow it to prevent usual errors with property access
-          const { suppressErrors } = this._config.selectorOptions;
+      catchError((err: Error): Observable<never> | Observable<undefined> => {
```
markwhitfeld: Try to keep changes unrelated to the PR out of the commits if possible, it just makes reviewing more difficult to see these changes and to try to figure out if anything has actually changed here.
These format changes, while valid, still cause a bit of confusion as to how this relates to the PR.

packages/store/src/host-environment/host-environment.ts
```diff
@@ -1,6 +1,5 @@
 import { Inject, Injectable } from '@angular/core';
-import { NG_DEV_MODE, NG_TEST_MODE } from '../symbols';
-import { Callback } from '../internal/internals';
+import { Callback, NG_DEV_MODE, NG_TEST_MODE } from '../symbols';
```
markwhitfeld: What is the reason for moving this from `internals.ts` to `symbols.ts`?

packages/hmr-plugin/src/hmr-manager.ts
```diff
@@ -1,13 +1,13 @@
...
-import { createNewHosts, removeNgStyles } from '@angularclass/hmr';
 
 import { HmrStorage } from './internal/hmr-storage';
 import {
   BootstrapModuleFn,
   NgxsHmrLifeCycle,
   NgxsHmrOptions,
-  NgxsHmrSnapshot
+  NgxsHmrSnapshot,
+  RemovableOldHostFn
```
markwhitfeld: Maybe a better name could be `OldHostRemoverFn` because this describes what it does. What do you think?

packages/hmr-plugin/src/tests/hmr-mock.ts
```diff
@@ -80,6 +80,7 @@ export class AppMockModuleNoHmrLifeCycle {
   }
 }
 
+// noinspection DuplicatedCode
```
markwhitfeld: What is this flag for?

packages/hmr-plugin/src/hmr-manager.ts
```diff
@@ -53,20 +61,54 @@ export class HmrManager<T extends Partial<NgxsHmrLifeCycle<S>>, S = NgxsHmrSnaps
...
+      };
+    });
+
+    return function removeOldHosts(): void {
+      removableList.forEach((removeOldHost: RemovableOldHostFn) => removeOldHost());
+    };
+  }
+
+  private removeNgStyles(): void {
+    const head: HTMLHeadElement | null = document.head;
```
arturovt: `const head: HTMLHeadElement = document.head!`

packages/store/src/store.ts
splincode: done

packages/hmr-plugin/src/hmr-manager.ts
splincode: done

packages/hmr-plugin/src/hmr-manager.ts
splincode: done

packages/hmr-plugin/src/tests/hmr-mock.ts
splincode: done

packages/hmr-plugin/src/symbols.ts
```diff
@@ -37,17 +51,18 @@ export interface NgxsHmrOptions {
    * (default: 100ms)
    */
   deferTime?: number;
+
+  /**
+   * @description
+   * Add a handler which is executed when the current module code is replaced.
+   * (default: null)
+   */
+  dispose?: DisposeHook;
```
markwhitfeld: This was a fix PR, and now it has grown into a feature PR with additional options and exported API.
Please scale this PR back to just be the fix and add these new features as a separate PR... unless these are essential for the fix?

splincode: revert

packages/store/src/host-environment/host-environment.ts
splincode: done

integration/main.browser.ts
```diff
@@ -1,23 +1,20 @@
...
-const bootstrap = () => platformBrowserDynamic().bootstrapModule(AppBrowserModule);
+const bootstrap: Bootstrap = () => platformBrowserDynamic().bootstrapModule(AppBrowserModule);
 
 if (environment.hmr) {
-  if (module['hot']) {
-    hmr(module, bootstrap).catch(err => console.error(err));
-  } else {
-    console.error('HMR is not enabled for webpack-dev-server!');
-    console.log('Are you using the --hmr flag for ng serve?');
-  }
```
markwhitfeld: What is the reason for removing the example check code above?

packages/hmr-plugin/src/hmr-bootstrap.ts
```diff
@@ -15,9 +16,6 @@ export async function hmr<T>(
 
   webpackModule.hot.accept();
 
-  interface HmrDataTransfer {
-    snapshot?: any;
-  }
```
markwhitfeld: This is only used within this method as far as I can see, so we should not move it to the symbols file.
It should stay here as internal implementation detail.
Unless I am missing something...

packages/hmr-plugin/src/hmr-manager.ts
```diff
@@ -1,13 +1,13 @@
...
-import { createNewHosts, removeNgStyles } from '@angularclass/hmr';
 
 import { HmrStorage } from './internal/hmr-storage';
 import {
   BootstrapModuleFn,
   NgxsHmrLifeCycle,
   NgxsHmrOptions,
-  NgxsHmrSnapshot
+  NgxsHmrSnapshot,
+  OldHostRemoverFn
```
markwhitfeld: The `OldHostRemoverFn` type should rather be defined in this file (and not exported) because its' only usages are within this file. We should keep types and interfaces close to their usage until they are needed in multiple places.

packages/hmr-plugin/src/symbols.ts
```diff
@@ -23,6 +23,14 @@ export interface NgxsHmrLifeCycle<T = NgxsHmrSnapshot> {
 export type HmrCallback<T> = (ctx: StateContext<T>, state: Partial<T>) => void;
 export type BootstrapModuleFn<T = any> = () => Promise<NgModuleRef<T>>;
 
+export interface WebpackHotApi<T = any> {
```
markwhitfeld: There is no need to separate this interface from the `WebpackHotModule` interface or export it.
The outside world will only ever need to interact with the `WebpackHotModule` interface.

integration/main.browser.ts
splincode: https://github.com/ngxs/store/blob/c11d5ffd385d9baf99aeb709070e026e148f5aa0/packages/hmr-plugin/src/hmr-bootstrap.ts#L12

Because it is not necessary, this check is done inside the plugin

packages/hmr-plugin/src/hmr-manager.ts
splincode: It is very unfortunate that you take this approach in your applications, in normal languages such as Java, all interfaces should be taken out in a separate file, since in any case you do not make them public for other packages.

But anyway, I will make it do as you want.

splincode: done

splincode:
@markwhitfeld @arturovt done

packages/hmr-plugin/src/hmr-bootstrap.ts
splincode: done

arturovt:
Do you have to remote this package from the root `package.json`?

packages/hmr-plugin/src/hmr-manager.ts
```diff
@@ -53,20 +62,54 @@ export class HmrManager<T extends Partial<NgxsHmrLifeCycle<S>>, S = NgxsHmrSnaps
...
+      this.optionsBuilder
+    );
+  }
+
+  private cloneHostsBeforeDestroy(): () => void {
+    const elements: Element[] = this.applicationRef.components.map(
+      (component: ComponentRef<Element>) => component.location.nativeElement
+    );
+
+    const removableList: OldHostRemoverFn[] = elements.map((componentNode: Element) => {
```
arturovt: Do we really need `OldHostRemoverFn` type? In theory TS compiler infers type from the return value itself

markwhitfeld: Although I agree that it is not entirely necessary, it carries value in terms of self documenting code to communicate the intent of the construct. I think it should remain.

packages/hmr-plugin/src/hmr-manager.ts
markwhitfeld: I really do appreciate you adjusting your typical approach. I come from over a decade of C# and those typical statically typed approaches. There is a paradigm shift in the way that types and interfaces are used in TypeScript. In this case this type is used as an internal clarification of functionality, not an extensibility point for polymorphism.

arturovt:
@markwhitfeld rebases after Renovate's updates

markwhitfeld:
@arturovt I was referring to the changes in @splincode's 'fix' commits. Was this just yarn being eager?

arturovt:
I guess it were merge conflicts :)

PR #1208, opened 2019-08-04
refactor(store): add unit and types tests for select decorator

splincode:
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

## What is the new behavior?

Refactor `@Select` decorator with improved test coverage

image


packages/store/src/decorators/select/symbols.ts
```diff
@@ -0,0 +1,31 @@
...
+  if (typeof selectorOrFeature === 'string') {
+    const propsArray: string[] = paths.length
+      ? [selectorOrFeature, ...paths]
+      : selectorOrFeature.split('.');
+
+    return propGetter(propsArray, config);
+  } else if (selectorOrFeature[META_KEY] && selectorOrFeature[META_KEY].path) {
+    return propGetter(selectorOrFeature[META_KEY].path.split('.'), config);
+  } else {
+    return selectorOrFeature;
```
arturovt: `return` without `else`?

packages/store/tests/select.spec.ts
```diff
@@ -5,12 +5,28 @@ import { first, last } from 'rxjs/operators';
...
+import { Select } from '../src/decorators/select/select';
 import { Selector } from '../src/decorators/selector';
 import { State } from '../src/decorators/state';
 import { Action } from '../src/decorators/action';
-import { StateContext } from '../src/symbols';
+import { SELECT_META_KEY, StateContext } from '../src/symbols';
 import { removeDollarAtTheEnd } from '../src/internal/internals';
+import { CONFIG_MESSAGES, VALIDATION_CODE } from '../src/configs/messages.config';
+
+function FrozenClass(target: Function): void {
```
arturovt: `Freeze`

packages/store/tests/select.spec.ts
```diff
@@ -183,6 +263,47 @@ describe('Select', () => {
     });
   }));
 
+  it('should be throw when use reserved select property in class', async(() => {
```
arturovt: `should throw an exception if the select symbol is reserved`
P.S. `async` is not needed here

packages/store/tests/select.spec.ts
```diff
@@ -68,6 +84,70 @@ describe('Select', () => {
...
+      });
+
+      const comp = TestBed.createComponent(MySelectComponent);
+      comp.componentInstance.state.subscribe();
+    } catch (e) {
+      expect(e.message).toEqual(CONFIG_MESSAGES[VALIDATION_CODE.SELECT_NOT_CONNECTED]());
+    }
+  });
+
+  it('should be throw error when field is frozen', () => {
```
arturovt: `should throw an exception when the field is frozen`

packages/store/tests/select.spec.ts
```diff
@@ -68,6 +84,70 @@ describe('Select', () => {
...
+
+      new SelectComponent().state.subscribe();
+    } catch (e) {
+      expect(e.message).toEqual(
+        CONFIG_MESSAGES[VALIDATION_CODE.SELECT_FACTORY_NOT_CONNECTED]()
+      );
+    }
+  });
+
+  it('should be throw error when component is frozen', () => {
```
arturovt: `should throw an exception when the component class is frozen`

packages/store/tests/select.spec.ts
```diff
@@ -68,6 +84,70 @@ describe('Select', () => {
 
   const states = [MySubState, MySubSubState, MyState];
 
+  it('should be throw error when forgot to import the ngxs module', () => {
```
arturovt: `should throw an exception when the user has forgotten to import the NGXS module`

packages/store/src/decorators/select/symbols.ts
```diff
@@ -0,0 +1,31 @@
+import { Observable } from 'rxjs';
+
+import { SelectFactory } from './select-factory';
+import { META_KEY, NgxsConfig } from '../../symbols';
+import { propGetter } from '../../internal/internals';
+import { CONFIG_MESSAGES, VALIDATION_CODE } from '../../configs/messages.config';
+
+export function createSelect(fn: any): Observable<any> {
```
arturovt: this function should be in the `select.ts` file as it's used only there

packages/store/src/decorators/select/select.ts
```diff
@@ -0,0 +1,50 @@
...
+    Object.defineProperty(target, selectorId, {
+      writable: true,
+      enumerable: false,
+      configurable: true
+    });
+
+    Object.defineProperty(target, name, {
+      enumerable: true,
+      configurable: true,
+      get: function() {
```
arturovt:
```ts
Object.defineProperty(target, name, {
  enumerable: true,
  configurable: true,
  get() {
    if (Object.isFrozen(target)) {
      throw new Error(CONFIG_MESSAGES[VALIDATION_CODE.SELECT_NOT_CONNECTED]());
    }

    return (
      this[selectorId] ||
      (this[selectorId] = createSelect.apply(this, [
        createSelector(
          selectorOrFeature,
          paths
        )
      ]))
    );
  }
});
```

splincode: what is prefer 

```ts
 this[selectorId] ||
      (this[selectorId] = ...)
```

instead

```ts
let stream: Observable<any> | null = this[selectorId];

if (!stream) {
   stream = this[selectorId] = createSelect.apply(this, [
        createSelector(
           selectorOrFeature,
             paths
         )
   ]);
}

return stream;
```

I think this is more readable

arturovt: less code = win, also the problem is that `return this[selectorId] || ...` will always return the cached value first, while `if (!stream)` condition is not necessary to be called every time. Thoughts?

splincode: Maybe then it’s worth writing a comment in the code then why do we do this

arturovt: Do we write comments for ourselves? I guess we boss understand this code :DD

splincode: But other people can read our code.

arturovt: There is a concept of "self-documenting code". I agree with documenting dark parts of any code, too much comments is also bad, sounds doubtful. So this code:

```ts
this[selectorId] ||
      (this[selectorId] = ...)
```

is self-documenting IMO, but it's up to you.

packages/store/src/decorators/select/symbols.ts
splincode: done

packages/store/tests/select.spec.ts
splincode: done

packages/store/tests/select.spec.ts
splincode: done

packages/store/tests/select.spec.ts
splincode: done

packages/store/tests/select.spec.ts
splincode: done

packages/store/src/decorators/select/symbols.ts
splincode: done

packages/store/src/decorators/select/select.ts
splincode: done

packages/store/tests/select.spec.ts
splincode: done

packages/store/src/decorators/select/select-factory.ts
```diff
@@ -1,20 +1,57 @@
...
 
   constructor(store: Store, config: NgxsConfig) {
     SelectFactory.store = store;
     SelectFactory.config = config;
   }
+
+  /**
+   * @example If `foo$` => make it just `foo`
+   */
+  public static removeDollarAtTheEnd(name: string): string {
```
arturovt: This doesn't improve code readability. I'm not a fan of this "Java" approach with classes and static members, and it definitely shouldn't be in our library as we're developing in TypeScript! Functions are first-class citizens in JavaScript.

There should be a clear purpose of moving `removeDollarAtTheEnd` from the `function removeDollarAtTheEnd` way to the `static removeDollarAtTheEnd`. What's the benefit?

`SelectFactory` is a service locator that should contain references to the `Store` and `NgxsConfig` instances, but it shouldn't contain internal API as static members.

Please, revert it back :pray:

splincode: I’ll think about how this can be improved, I hate functions in JS, and all my projects are saturated with classes, since I am a backend developer in Kotlin and Java.

arturovt: @splincode 

Dude, you're very good at development, but let's stick with the JS approach as we're developing in JS :heart_decoration:

splincode: I understand, I will fix it

markwhitfeld:
I'll review once these changes are made

packages/store/src/decorators/select/select-factory.ts
splincode: done

splincode:
@markwhitfeld could you see?

packages/store/src/configs/messages.config.ts
```diff
@@ -31,5 +33,8 @@ export const CONFIG_MESSAGES: ObjectKeyMap<Function> = {
   [VALIDATION_CODE.SELECT_NOT_CONNECTED]: () =>
     `You can't use @Select your instance is frozen`,
   [VALIDATION_CODE.SELECT_CLASS_NOT_EXTENSIBLE]: () =>
-    `Your class is not extensible for @Select`
+    `Your class is not extensible for @Select`,
```
arturovt: `@Select() decorator cannot be used with non-extensible classes`

packages/store/src/decorators/select/symbols.ts
```diff
@@ -0,0 +1,45 @@
...
+import { propGetter } from '../../internal/internals';
+import { SelectFactory } from './select-factory';
+import { META_KEY } from '../../symbols';
+
+const DOLLAR_CHAR_CODE = 36;
+
+/**
+ * The SelectUtils namespace is used for logical grouping of functionalities.
+ */
+export namespace SelectUtils {
```
markwhitfeld: Why is this nested in a namespace? Feels like it is trying to be `java` class.
We should just export the functions. This is JavaScript. Functions are first-class citizens.
If there is a good motivation I'm happy to hear it. 👍

packages/store/src/decorators/select/symbols.ts
```diff
@@ -0,0 +1,45 @@
...
+  export function selectBySelector<T = any>(selector: any): Observable<T> {
```
markwhitfeld: possibly better name?
`createSelectObservable`

packages/store/src/decorators/select/symbols.ts
```diff
@@ -0,0 +1,45 @@
...
+    if (!SelectFactory.store) {
+      throw new Error(CONFIG_MESSAGES[VALIDATION_CODE.SELECT_FACTORY_NOT_CONNECTED]());
+    }
+
+    return SelectFactory.store.select(selector);
+  }
+
+  export function unwrapSelector(name: string, rawSelector?: any, paths: string[] = []): any {
```
markwhitfeld: Maybe this could have a better name?
possibly `createSelectorFn`

packages/store/src/configs/messages.config.ts
splincode: done

packages/store/src/decorators/select/select.ts
```diff
@@ -0,0 +1,41 @@
...
+ * Decorator for selecting a slice of state from the store.
+ */
+export function Select(rawSelector?: any, ...paths: string[]): PropertyDecorator {
+  return function(target: any, key: string | symbol): void {
+    if (!Object.isExtensible(target)) {
+      throw new Error(CONFIG_MESSAGES[VALIDATION_CODE.SELECT_CLASS_NOT_EXTENSIBLE]());
+    }
+
+    const name: string = key.toString();
+    const selectorId: unique symbol = Symbol(`${SELECT_META_KEY}__${name}`);
```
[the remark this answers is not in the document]
splincode: reverted

packages/store/tests/select.spec.ts
```diff
@@ -68,9 +84,73 @@ describe('Select', () => {
...
+      new SelectComponent().state.subscribe();
+    } catch (e) {
+      expect(e.message).toEqual(
+        CONFIG_MESSAGES[VALIDATION_CODE.SELECT_FACTORY_NOT_CONNECTED]()
+      );
+    }
+  });
+
+  it('should throw an exception when the component class is frozen', () => {
+    try {
```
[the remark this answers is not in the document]
splincode: done

packages/store/src/decorators/select/symbols.ts
splincode: done

packages/store/src/decorators/select/symbols.ts
splincode: done

packages/store/tests/select.spec.ts
```diff
@@ -1,16 +1,32 @@
...
+
+function SealField(target: any, name: string | symbol) {
+  Object.defineProperty(target, name, {
+    enumerable: false,
+    writable: false,
+    configurable: false
+  });
+
+  Object.seal(target);
+}
```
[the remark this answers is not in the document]
splincode: done

packages/store/src/decorators/select/symbols.ts
splincode: revert

packages/store/src/utils/store-validators.ts
```diff
@@ -47,4 +48,16 @@ export abstract class StoreValidators {
 
     return meta;
   }
+
+  public static checkExtensibleBeforeDecorate(target: object, name: NGXS_DECORATOR): void {
+    if (!Object.isExtensible(target)) {
```
arturovt: Isn't this check redundant? I mean who does `preventExtensions` on the component? I've never seen this:
```ts
@Component({ ... })
class AppComponent {
  constructor() {
    Object.preventExtensions(this);
  }
}
```

packages/store/src/utils/store-validators.ts
```diff
@@ -47,4 +48,16 @@ export abstract class StoreValidators {
...
+      throw new Error(MESSAGES[CODE.CHECK_EXTENSIBLE_BEFORE_DECORATE](name));
+    }
+  }
+
+  public static checkIsFrozenBeforeDecorate(target: object, name: NGXS_DECORATOR): void {
+    if (Object.isFrozen(target)) {
```
arturovt: Isn't this check redundant? I mean who does `freeze` the class? I've never seen this:
```ts
@Component({ ... })
class AppComponent {
  @Select() counter$;

  constructor() {
    Object.freeze(this);
  }
}
```

packages/store/src/utils/store-validators.ts
splincode: @markwhitfeld what do you think?

packages/store/src/utils/store-validators.ts
splincode: @markwhitfeld what do you think?

packages/store/src/decorators/select/select.ts
```diff
@@ -0,0 +1,34 @@
...
+/**
+ * Decorator for selecting a slice of state from the store.
+ */
+export function Select(rawSelector?: any, ...paths: string[]): PropertyDecorator {
+  return function(target: any, propertyKey: string | symbol): void {
+    StoreValidators.checkExtensibleBeforeDecorate(target, NGXS_DECORATOR.SELECT);
+
+    const name: string = propertyKey.toString();
+    const selectorId = `__${name}__selector`;
+    const selector: any = createSelectorFn(name, rawSelector, paths);
```
markwhitfeld: Does this need to be cast as `any`? Surely the inferred type would be better?
```suggestion
    const selector = createSelectorFn(name, rawSelector, paths);
```

packages/store/src/utils/store-validators.ts
markwhitfeld: Sorry, I also don't think that these checks add anything useful.

packages/store/src/utils/store-validators.ts
markwhitfeld: Sorry, I also don't think that these checks add anything useful.

splincode: done

packages/store/src/decorators/select/select.ts
splincode: done

packages/store/src/utils/store-validators.ts
splincode: done

Issue #429, opened 2018-06-08
@ngxs/form-plugin State changes every time a form action is dispatched


maxencefrenette:
### Versions

```
* ngxs: 3.1.3
* @angular/core: 6.0.3
```

### Repro steps

* Load any app using @ngxs/form-plugin
* Interact with the form

### Observed behavior

The state changes after each form action, which retriggers all selectors and rerenders components that depend on them.

### Desired behavior

For performance reasons, the state should only change identity when it actually changes.

### Mention any other details that might be useful (optional)

Sorry, I didn't take the time to create a stackblitz, but this issue is visible straight from the [code](https://github.com/ngxs/store/blob/master/packages/form-plugin/src/form.plugin.ts). As you can see, when an action such as `UpdateFormDirty` is dispatched, the state is updated even if the value of the `dirty` flag didn't change. This is horrible for performance and can easily fixed. I am willing to work on a PR if needed.

amcdnl:
If you have some ideas how to fix, feel free to PR them :)

maxencefrenette:
I played with it for a bit and I see two solutions non mutually exclusive improvements that can be made. It's not perfect, but let's just say the way Angular handles forms is not made for a Flux achitecture :stuck_out_tongue: .

First would be adding a check on `setValue()` so that it only modifies the value if it has changed. This would make the amount of selector reevaluation go down accross the board. It may cause wierd inconsistencies though. Maybe make a `setValueIfChanged()` helper ?

Second would be to dispatch one UpdateForm action instead of dispatching multiple actions. This would also have the benefit of uncluttering the devtools. However, to do that, we would need to either modify the semantics of `UpdateForm` or make a new action type.

Toughts ?

amcdnl:
I think option 2 sounds better. I've always wanted to do that anyways.

xban1x:
I noticed that with async validators you keep retrigerring them.

xban1x:
In its current form the plugin is unusable with async validators.

markwhitfeld:
@maxencefrenette Please could you confirm that this is fixed with the v3.5.1 release and close the issue.

markwhitfeld:
@xban1x Could you also test?

markwhitfeld:
Closing issue. Please reopen if it is not fixed.

PR #1216, opened 2019-08-07
feat(form-plugin): implement `NgxsFormPluginValueChangesStrategy`

arturovt:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: #429 

```
[ ] Yes
[x] No
```

## Other information

Currently our users don't have an access to the `valueChanges` stream that is listened under the hood and followed by dispatching 3 actions `UpdateFormValue`, `UpdateFormDirty`, `UpdateFormErrors`. `ngxsFormDebounce` binding provides the ability to create a timeout between dispatched actions.

This PR is very lightweight (not a lot code was added) and introduces a mechanism that is called "value changes strategy". The "noop" strategy, that doesn't change the behavior, is used by default. You can look at the `DeepEquals` strategy that pipes `valueChanges` stream with `distinctUntilChanged` operator and passes a callback that serializes previous and current value, and will generate event only if they differ. Thus we can reduce the amount of dispatched actions and give users the ability to do anything with `valueChanges` stream before actions are dispatched and state changed.

API:
```html
<form ngxsForm="..." [ngxsFormValueChangesStrategy]="ClassReference"> ... </form>
```

Notice that `ngxsFormValueChangesStrategy` binding awaits the class reference and not the class instance. Thus its instance can be retrieved lazily.

The same approach is used inside the `ngComponentOutlet` directive:
```ts
@Component({selector: 'hello-world', template: 'Hello World!'})
export class HelloWorld {
}

@Component({
  selector: 'ng-component-outlet-simple-example',
  template: `<ng-container *ngComponentOutlet="HelloWorld"></ng-container>`
})
export class NgTemplateOutletSimpleExample {
  // This field is necessary to expose HelloWorld to the template.
  HelloWorld = HelloWorld;
}
```

---

Strategies are provided on the directive level and not globally.  E.g. I'm using the `ngxsForm` directive at  10 places. I want to provide the `DeepEqualNgxsFormPluginValueChangesStrategy` only at 2 places.

---

Basically any developer can do ANY stuff with the underneath form as it gives an access to the `valueChanges` stream.


arturovt:
@splincode umm, hard to test, I thought if any test is broken then it did OK. :D

splincode:
@arturovt current behavior `state changes every time a form action is dispatched`, you can send PR with calculate count state changes before current pull request and update this PR with new behavior on tests

arturovt:
@splincode 

> calculate count state changes 

Dispatching 1 action instead of 3 is 3 three times less :smile:

splincode:
@arturovt This does not violate backward compatibility? 

```ts
import { UpdateFormDirty } from '@ngxs/form-plugin';

class MyComponent {
 constructor(private actions$: Actions) {}
 public ngOnInit(actions$: Action) {
   this.actions$.pipe(ofActionSuccessful(UpdateFormDirty)).subscribe(() => alert('update dirty'));
 }
}
```

arturovt:
@splincode yeah, you're right. Still need to discuss how we can reduce it

arturovt:
@splincode what do you think about `distinctUntilChanged` that compares objects?

packages/form-plugin/src/directive.ts
```diff
@@ -32,6 +39,11 @@ export class FormDirective implements OnInit, OnDestroy {
   ) {}
 
   ngOnInit() {
+    if (this.useDefaultCompare) {
+      // Create it lazily by condition
+      this.compare = (a, b) => JSON.stringify(a) === JSON.stringify(b);
```
splincode: I think is very hard

```ts
JSON.stringify(a) === JSON.stringify(b)
```

packages/form-plugin/tests/value-changes-strategy.spec.ts
```diff
@@ -0,0 +1,168 @@
...
+
+    subscription.unsubscribe();
+
+    // Assert
+    expect(dispatchedTimes).toBe(6);
+  });
+
+  it('should pipe via "distinctUntilChanged" if custom strategy provided with such operator', () => {
+    // Arrange
+    class CustomNgxsFormPluginValueChangesStrategy
```
splincode: I think you can make it public api.

```CustomNgxsFormPluginValueChangesStrategy``` -> ```NgxsFormPluginDeepEqualsStrategy```

example for people

```ts
import { NgxsFormPluginDeepEqualsStrategy } from '@ngxs/form-plugin';

NgxsFormPluginModule.forRoot({
    valueChangesStrategy: NgxsFormPluginDeepEqualsStrategy
})
```

markwhitfeld:
@arturovt Could you please describe the problem that this solves and what the new behavior provides in the PR description. Just a link to the original issue does not give enough information on what this is supposed to do.

splincode:
@arturovt Update docs in current PR

arturovt:
> @arturovt Could you please describe the problem that this solves and what the new behavior provides in the PR description. Just a link to the original issue does not give enough information on what this is supposed to do.

Sure. As I told you I will add descriptions for next PRs.

Currently our users don't have an access to the `valueChanges` stream that is listened under the hood and followed by dispatching 3 actions `UpdateFormValue`, `UpdateFormDirty`, `UpdateFormErrors`. `ngxsFormDebounce` binding provides the ability to create a timeout between dispatched actions.

This PR is very lightweight (not a lot code was added) and introduces a strategy that users can apply to the `valueChanges` stream. The "noop" strategy, that doesn't change the behavior, is used by default. You can look at the `DeepEquals` strategy that pipes `valueChanges` stream with `distinctUntilChanged` operator and passes a callback that serializes previous and current value, and will generate event only if they differ. Thus we can reduce the amount of dispatched actions and give users the ability to do anything with `valueChanges` stream before actions are dispatched and state changed.

splincode:
@arturovt @markwhitfeld  any updates?

packages/form-plugin/src/directive.ts
```diff
@@ -25,15 +36,22 @@ export class FormDirective implements OnInit, OnDestroy {
   @Input('ngxsFormClearOnDestroy')
   clearDestroy = false;
 
+  /**
+   * `valueChangesStrategy` is noop by default but can be overwritten via binding
+   */
+  @Input('ngxsFormValueChangesStrategy')
+  valueChangesStrategy: Type<
+    NgxsFormPluginValueChangesStrategy
+  > = NoopNgxsFormPluginValueChangesStrategy;
```
splincode: And if I get tired every time for each form to redefine the strategy?

packages/form-plugin/tests/value-changes-strategy.spec.ts
```diff
@@ -1,19 +1,19 @@
...
 import {
   NgxsFormPluginModule,
   UpdateFormValue,
   NgxsFormPluginValueChangesStrategy,
   DeepEqualNgxsFormPluginValueChangesStrategy
 } from '..';
+import { FormDirective } from '../src/directive';
 
-describe('NgxsFormPluginValueChangesStrategy', () => {
+fdescribe('NgxsFormPluginValueChangesStrategy', () => {
```
splincode: `fdescribe` instead `describe`?

packages/form-plugin/src/directive.ts
arturovt: Your suggestions?

splincode: support 

```ts
NgxsFormPlugin.forRoot({ defaultStategy: NoopNgxsFormPluginValueChangesStrategy })
```

arturovt: What is this for? Noop strategy is used by default.

splincode: But if I want to make a different strategy in the application for all forms at once

```ts
NgxsFormPlugin.forRoot({ defaultStategy: DeepEqualNgxsFormPluginValueChangesStrategy })
```

or 

```ts
NgxsFormPlugin.forRoot({ defaultStategy: CustomStrategy })
```

arturovt: That makes sense.

Tho to avoid making changes and then revert it back, let's wait for @markwhitfeld response. If it's OK then I will add this option.

arturovt:
@markwhitfeld 

The reason to add this feature is to let developers access the `valueChanges` property before dispatching those 3 actions.

> that could just be a simple setting on the directive

I'm open to any other solutions. Really. I do not insist on this "value changes strategy" so if you'd offer something easier - that would be great.

---

The first use case is that "deep equal strategy" that's provided in the PR. `distinctUntilChanged` is basically a "standard" today. By "standard' I mean that it's on of the most popular operators used in ~100% applications I guess.

Assume we've got a form with `username` and `email` controls and we want to check their uniqueness:

```ts
@Component({
  template: `
    <form [formGroup]="form">
      <input formControlName="username">
      <input formControlName="email">
      <p *ngIf="form.hasError('unique')">Values are not unique</form>
    </form>
  `
})
export class FormComponent {
  form = new FormGroup({
    username: new FormControl(),
    email: new FormControl()
  });

  constructor(private uniqueService: UniqueService) {
    this.setupListener();
  }

  private setupListener(): void {
    this.form.valueChanges
      .pipe(
        debounceTime(500),
        distinctUntilChanged((a, b) => JSON.stringify(a) === JSON.stringify(b)),
        switchMap(value => this.uniqueService.checkIfUsernameAndEmailAreUnique(value))
      )
      .subscribe(unique => {
        this.form.setErrors({ unique });
      });
  }
}
```

Async validator is not suitable here as we can't debounce it.

---

The second use case would be adding some "stream projection" operator like `switch|concatMap` as in the example above. This could be search for example among different predicates (for example name/start date/end date). I want to synchronize those predicates with my state also to save to local storage for example and to show the user next time:

```ts
@Component({
  template: `
    <form [formGroup]="form">
      <mat-form-field>
        <input matInput formControlName="name" />
      </mat-form-field>

      <mat-form-field>
        <input matInput [matDatepicker]="startDatePicker" formControlName="startDate" />
        <mat-datepicker-toggle matSuffix [for]="startDatePicker"></mat-datepicker-toggle>
        <mat-datepicker #startDatePicker></mat-datepicker>
      </mat-form-field>

      <mat-form-field>
        <input matInput [matDatepicker]="endDatePicker" formControlName="endDate" />
        <mat-datepicker-toggle matSuffix [for]="endDatePicker"></mat-datepicker-toggle>
        <mat-datepicker #endDatePicker></mat-datepicker>
      </mat-form-field>
    </form>
  `
})
export class FormComponent {
  form = new FormGroup({
    name: new FormControl(),
    startDate: new FormControl(),
    endDate: new FormControl()
  });

  constructor(private searchService: SearchService) {
    this.setupListener();
  }

  private setupListener(): void {
    this.form.valueChanges
      .pipe(
        distinctUntilChanged((a, b) => JSON.stringify(a) === JSON.stringify(b)),
        switchMap(value => this.searchService.search(value))
      )
      .subscribe(items => {
        this.store.dispatch(new UpdateFormValue(...))
      });
  }
}
```

---

So my point is that developers have to be able to access the `valueChanges` stream under the hood before dispatching 3 actions. As even my logger plugin tears apart when I put some letter into some input as it prints 3 actions immediately.

splincode:
@arturovt I think I know how it would be better, I will write an example

markwhitfeld:
I see what you are getting at, but the developer has access to the `valueChanges` stream on the form anyway. I think that the `distinctUntilChanged` approach using deep equality should be a simple attribute for the directive.
Something like:
```
<form [formGroup]="newNovelForm" ngxsForm="novels.newNovelForm" ngxsFormDistinct="deepEquality" >
```
This would expect the string "deepEquality".
What do you think?

arturovt:
@markwhitfeld 

> I see what you are getting at, but the developer has access to the `valueChanges` stream on the form anyway.

He has. But this stream is subscribed separately under the hood.

> I think that the distinctUntilChanged approach using deep equality should be a simple attribute for the directive

`distinctUntilChanged` was used for demonstration purposes. You can have the ability to apply ANY set of operators.

splincode:
@markwhitfeld @arturovt any progress?

arturovt:
> @markwhitfeld @arturovt any progress?

> @arturovt I think I know how it would be better, I will write an example

splincode:
@arturovt not idea

splincode:
Maybe, you can

```html
<form [formGroup]="newNovelForm" ngxsForm="novels.newNovelForm" ngxsFormDistinct="deepEquality" .. />
```

splincode:
Why closed?((

PR #1217, opened 2019-08-07
feat(form-plugin): reduce `UpdateFormStatus` action dispatching

arturovt:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: #429 

```
[ ] Yes
[x] No
```

### DIFF
--- a/packages/form-plugin/src/actions.ts
+++ b/packages/form-plugin/src/actions.ts
@@ -1,5 +1,3 @@
-import { FormStatus } from './symbols';
-
 export class UpdateFormStatus {
   static get type() {
     // NOTE: Not necessary to declare the type in this way in your code. See https://github.com/ngxs/store/pull/644#issuecomment-436003138
@@ -7,7 +5,7 @@ export class UpdateFormStatus {
   }
   constructor(
     public payload: {
-      status: FormStatus | null;
+      status: string | null;
       path: string;
     }
   ) {}
--- a/packages/form-plugin/src/actions.ts
+++ b/packages/form-plugin/src/actions.ts
@@ -1,3 +1,5 @@
+import { FormStatus } from './symbols';
+
 export class UpdateFormStatus {
   static get type() {
     // NOTE: Not necessary to declare the type in this way in your code. See https://github.com/ngxs/store/pull/644#issuecomment-436003138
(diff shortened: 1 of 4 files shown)


arturovt:
`statusChanges` stream emits every time the form value changes. It can emit `VALID` value 10 times, piping via `distinctUntilChanged` reduces the amount of dispatched actions.

packages/form-plugin/src/actions.ts
```diff
@@ -3,7 +3,12 @@ export class UpdateFormStatus {
     // NOTE: Not necessary to declare the type in this way in your code. See https://github.com/ngxs/store/pull/644#issuecomment-436003138
     return '[Forms] Update Form Status';
   }
-  constructor(public payload: { status: string | null; path: string }) {}
+  constructor(
+    public payload: {
+      status: 'VALID' | 'INVALID' | 'PENDING' | 'DISABLED' | null;
```
splincode: maybe use enum for status?

```ts
enum UpdateStatusType {
  VALID = "VALID"
  INVALID = "INVALID",
  ...
}
```

splincode:
*```feat``` instead ```fix```

packages/form-plugin/src/actions.ts
arturovt: @splincode yeah, I thought about it, the problem is that it can be a breaking change for our users.

P.S. `@angular/forms` doesn't use `enum`.

splincode: Okay, use type alias for our code

```
type UpdateFormStatus = 'VALID' | 'INVALID' | 'PENDING' | 'DISABLED' | null;
```

arturovt: @splincode done

arturovt:
@splincode I guess it's very strange that `@angular/forms` doesn't expose some `type` alias for these statuses. Btw I checked their code, they do type assertion everywhere like `status as 'VALID'` etc.

arturovt:
I removed `FormStatus` as it can really be a breaking change. We can add a `FormStatus` into v4 release.

Issue #700, opened 2018-12-04
Use classes for models 


CROSP:
Hi, thank you for awesome library. I have one question regarding use of classes with the store.

In your documentation models are defined as interfaces. As follows 

```
export interface ZebraFood {
  name: string;
  hay: number;
  carrots: number;
}
```

However consider the following case

I am defining my model as a class in order to be able to invoke methods. 

```
export default class Profile {
    firstName: string;
    lastName: string;
    avatarUrl: string;
    email: string;
    role: string;
    permissions: Array;
    age: number;

    public getFullName() {
        return this.firstName + ' ' + this.lastName;
    }
}
```
And my state module looks as defined further 

```
export class ProfileStateModel {
    currentUser: Profile;
}

@State<ProfileStateModel>({
    name: 'profile',
    defaults: {
        currentUser: null
    }
})
export class ProfileState {
    @Selector()
    static getCurrentUserProfile(state: ProfileStateModel) {
        return state.currentUser;
    }

    @Action(SetCurrentProfileUser)
    set({getState, patchState}: StateContext<ProfileStateModel>, {payload}: SetCurrentProfileUser) {
        patchState({
            currentUser: Object.assign({}, payload)
        });
    }
}
```

But when I request a state model from the store a simple object is returned instead. 

```
    @Select(ProfileState.getCurrentUserProfile) currentUserProfile$: Observable<Profile>;
     this.currentUserProfile$.subscribe(value => {
            console.log('CURRENT PROFILE IS ');
            console.log(value);
            this.currentProfile = value;
        });
```

So I am not able to call any methods. 
Without any doubts this decision makes sense. 

But what is the best way to have non anemic domain models?
Should I create a model from the state object while fetching it? 

 ```
       this.currentUserProfile$
            .map(value => Profile.createFromState(value))
            .subscribe(value => {
                console.log('CURRENT PROFILE IS ');
                console.log(value);
                this.currentProfile = value;
            });
```

Thanks in advance

viters:
https://stackblitz.com/edit/angular-dndwz6

Your example is not working because you are using `Object.assign({}, payload)`.

![image](https://user-images.githubusercontent.com/13100466/49690184-dd049d00-fb2c-11e8-9971-54bf71347fa6.png)

splincode:
@CROSP I will soon update the documentation and show how this can be achieved.

use transforms in the selector with
https://github.com/typestack/class-transformer

markwhitfeld:
[...]

PR #1231, opened 2019-08-13
docs: use classes for models

splincode:
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: https://github.com/ngxs/store/issues/700#issuecomment-446328500

### DIFF
--- a/docs/SUMMARY.md
+++ b/docs/SUMMARY.md
@@ -36,6 +36,7 @@
   * [Dynamic Plugins](recipes/dynamic-plugins.md)
   * [Authentication](recipes/authentication.md)
   * [Caching](recipes/cache.md)
+  * [Classes for State](recipes/classes-for-state.md)
   * [Unit Testing](recipes/unit-testing.md)
   * [Style Guide](recipes/style-guide.md)
 * Plugins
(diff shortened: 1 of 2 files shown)


markwhitfeld:
_This docs page is leading the dev to do the wrong thing._
You should not be storing instances of Classes in your state. You should only be storing object literals in your state. Please re-read what I said in this comment: https://github.com/ngxs/store/issues/700#issuecomment-446328500
The idea is that you would store them in your state as object literals and only convert them to classes in your selectors. I would also recommend only having object literals in your Action payloads, so the transformation from class to plain object should happen before adding to the payload of the action.

splincode:
Unfortunately, I do not agree. This is the only reason we use classes so that you do not need to write simple descriptions of large objects.

```ts
@State<BigAppModel>({
  name: 'app',
  defaults: new BigAppModel()
})
class BigAppState {}
```

This is very tiring if you have many states, we have about 100 states in the application. It’s bad to keep silent about what people do in real applications. In the same way, they can send a payload in the form of models, because TypeScript does not forbid you to do this.

PR #1247, opened 2019-08-19
fix(logger-plugin): print next state even if error was thrown

arturovt:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

prev

## What is the new behavior?
after

```
[ ] Yes
[x] No
```

## Other information
@splincode @markwhitfeld 

This is very small feature but as you can see at screenshots that the next state should be still logged into console. Given the following code:
```ts
catchError(error => {
  ctx.patchState({ counter });
  throw error;
})
```
We catch error, patch state and re-throw it. The state WILL be updated, but it will not be logged into console in the current behavior. That small feature fixes it.

### DIFF
--- a/packages/logger-plugin/src/logger.plugin.ts
+++ b/packages/logger-plugin/src/logger.plugin.ts
@@ -1,21 +1,24 @@
-import { Injectable, Inject } from '@angular/core';
+import { Injectable, Inject, Injector } from '@angular/core';
 import { tap, finalize, catchError } from 'rxjs/operators';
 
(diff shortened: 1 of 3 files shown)


markwhitfeld:
@arturovt I'll have a look now, but isn't this actually a fix for a bug and not a feature?

arturovt:
@markwhitfeld that could be also a fix thus we could publish it with `3.5.1`. I'm relying on your opinion in that case

Issue #329, opened 2018-04-26
patchState new value does not reflect in Redux DevTools


kuncevic:
### Versions
```
* ngxs: 3.0.0-rc.1 and 3.0.0-dev.master-2df48fe
* @angular/core: 6.0.0-rc.5
```

### Repro steps

Have `loading: false` in the state.
Patch the state  from an action `sc.patchState({ loading: true });`

### Observed behavior

 The Redux DevTools still show `loading: false` 

### Desired behavior

 Redux DevTools shold show `loading: true`

markwhitfeld:
This may be related in part to issue #311.
Is this behavior only evident when using the patch method, or does it happen if you use setState too?

kuncevic:
If I `@select` the store I am getting the new updated value, so there is only dev-tools does not updating/syncing with a new value in case `patchState`. Cant see any issues with `setState` atm.

kuncevic:
Just updated to rc.4 the issue still there

amcdnl:
I don’t know how It’s possible this is still not working. And just to confirm you said it worked fine with set state?

kuncevic:
All good now

garthmason:
I confirm I am not seeing any issues using `patchState` with the following versions:

`"@angular/core": "6.0.0"`
and 
`"@ngxs/devtools-plugin": "^3.0.0-rc.4"`
` "@ngxs/logger-plugin": "^3.0.0-rc.4"`
` "@ngxs/store": "^3.0.0-rc.4",`

GreenMonkeyBoy:
Looks like it happens if you use patchState() or setState() in the callback methods of a http request.

Demo 1: https://stackblitz.com/edit/ngxs-devtools-not-reflecting-state-subscribe
In this demo, I use subscribe() in the action handler (not sure if it's recommended though), then I use setState() but the devtools is always "one state behind" the state in the real store.

Demo 2: https://stackblitz.com/edit/ngxs-devtools-not-reflecting-state-tap
In this demo, I use tap() like in the [Authentication recipe](https://ngxs.gitbook.io/ngxs/recipes/authentication) from the doc. The devtools is sync on a succeed request, but nothing happens on a failed request.

Hope it will help you.

piernik:
I have the same problem. Any news?

arturovt:
@kuncevic Is it still reproducable? Because I cannot reproduce it :(

arturovt:
[...]

PR #1249, opened 2019-08-19
fix(devtools-plugin): send action even if error was thrown

arturovt:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: #329

prev-devtools

## What is the new behavior?
after-devtools

```
[ ] Yes
[x] No
```

## Other information
@splincode @markwhitfeld 

This is very small bugfix for the old #329 issue. In the current behavior our `devtools-plugin` notifies Chrome extension only if everything is OK and no error is thrown. This PR fixes this behavior and notifies Chrome extension to log the state even if error was thrown. Given the following code:
```ts
catchError(error => {
  ctx.patchState({ counter });
  throw error;
})
```

This change will not be shown in the Chrome extension in the current behavior.

### DIFF
--- a/packages/devtools-plugin/src/devtools.plugin.ts
+++ b/packages/devtools-plugin/src/devtools.plugin.ts
@@ -1,6 +1,6 @@
 import { Inject, Injectable, Injector } from '@angular/core';
 import { getActionTypeFromInstance, NgxsNextPluginFn, NgxsPlugin, Store } from '@ngxs/store';
-import { tap } from 'rxjs/operators';
+import { tap, catchError } from 'rxjs/operators';
 
 import {
   NGXS_DEVTOOLS_OPTIONS,
@@ -57,6 +57,11 @@ export class NgxsReduxDevtoolsPlugin implements NgxsPlugin {
 
           this.devtoolsExtension!.send({ ...action, type }, newState);
         }
+      }),
+      catchError(error => {
+        const type = getActionTypeFromInstance(action);
+        this.devtoolsExtension!.send({ ...action, type }, this.store.snapshot());
+        throw error;
       })
     );
   }
--- a/packages/devtools-plugin/src/devtools.plugin.ts
+++ b/packages/devtools-plugin/src/devtools.plugin.ts
@@ -48,22 +48,25 @@ export class NgxsReduxDevtoolsPlugin implements NgxsPlugin {
 
     return next(state, action).pipe(
       catchError(error => {
-        const type = getActionTypeFromInstance(action);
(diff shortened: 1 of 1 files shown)


packages/devtools-plugin/src/devtools.plugin.ts
```diff
@@ -57,6 +57,11 @@ export class NgxsReduxDevtoolsPlugin implements NgxsPlugin {
 
           this.devtoolsExtension!.send({ ...action, type }, newState);
         }
+      }),
+      catchError(error => {
+        const type = getActionTypeFromInstance(action);
+        this.devtoolsExtension!.send({ ...action, type }, this.store.snapshot());
+        throw error;
```
markwhitfeld: We could also get here if something in the `tap` part throws an error (like an error calling the dev tool).
Maybe we could use the `finally` pipe instead and combine the code?
Also, if there is an error for the `@@INIT` action would the dev tools extension not get initialized?

arturovt: The problem is that Rx doesn't pass anything as arguments to the `finalize` operator, no error and no `nextState`

markwhitfeld: Hmmm. Then could you catch before the `tap`?

arturovt: I'll try

packages/devtools-plugin/src/devtools.plugin.ts
```diff
@@ -47,6 +47,11 @@ export class NgxsReduxDevtoolsPlugin implements NgxsPlugin {
     }
 
     return next(state, action).pipe(
+      catchError(error => {
+        const type = getActionTypeFromInstance(action);
+        this.devtoolsExtension!.send({ ...action, type }, this.store.snapshot());
+        throw error;
+      }),
```
markwhitfeld: Ok, this is better but what if the action that errors here was the `@@INIT` action?
The dev tool would still need to init. Should we maybe move the init code somewhere else?

markwhitfeld: I refactored so that the same logic is shared when sending to the dev tools.

markwhitfeld:
@arturovt @splincode Please review my changes. Just made the logic for sending to the dev tool shared.

arturovt:
@markwhitfeld I'm OK with your changes :)

Issue #561, opened 2018-09-05
Logger logs out of order when multiple actions are dispatched


guilhermechiara:
### Versions

```
* @angular/core: 6.0.0
* @ngxs/store: 3.2.0
* @ngxs/logger-pluggin: 3.2.0
```

### Repro steps

Stackblitz: https://stackblitz.com/edit/angular-l4zxsn

* Create two different actions and their respective Success actions.
* Use a service to return data.
* Dispatch the two actions at the same time.

### Observed behavior

When you have multiple actions that are dispatched at the same time, the logger group the actions, creating some confusion to understand what is happening.

![image](https://user-images.githubusercontent.com/5012603/45090530-e2a6ad00-b10f-11e8-8eab-75e3b7d0082d.png)

```
ngOnInit() {
    this.store.dispatch(new LoadPizzas());
    this.store.dispatch(new LoadToppings());
  }
```

```
@Action(LoadPizzas)
  load(context: StateContext<PizzasStateModel>) {
    return this.http.get('https://www.mocky.io/v2/5b8fb99d2e00002a00a89be4?mocky-delay=3000ms')
    .pipe(
      map(value => context.dispatch(new LoadPizzaSuccess(value)))
    );
  }
```

### Desired behavior

I think that the logger should not group those Actions, since they are not exactly children of each other. It creates a little bit confusion when you look at your logs and try to understand what is happening. Actually, I don't know whether this is a normal behavior or not.

Thanks in advance! :)

markwhitfeld:
This is a duplicate of issue #213.

internalsystemerror:
Cancelling a previous uncompleted action results in this nesting also, I suspect the resolution for that would be the same as for this.

Issue #276, opened 2018-04-12
Remove Log Groups


amcdnl:
The log groups don't work right and can't work right w/ the way we do async actions.

amcdnl:
Ref #213

markwhitfeld:
@amcdnl The ordered actions does not fix this.
We will have to look at implementing this or an alternative fix.

Steven-Harris:
+1 also experiencing this issue

Issue #213, opened 2018-04-02
Logger plugin nests log data for actions when dispatching multiple actions


mflorence99:
Hey I totally love ngxs for all kinds of reasons, and I have built several large projects with ngrx. So this is not a big feature request in the scale of things.

When dispatching multiple actions at once, the console log grouping is such that the log data for the second action is nested inside the first, the third inside the second and so on. It has tripped me up a couple of times because I look at the log and think I have a bug because it appears that an action is not being dispatched. Of course it is, I just have to click and expand multiple levels of grouping.

Could it be that the log lines are issued serially? So `console.group(...) ... console.groupEnd()` sequences one after the other, rather than nested inside each other?

Thanks!

amcdnl:
Good catch, things can execute out of sequence so I can see how this might happen.

tsawan:
@mflorence99 I can see this is fixed now. Can you try again with latest code?

mflorence99:
I tried `"@ngxs/store": "2.1.0-beta.9"` and `"@ngxs/logger-plugin": "latest"` if that's what you meant, but the anomaly appears worse, sadly. All the log lines are buried inside of the `action @@INIT` log (so that you have to expand it to see them). It's as if `console.groupEnd()` was not being called. Should I have tried the v3 beta?

un33k:
Yep, at some point, all other logs are nested.

![screen shot 2018-04-09 at 12 13 05 pm](https://user-images.githubusercontent.com/407537/38509278-74437ca2-3bef-11e8-8cb1-39a939caeba8.png)

Also, it would be nice to make the option `logger: any` -> `logger?: any`.
In most cases, the console is the default choice.

amcdnl:
The fix for this is it only needs to log on success, the before logger is getting messed up by async actions.

mflorence99:
I agree with you on nested actions. If an action itself issues a dispatch for another action then fair enough the log lines themselves should appear nested. But the case I initially observed originated from using the dispatch API to trigger multiple actions at once by passing an array. I don’t think of these actions as being nested.

un33k:
@markwhitfeld as of `3.0.0-dev.4e68823`, the problem persists.

amcdnl:
We are going to have to remove grouping. Its just not going to work the way we do thing. I'll do this in 3.0

un33k:
It makes sense. Looking forward to it. 

[...]

PR #1262, opened 2019-08-23
 refactor(logger-plugin): extract ActionLogger for logging action lifecycle

markwhitfeld:
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

## What is the reason for refactoring?

The logger plugin has so many responsibilities mixed in one place that it is a real effort to figure out what is happening. This refactoring is in preparation for adding a fix for issue #276,  #213 and #561

```
[ ] Yes
[x] No
```

### DIFF
--- a/packages/logger-plugin/src/action-logger.ts
+++ b/packages/logger-plugin/src/action-logger.ts
@@ -1,0 +1,103 @@
+import { getActionTypeFromInstance, Store } from '@ngxs/store';
+
+import { pad } from './internals';
+import { NgxsLoggerPluginOptions } from './symbols';
+
+export class ActionLogger {
+  constructor(
+    private action: any,
+    private logger: any,
+    private options: NgxsLoggerPluginOptions,
+    private store: Store
+  ) {}
+
+  dispatched(state: any) {
+    const options = this.options;
+    const logger = this.logger;
+    const event = this.action;
+
+    const actionName = getActionTypeFromInstance(event);
+    const time = new Date();
+
+    // tslint:disable-next-line
+    const formattedTime = ` @ ${pad(time.getHours(), 2)}:${pad(time.getMinutes(), 2)}:${pad(
+      time.getSeconds(),
+      2
+    )}.${pad(time.getMilliseconds(), 3)}`;
+
+    const message = `action ${actionName}${formattedTime}`;
+    const startMessage = options.collapsed ? logger.groupCollapsed : logger.group;
+
+    try {
+      startMessage.call(logger, message);
+    } catch (e) {
+      console.log(message);
+    }
+
+    // print payload only if at least one property is supplied
+    if (this._hasPayload(event)) {
+      this.log('payload', 'color: #9E9E9E; font-weight: bold', { ...event });
+    }
+
+    this.log('prev state', 'color: #9E9E9E; font-weight: bold', state);
+  }
+
+  completed(nextState: any) {
+    this.log('next state', 'color: #4CAF50; font-weight: bold', nextState);
+    this.endGroup();
+  }
+
+  errored(error: any) {
+    this.log(
+      'next state after error',
+      'color: #FD8182; font-weight: bold',
+      this.store.snapshot()
+    );
+    this.log('error', 'color: #FD8182; font-weight: bold', error);
+    this.endGroup();
+  }
+
+  endGroup() {
+    try {
+      this.logger.groupEnd();
+    } catch (e) {
+      this.logger.log('—— log end ——');
+    }
+  }
+
+  log(title: string, color: string, payload: any) {
+    if (this.isIE()) {
+      this.logger.log(title, payload);
+    } else {
+      this.logger.log('%c ' + title, color, payload);
+    }
+  }
+
+  isIE(): boolean {
+    const ua =
+      typeof window !== 'undefined' && window.navigator.userAgent
+        ? window.navigator.userAgent
+        : '';
+    let msIE = false;
+
+    const oldIE = ua.indexOf('MSIE ');
+    const newIE = ua.indexOf('Trident/');
+
+    if (oldIE > -1 || newIE > -1) {
+      msIE = true;
+    }
+
+    return msIE;
+  }
+
+  private _hasPayload(event: any) {
+    const nonEmptyProperties = this._getNonEmptyProperties(event);
+    return nonEmptyProperties.length > 0;
+  }
+
+  private _getNonEmptyProperties(event: any) {
+    const keys = Object.keys(event);
+    const values = keys.map(key => event[key]);
+    return values.filter(value => !!value);
+  }
+}
--- a/packages/logger-plugin/src/action-logger.ts
+++ b/packages/logger-plugin/src/action-logger.ts
@@ -1,93 +1,35 @@
 import { getActionTypeFromInstance, Store } from '@ngxs/store';
 
-import { pad } from './internals';
-import { NgxsLoggerPluginOptions } from './symbols';
+import { formatTime } from './internals';
+import { LogWriter } from './log-writer';
 
 export class ActionLogger {
-  constructor(
-    private action: any,
-    private logger: any,
-    private options: NgxsLoggerPluginOptions,
-    private store: Store
-  ) {}
+  constructor(private action: any, private store: Store, private logWriter: LogWriter) {}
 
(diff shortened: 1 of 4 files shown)


packages/logger-plugin/src/action-logger.ts
```diff
@@ -0,0 +1,45 @@
...
+
+import { formatTime } from './internals';
+import { LogWriter } from './log-writer';
+
+export class ActionLogger {
+  constructor(private action: any, private store: Store, private logWriter: LogWriter) {}
+
+  dispatched(state: any) {
+    const actionName = getActionTypeFromInstance(this.action);
+    const formattedTime = formatTime(new Date());
```
splincode: Why you used `formatTime(new Date())` ? instead

```ts
import { DatePipe } from '@angular/common';

const formattedTime = new DatePipe('en-EU').transform(new Date(), 'HH:mm:ss.SSS');
```

packages/logger-plugin/src/action-logger.ts
```diff
@@ -0,0 +1,45 @@
...
+
+    const message = `action ${actionName} @ ${formattedTime}`;
+    this.logWriter.startGroup(message);
+
+    // print payload only if at least one property is supplied
+    if (this._hasPayload(this.action)) {
+      this.logWriter.logGrey('payload', { ...this.action });
+    }
+
+    this.logWriter.logGrey('prev state', state);
```
splincode: what is 'prev state', 'next state', 'next state after error', 'error' ?

Maybe we usage 

```ts
export enum REDUX_LOG_PREFIX {
 PREV = 'prev state',
 NEXT = 'next state',
 ERROR = 'error',
 NEXT_AFTER_ERROR = 'next state after error'
}
```

or something

packages/logger-plugin/src/log-writer.ts
```diff
@@ -0,0 +1,62 @@
...
+  endGroup() {
+    try {
+      this.logger.groupEnd();
+    } catch (e) {
+      this.logger.log('—— log end ——');
+    }
+  }
+
+  logGrey(title: string, payload: any) {
+    this.log(title, 'color: #9E9E9E; font-weight: bold', payload);
```
splincode: maybe we can

```ts
export enum STYLE {
 GREY = 'color: #9E9E9E; font-weight: bold',
 GREEN = 'color: #4CAF50; font-weight: bold',
 REDIS = 'color: #FD8182; font-weight: bold'
}
```

packages/logger-plugin/src/log-writer.ts
```diff
@@ -0,0 +1,62 @@
...
+
+  log(title: string, color: string, payload: any) {
+    if (this.isIE()) {
+      this.logger.log(title, payload);
+    } else {
+      this.logger.log('%c ' + title, color, payload);
+    }
+  }
+
+  isIE(): boolean {
```
splincode: There is no support anywhere else

*Internet Explorer (IE) 10 will reach end of support on January 31, 2020. From this point forward, IE 11 will be the only supported version of Internet Explorer on Windows Server 2012 and Windows Embedded 8 Standard. IE 11 is now available for piloting via Microsoft Update Catalog

packages/logger-plugin/src/action-logger.ts
markwhitfeld: Good idea. It was a refactor, so I just extracted code that was already there.

packages/logger-plugin/src/logger.plugin.ts
```diff
@@ -18,94 +19,22 @@ export class NgxsLoggerPlugin implements NgxsPlugin {
       return next(state, event);
     }
 
-    const options = this._options || {};
-    const logger = options.logger || console;
-    const actionName = getActionTypeFromInstance(event);
-    const time = new Date();
+    const logWriter = new LogWriter(this._options);
+    // Retrieve lazily to avoid cyclic dependency exception
+    const store = this._injector.get<Store>(Store);
```
splincode:
```ts
class NgxsLoggerPlugin {
   myMethodExample(event) {
     const actionLogger = new ActionLogger(event, this.store, this.logWriter)
   }

   private get store(): Store {
     return this._injector.get<Store>(Store);
   }

   private get logWriter(): LogWriter {
    return new LogWriter(this._options);
   }

}

```

packages/logger-plugin/src/action-logger.ts
markwhitfeld: I suspect that the current method may be faster than this because it is less generic, and it might be a smaller bundle size too (doesn't bring in the Angular pipe dependency).

splincode: @angular/common - every where usage in angular app as dependency
also you can check bundle size after refactor (npm run bundlesize)

splincode: `I suspect that the current method may be faster` 

firstly, the previous code is completely incomprehensible, I had to run to understand what it was doing

secondly, people should not use the plugin in production

```ts
@NgModule({
 imports: [
   environment.production ? [] : [ NgxsLoggerPlugin.forRoot() ]
 ]
})
export class AppModule {}
```

packages/logger-plugin/src/log-writer.ts
markwhitfeld: They are only used in this one place so the name of the colour is obvious.
Are you aware of the amount of generated code that these enums add? It is quite a bit.
I may just extract some variables to make it clearer though.

packages/logger-plugin/src/log-writer.ts
markwhitfeld: Interesting. We will need to keep this in mind.
This is a refactoring, so I will not remove code that is possibly defining a behavior.

packages/logger-plugin/src/logger.plugin.ts
markwhitfeld: This introduces abstraction without adding clarity. I don't think that this is necessary.

packages/logger-plugin/src/action-logger.ts
markwhitfeld: I am happy to use the pipe, but please bear in mind that this is a refactor, so I am just changing the structure of the code, not rewriting it.

packages/logger-plugin/src/action-logger.ts
splincode: But if you want to write tests, you should use them to compare

packages/logger-plugin/src/log-writer.ts
splincode: `I may just extract some variables to make it clearer though` - please

packages/logger-plugin/src/logger.plugin.ts
splincode: up to you

packages/logger-plugin/src/log-writer.ts
splincode: ok

packages/logger-plugin/src/action-logger.ts
splincode: up to you

packages/logger-plugin/src/action-logger.ts
markwhitfeld: Your tests should not use production code to validate production code.
The test should be explicit about its expectations. If the production code changes behaviour then the test should fail. Just a small tip from our TDD 101 course

packages/logger-plugin/src/logger.plugin.ts
markwhitfeld: Left it as is.

packages/logger-plugin/src/action-logger.ts
markwhitfeld: Left it as is.

packages/logger-plugin/src/action-logger.ts
markwhitfeld: Left it as is.

packages/logger-plugin/src/log-writer.ts
markwhitfeld: I added variables to be more descriptive of the style.

markwhitfeld:
Should be good to go in my opinion. @arturovt could you review?

PR #1269, opened 2019-08-24
fix(router-plugin): log async completion in separate group

markwhitfeld:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number:  #276, #213 and #561

Currently the logger writes an action to the console with one group that closes after the action completes. This works great if all actions only do synchronous work but creates a mess in the console as soon as there is asynchronous completion of actions.

## What is the new behavior?
If an action will complete asynchronously then the console group for the action is closed after all synchronous work is done and a new group is opened when the asynchronous completion or error happens.

```
[ ] Yes
[x] No
```

### DIFF
--- a/packages/logger-plugin/src/action-logger.ts
+++ b/packages/logger-plugin/src/action-logger.ts
@@ -4,13 +4,14 @@ import { formatTime } from './internals';
 import { LogWriter } from './log-writer';
 
 export class ActionLogger {
+  private synchronousWorkEnded = false;
+  private actionCompleted = false;
+  private startedTime = new Date();
+
   constructor(private action: any, private store: Store, private logWriter: LogWriter) {}
 
   dispatched(state: any) {
-    const actionName = getActionTypeFromInstance(this.action);
-    const formattedTime = formatTime(new Date());
-
-    const message = `action ${actionName} @ ${formattedTime}`;
+    const message = this.getActionLogHeader();
     this.logWriter.startGroup(message);
 
     // print payload only if at least one property is supplied
@@ -22,14 +23,40 @@ export class ActionLogger {
   }
 
   completed(nextState: any) {
+    if (this.synchronousWorkEnded) {
+      const message = `(async work completed) ${this.getActionLogHeader()}`;
+      this.logWriter.startGroup(message);
+    }
     this.logWriter.logGreen('next state', nextState);
     this.logWriter.endGroup();
+    this.actionCompleted = true;
   }
 
   errored(error: any) {
+    if (this.synchronousWorkEnded) {
+      const message = `(async work error) ${this.getActionLogHeader()}`;
+      this.logWriter.startGroup(message);
+    }
     this.logWriter.logRedish('next state after error', this.store.snapshot());
     this.logWriter.logRedish('error', error);
     this.logWriter.endGroup();
+    this.actionCompleted = true;
+  }
+
+  syncWorkComplete() {
+    if (!this.actionCompleted) {
+      this.logWriter.logGreen('next state (synchronous)', this.store.snapshot());
+      this.logWriter.logGreen('( action doing async work... )', undefined);
+      this.logWriter.endGroup();
+    }
+    this.synchronousWorkEnded = true;
+  }
+
+  private getActionLogHeader() {
+    const actionName = getActionTypeFromInstance(this.action);
+    const formattedTime = formatTime(this.startedTime);
+    const message = `action ${actionName} (started @ ${formattedTime})`;
+    return message;
   }
 
   private _hasPayload(event: any) {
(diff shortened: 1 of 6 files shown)


packages/logger-plugin/src/logger.plugin.ts
```diff
@@ -36,5 +37,17 @@ export class NgxsLoggerPlugin implements NgxsPlugin {
...
+    return afterSubscribe(result, () => actionLogger.syncWorkComplete());
   }
 }
+
+function afterSubscribe<T>(source: Observable<T>, callback: VoidFunction) {
+  return merge(
+    source,
+    defer(() => {
+      callback();
+      return empty();
```
arturovt:
```suggestion
      return EMPTY;
```

markwhitfeld: done

packages/logger-plugin/src/action-logger.ts
```diff
@@ -22,14 +23,40 @@ export class ActionLogger {
...
+      this.logWriter.logGreen('( action doing async work... )', undefined);
+      this.logWriter.endGroup();
+    }
+    this.synchronousWorkEnded = true;
+  }
+
+  private getActionLogHeader() {
+    const actionName = getActionTypeFromInstance(this.action);
+    const formattedTime = formatTime(this.startedTime);
+    const message = `action ${actionName} (started @ ${formattedTime})`;
```
splincode: is redunant, you can 

```ts
return `action ${actionName} (started @ ${formattedTime})`;
```

In addition, I recommend always describing the return type, since in the future you can forget to return an object instead of a returned string.

```ts
private getActionLogHeader(): string {
 
}
```

packages/logger-plugin/src/action-logger.ts
```diff
@@ -22,14 +23,40 @@ export class ActionLogger {
...
+      const message = `(async work completed) ${this.getActionLogHeader()}`;
+      this.logWriter.startGroup(message);
+    }
     this.logWriter.logGreen('next state', nextState);
     this.logWriter.endGroup();
+    this.actionCompleted = true;
   }
 
   errored(error: any) {
+    if (this.synchronousWorkEnded) {
```
splincode: What is `synchronousWorkEnded`?

packages/logger-plugin/src/log-writer.ts
```diff
@@ -8,11 +8,12 @@ export class LogWriter {
   }
 
   startGroup(message: string) {
-    const startGroupFn = this.options.collapsed
-      ? this.logger.groupCollapsed
-      : this.logger.group;
     try {
```
splincode: why we need use try-catch it?

packages/logger-plugin/tests/logger.plugin.spec.ts
```diff
@@ -69,93 +104,193 @@ describe('NgxsLoggerPlugin', () => {
...
+    const payload = 'qux';
 
-    expect(logger.callStack).toEqual(expectedCallStack);
+    // Act
+    const promise = store.dispatch(new AsyncAction(payload)).toPromise();
+    logger.log('Some other work');
+    await promise;
+
+    // Assert
+    const expectedCallStack = [
```
splincode: I suggest also adding a test that will call three asynchronous actions and see how it will look in the logs

packages/logger-plugin/src/action-logger.ts
arturovt: `ActionLogger` instance is created when any action is dispatched. As actions can be sync and async this is a flag that basically says "oh I've completed doing my synchronous job".

splincode: I think for some reason this is all very complicated, but oh well

arturovt: Need more comments I guess

packages/logger-plugin/src/action-logger.ts
markwhitfeld: Agreed. Will do

packages/logger-plugin/src/log-writer.ts
markwhitfeld: I'm not sure. It was in the original code, so I didn't want to mess with something I don't know the reason for.

packages/logger-plugin/tests/logger.plugin.spec.ts
markwhitfeld: Great idea. Especially if they call each other.

packages/logger-plugin/src/log-writer.ts
splincode: ok

splincode:
Will we see this in the release?

markwhitfeld:
No, probably not. This will require quite a bit more code to resolve and I am spending all my available time on reviewing PRs 😄

blockerdude:
Just wanted to bump this and see how things are coming along.

PR #1270, opened 2019-08-25
fix(store): warn if the zone is not actual "NgZone"

arturovt:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: N/A

```
[ ] Yes
[x] No
```

## Other information
Nooped zone is used by people who prefer to run change detection manually in their apps, this requires permanent calls to tick. If the zone is nooped then nothing will happen actually and the user will not benefit from using our default execution strategy, because the zone does nothing basically.

### DIFF
--- a/packages/store/src/configs/messages.config.ts
+++ b/packages/store/src/configs/messages.config.ts
@@ -21,8 +21,8 @@ export const CONFIG_MESSAGES: ObjectKeyMap<Function> = {
     `State name '${current}' from ${newName} already exists in ${oldName}`,
   [VALIDATION_CODE.STATE_DECORATOR]: () => 'States must be decorated with @State() decorator',
   [VALIDATION_CODE.INCORRECT_PRODUCTION]: () =>
-    'Angular is running in the production mode but NGXS is still running in the development mode!\n' +
-    'Please set developmentMode to false on the NgxsModule options in the production mode.\n' +
+    'Angular is running in production mode but NGXS is still running in the development mode!\n' +
+    'Please set developmentMode to false on the NgxsModule options when in production mode.\n' +
     'NgxsModule.forRoot(states, { developmentMode: !environment.production })',
   [VALIDATION_CODE.INCORRECT_DEVELOPMENT]: () =>
     'RECOMMENDATION: Set developmentMode to true on the NgxsModule when Angular is running in development mode.\n' +
--- a/packages/store/src/configs/messages.config.ts
+++ b/packages/store/src/configs/messages.config.ts
@@ -21,8 +21,8 @@ export const CONFIG_MESSAGES: ObjectKeyMap<Function> = {
     `State name '${current}' from ${newName} already exists in ${oldName}`,
   [VALIDATION_CODE.STATE_DECORATOR]: () => 'States must be decorated with @State() decorator',
   [VALIDATION_CODE.INCORRECT_PRODUCTION]: () =>
-    'Angular is running in production mode but NGXS is still running in the development mode!\n' +
-    'Please set developmentMode to false on the NgxsModule options when in production mode.\n' +
+    'Angular is running in the production mode but NGXS is still running in the development mode!\n' +
+    'Please set developmentMode to false on the NgxsModule options in the production mode.\n' +
     'NgxsModule.forRoot(states, { developmentMode: !environment.production })',
   [VALIDATION_CODE.INCORRECT_DEVELOPMENT]: () =>
     'RECOMMENDATION: Set developmentMode to true on the NgxsModule when Angular is running in development mode.\n' +
@@ -33,5 +33,7 @@ export const CONFIG_MESSAGES: ObjectKeyMap<Function> = {
     '@Action() decorator cannot be used with static methods',
   [VALIDATION_CODE.SELECTOR_DECORATOR]: () => 'Selectors only work on methods',
   [VALIDATION_CODE.ZONE_WARNING]: () =>
-    'Your application was bootstrapped with nooped zone and your execution strategy requires an actual NgZone'
+    'Your application was bootstrapped with nooped zone and your execution strategy requires an actual NgZone!\n' +
+    'Please set the value of the executionStrategy property to NoopNgxsExecutionStrategy.\n' +
+    'NgxsModule.forRoot(states, { executionStrategy: NoopNgxsExecutionStrategy })'
 };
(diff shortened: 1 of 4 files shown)


packages/store/tests/execution/dispatch-outside-zone-ngxs-execution-strategy.spec.ts
```diff
@@ -295,62 +285,89 @@ describe('DispatchOutsideZoneNgxsExecutionStrategy', () => {
...
+        bootstrap: [MockComponent]
+      })
+      class MockModule {}
+
+      // Act
+      const warnings: string[] = [];
+
+      console.warn = (...args: string[]) => {
+        warnings.push(args[0]);
+      };
```
markwhitfeld: This should be restored back to the normal console.warn afterwards, otherwise this test could cause unexpected results.
This can be done in a `finally`.

markwhitfeld: Potentially you could make a helper method similar to `freshPlatform` called `preserveConsole` that just stores a copy of all the properties of console to another object (`const backupOfConsole = { ...console };`) then it executes the test, and then in a `finally` it writes all those properties back to console (`Object.assign(console, backupOfConsole)`);
... or something like that.

markwhitfeld:
Just one question. If someone has nooped the ngZone then what is their intention?
How would it impact a normal app?
Can we maybe change the default ngxs execution strategy accordingly?
Just thinking out aloud.

arturovt:
> Just one question. If someone has nooped the ngZone then what is their intention?
> How would it impact a normal app?
> Can we maybe change the default ngxs execution strategy accordingly?
> Just thinking out aloud.

Nooped zone is used by people who prefer to run change detection manually in their apps, this requires permanent calls to `tick`. If the zone is nooped then nothing will happen actually and the user will not benefit from using our default execution strategy, because the zone does nothing basically.

arturovt:
> Potentially you could make a helper method similar to freshPlatform called preserveConsole that just stores a copy of all the properties of console to another object (const backupOfConsole = { ...console };)

I guess it could be quite complicated so I just used native Jest's `spyOn`.

packages/store/tests/execution/dispatch-outside-zone-ngxs-execution-strategy.spec.ts
```diff
@@ -311,16 +311,13 @@ describe('DispatchOutsideZoneNgxsExecutionStrategy', () => {
...
-        warnings.push(args[0]);
-      };
+      const spy = jest.spyOn(console, 'warn').mockImplementation();
 
       await platformBrowserDynamic().bootstrapModule(MockModule, { ngZone: 'noop' });
 
       // Assert
-      expect(warnings).toEqual([CONFIG_MESSAGES[VALIDATION_CODE.ZONE_WARNING]()]);
+      expect(spy).toHaveBeenCalledWith(CONFIG_MESSAGES[VALIDATION_CODE.ZONE_WARNING]());
+      spy.mockRestore();
```
markwhitfeld: This line won't get called if the expectation on the previous line fails.

markwhitfeld:
> Nooped zone is used by people who prefer to run change detection manually in their apps, this requires permanent calls to `tick`. If the zone is nooped then nothing will happen actually and the user will not benefit from using our default execution strategy, because the zone does nothing basically.

Please add this to the PR description...
Also, maybe the error message could recommend the `NoopNgxsExecutionStrategy`

packages/store/tests/execution/dispatch-outside-zone-ngxs-execution-strategy.spec.ts
arturovt: I'm not able to call `mockRestore` before `expect` as Jest releases underneath resources.

markwhitfeld: This is why I suggested wrapping it in a function or as a finally part of a try catch block.

arturovt: Oh, okay :wink:

packages/store/tests/config-validator.spec.ts
```diff
@@ -9,9 +9,6 @@ import { CONFIG_MESSAGES, VALIDATION_CODE } from '../src/configs/messages.config
 describe('ConfigValidator', () => {
   let validator: ConfigValidator;
   let host: HostEnvironment;
-  let actualWarning: string;
-
-  console.warn = (...args: string[]) => (actualWarning = args[0]);
```
arturovt: Had to edit this file because it patches `console.warn`.

packages/store/tests/execution/dispatch-outside-zone-ngxs-execution-strategy.spec.ts
arturovt: Done

packages/store/src/configs/messages.config.ts
```diff
@@ -21,8 +21,8 @@ export const CONFIG_MESSAGES: ObjectKeyMap<Function> = {
     `State name '${current}' from ${newName} already exists in ${oldName}`,
   [VALIDATION_CODE.STATE_DECORATOR]: () => 'States must be decorated with @State() decorator',
   [VALIDATION_CODE.INCORRECT_PRODUCTION]: () =>
-    'Angular is running in production mode but NGXS is still running in the development mode!\n' +
-    'Please set developmentMode to false on the NgxsModule options when in production mode.\n' +
+    'Angular is running in the production mode but NGXS is still running in the development mode!\n' +
+    'Please set developmentMode to false on the NgxsModule options in the production mode.\n' +
```
markwhitfeld: Revert the changes to both lines, it was correct before.

markwhitfeld: Oh, and `in the development mode` should rather be `in development mode`

arturovt: Not a native speaker and cannot argue, just saw such statement in Angular's code:
image
But no problem for sure!

arturovt: Done!

PR #1274, opened 2019-08-25
feat(store): represents a unique identifier for select decorator

splincode:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

## What is the new behavior?

Until the internal api (selectorId) was not hidden from developer, now we can have unique selectors inside the class instance

```
[ ] Yes
[x] No
```

### DIFF
--- a/packages/store/src/decorators/select/select.ts
+++ b/packages/store/src/decorators/select/select.ts
@@ -7,20 +7,13 @@ import { createSelectObservable, createSelectorFn } from './symbols';
 export function Select(rawSelector?: any, ...paths: string[]): PropertyDecorator {
   return function(target: any, propertyKey: string | symbol): void {
     const name: string = propertyKey.toString();
-    const selectorId = `__${name}__selector`;
+    const selectorId: unique symbol = Symbol(`__${name}__selector`);
     const selector = createSelectorFn(name, rawSelector, paths);
-    Object.defineProperties(target, {
-      [selectorId]: {
-        writable: true,
-        enumerable: false,
-        configurable: true
-      },
-      [name]: {
-        enumerable: true,
-        configurable: true,
-        get(): Observable<any> {
-          return this[selectorId] || (this[selectorId] = createSelectObservable(selector));
-        }
+    Object.defineProperty(target, name, {
+      enumerable: true,
+      configurable: true,
+      get(): Observable<any> {
+        return this[selectorId] || (this[selectorId] = createSelectObservable(selector));
       }
     });
   };
--- a/packages/store/tests/select.spec.ts
+++ b/packages/store/tests/select.spec.ts
@@ -83,36 +83,39 @@ describe('Select', () => {
     }
   });
 
-  it('should throw an exception when the component class is frozen', () => {
+  it('should get state even when the component class is frozen', () => {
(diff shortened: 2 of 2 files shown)


markwhitfeld:
What bug is this fixing? Is there an issue number?

splincode:
@markwhitfeld This is most likely a feature or refactor.

1. template reduction for selector decorator (improved bundle size)
2. type symbol is not enumerated by default, so you cannot access it outside
3. as you can see in tests, we cannot conflict with frozen objects or even redefine internal behavior

arturovt:
> @markwhitfeld This is most likely a feature or refactor.
> 
> 1. template reduction for selector decorator (improved bundle size)
> 2. type symbol is not enumerated by default, so you cannot access it outside
> 3. as you can see in tests, we cannot conflict with frozen objects or even redefine internal behavior

1. Yeah, less code.
2. Still it can be accessed via `Object.getOwnPropertySymbols`, but I guess it's not a problem.
3. Could this potentially be a breaking change? Just thinking aloud. I just remember the issue regarding `Object.entries` inside our `logger-plugin` where it broke IE10/IE11 apps.

I see at MDN that `Symbol` is not supported by IE at all.

packages/store/tests/select.spec.ts
```diff
@@ -251,18 +256,19 @@ describe('Select', () => {
...
-    try {
-      component.mySelect.subscribe();
-    } catch (e) {
-      error = e;
+    const keysFromComponent: string[] = [];
+
+    for (const key in component) {
+      // we need push all keys from instance without has own property
+      // noinspection JSUnfilteredForInLoop
+      keysFromComponent.push(key);
```
markwhitfeld: Couldn't this be achieved using `Object.keys(...)` instead?

packages/store/tests/select.spec.ts
```diff
@@ -241,6 +244,8 @@ describe('Select', () => {
       @Select((state: any) => state)
       public mySelect: Observable<string>;
 
+      constructor(public readonly store: Store) {}
```
markwhitfeld: What is this `store` reference needed for?

packages/store/tests/select.spec.ts
```diff
@@ -231,7 +234,7 @@ describe('Select', () => {
   }));
 
   it('should throw an exception if reserved key used in class', () => {
```
markwhitfeld: This description of the expectation does not match the assertion at the end anymore.

packages/store/src/decorators/select/select.ts
```diff
@@ -7,20 +7,13 @@ import { createSelectObservable, createSelectorFn } from './symbols';
 export function Select(rawSelector?: any, ...paths: string[]): PropertyDecorator {
   return function(target: any, propertyKey: string | symbol): void {
     const name: string = propertyKey.toString();
-    const selectorId = `__${name}__selector`;
+    const selectorId: unique symbol = Symbol(`__${name}__selector`);
```
markwhitfeld: `Symbol` is not used anywhere else in the library. Adding this will eliminate support for IE and there is no pollyfill that can be used.

PR #1276, opened 2019-08-25
test(store): whole check in simplePatch

splincode:
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

### DIFF
--- a/packages/store/tests/state.spec.ts
+++ b/packages/store/tests/state.spec.ts
@@ -348,6 +348,33 @@ describe('State', () => {
       expect(simplePatch(null!)(simple)).toEqual({ a: 1, b: 2 });
     });
 
+    it('should avoid the whole check if you want own properties only', () => {
+      function Origin() {
+        // @ts-ignore
+        this.x = `I'm an own property`;
+      }
+
+      Origin.prototype.y = `I'm not an own property`;
+
+      // @ts-ignore
+      const origin: any = new Origin();
+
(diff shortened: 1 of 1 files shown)


splincode:
> We should really only be patching object literals and not prototypal objects.
Yes, this test shows that we have a bug and therefore we need to fix it in the future

markwhitfeld:
The test is demonstrating that priories from the prototype would be copied. This is the behaviour I would expect if you did pass a class created in this way as the original.
What are you saying is the bug?

splincode:
@markwhitfeld @arturovt 
https://stackoverflow.com/a/32422910

We have non-obvious behavior

setState((state) => { ...state, value: 1 }); // shallow copy origin state
patchState({ value: 1 }); // deep copy origin state

markwhitfeld:
Ok, I think I am getting a better idea of the issue you are raising.
Doe this test currently pass?
The object spread operator should only copy the own properties as far as I understand.

splincode:
Current behavior

image

Expected behavior
image

markwhitfeld:
The test does not quite show this scenario that you describe now . 
`val` in the case of this code is the patch value being passed in, not the original value.
The original value is being spread. Your test would show that the original would lose any prototype properties because of the use if spread.

Issue #1264, opened 2019-08-23
🐞[BUG]: It looks like the migrations version is never set.


kayvanbree:
### Affected Package

@ngxs/storage-plugin

### Description

The code in https://github.com/ngxs/store/blob/master/packages/storage-plugin/src/storage.plugin.ts never seems to set a version number for a migration. It gets the version number from either `state.version` or `state[options.versionKey`. 

I think the design of version management in the migrations plugin may be a bit off. I suggest taking the index of the migration as the version and checking if that version is higher than `state.version`. At the moment it will migrate if it finds the same version, and it won't migrate if the version of the state is a different one (neither higher or lower).

## 🔬 Minimal Reproduction

- Create a new application with @ngxs/store and @ngxs/storage-plugin
- Create some kind of a state
- Create a migration with version 1
  - It won't ever set version 1 as the current migration in the state.

## Environment

I looked at the code at the master branch, and it's the same as in my environment. 

<pre><code>
Libs:
"@angular/animations": "^8.2.2",
    "@angular/cdk": "^8.1.2",
    "@angular/common": "~8.2.0",
    "@angular/compiler": "~8.2.0",
    "@angular/core": "~8.2.0",
    "@angular/forms": "~8.2.0",
    "@angular/material": "^8.1.2",
    "@angular/platform-browser": "~8.2.0",
    "@angular/platform-browser-dynamic": "~8.2.0",
    "@angular/router": "~8.2.0",
    "@ngxs/router-plugin": "^3.5.0",
    "@ngxs/storage-plugin": "^3.5.0",
    "@ngxs/store": "^3.5.0",
    "hammerjs": "^2.0.8",
    "rxjs": "~6.4.0",
    "tslib": "^1.10.0",
    "zone.js": "~0.9.1"

Browser:
- [x ] Chrome (desktop) version XX
- [ ] Chrome (Android) version XX
- [ ] Chrome (iOS) version XX
- [ ] Firefox version XX
- [ ] Safari (desktop) version XX
- [ ] Safari (iOS) version XX
- [ ] IE version XX
- [ ] Edge version XX
</code></pre>

kayvanbree:
Some more information: 

When I set the version of the first migration to 1 it won't ever run. If I set it to undefined it will run, but it will never set a version because its undefined. 

I just got one user yet so I will just push my new version without backward compatibility in the store, but this issue should be solved I think.

arturovt:
Wouldn't you want to make a PR?

[...]

PR #1295, opened 2019-08-30
fix: migrate based on higher versions in migrations

kayvanbree:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

You have to set the migration version of the next migration in the migrate function:

```
  migrations: [
    {
      version: 1,
      key: 'zoo',
      versionKey: 'myVersion',
      migrate: (state) => {
        return {
          newAnimals: state.animals,
          version: 2 // Important to set this to the next version!
        }
      }
    }
  ]
})
```

Also, looking at the tests, you will need to set the version in the default state. If you didn't before running the migrations plugin for the first time, it will never trigger the first migration.

Issue Number: #1264 

## What is the new behavior?

When there is no version in the state, the storage-plugin will assume version 0. This way it will do migrations even if you forgot to set one in your default state before running the plugin for the first time. 

It will also save the version of a migration to the state, so you don't have to do it in your migrate function:

```
  migrations: [
    {
      version: 1,
      key: 'zoo',
      versionKey: 'myVersion',
      migrate: (state) => {
        return {
          newAnimals: state.animals,
        }
      }
    }
  ]
})
```

```
[x] Yes
[ ] No
```

Because versions are now evaluated with `>`, version can now only be a number. Also, you will have to remove any code setting the version from migrate functions, and the next version you can add would be a version higher than the last one you set:

```
  migrations: [
    {
      version: 1,
      key: 'zoo',
      versionKey: 'myVersion',
      migrate: (state) => {
        return {
          newAnimals: state.animals,
          // version: 2 (this should be removed from every migrate function)
        }
      }
    },
    {
     version: 3,
     key: 'zoo',
      versionKey: 'myVersion',
      migrate: (state) => {
        // Your next migration
      }
    }
  ]
})
```

## Other information

A bug that still remains is that you can mess up your application by messing up the order of the migrations. That's why I thought of changing it even more, so it just uses the index of the migration as a version. Let me know if this would be something you'd like to see in this PR, else I will add a warning about the order to the docs.

### DIFF
--- a/packages/storage-plugin/src/symbols.ts
+++ b/packages/storage-plugin/src/symbols.ts
@@ -25,7 +25,7 @@ export interface NgxsStoragePluginOptions {
     /**
      * Version to key off.
      */
-    version: number | string;
+    version: number;
 
     /**
      * Method to migrate the previous state.
--- a/packages/storage-plugin/tests/storage.plugin.spec.ts
+++ b/packages/storage-plugin/tests/storage.plugin.spec.ts
@@ -164,7 +164,7 @@ describe('NgxsStoragePlugin', () => {
 
   it('should migrate global localstorage', () => {
     // Arrange
-    const data = JSON.stringify({ counter: { count: 100, version: 1 } });
+    const data = JSON.stringify({ counter: { count: 100 } });
     localStorage.setItem('@@STATE', data);
 
     // Act
@@ -178,8 +178,7 @@ describe('NgxsStoragePlugin', () => {
               versionKey: 'counter.version',
               migrate: (state: any) => {
                 state.counter = {
-                  counts: state.counter.count,
-                  version: 2
+                  counts: state.counter.count
                 };
                 return state;
               }
@@ -196,13 +195,13 @@ describe('NgxsStoragePlugin', () => {
 
     // Assert
     expect(localStorage.getItem('@@STATE')).toBe(
-      JSON.stringify({ counter: { counts: 100, version: 2 } })
+      JSON.stringify({ counter: { counts: 100, version: 1 } })
     );
   });
 
   it('should migrate single localstorage', () => {
     // Arrange
-    const data = JSON.stringify({ count: 100, version: 1 });
+    const data = JSON.stringify({ count: 100 });
     localStorage.setItem('counter', data);
 
     // Act
@@ -218,8 +217,7 @@ describe('NgxsStoragePlugin', () => {
               versionKey: 'version',
               migrate: (state: any) => {
                 state = {
-                  counts: state.count,
-                  version: 2
+                  counts: state.count
                 };
                 return state;
               }
@@ -235,7 +233,7 @@ describe('NgxsStoragePlugin', () => {
     store.selectSnapshot(CounterState);
 
     // Assert
-    expect(localStorage.getItem('counter')).toBe(JSON.stringify({ counts: 100, version: 2 }));
+    expect(localStorage.getItem('counter')).toBe(JSON.stringify({ counts: 100, version: 1 }));
   });
 
   it('should correct get data from session storage', () => {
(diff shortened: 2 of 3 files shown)


packages/storage-plugin/tests/storage.plugin.spec.ts
```diff
@@ -196,13 +195,13 @@ describe('NgxsStoragePlugin', () => {
...
     // Assert
     expect(localStorage.getItem('@@STATE')).toBe(
-      JSON.stringify({ counter: { counts: 100, version: 2 } })
+      JSON.stringify({ counter: { counts: 100, version: 1 } })
     );
   });
 
   it('should migrate single localstorage', () => {
     // Arrange
-    const data = JSON.stringify({ count: 100, version: 1 });
```
splincode: Do you have to change these tests? Can't add a new one?

packages/storage-plugin/src/symbols.ts
```diff
@@ -25,7 +25,7 @@ export interface NgxsStoragePluginOptions {
     /**
      * Version to key off.
      */
-    version: number | string;
```
splincode: it's breaking changes

kayvanbree: Yes this is a breaking change, but as of now if you install the plugin and don't have a version in your default state, it will save the default state without a version and then the check doesn't work. I changed it so that it looks for a higher version in the migrations array, for which I need a number.

packages/storage-plugin/tests/storage.plugin.spec.ts
kayvanbree: This is because of the breaking change.

packages/storage-plugin/src/symbols.ts
kayvanbree: I can change this PR to fix the bug (not finding a version in state prevents migrating) in another way and request a new feature if you'd prefer that.

splincode: You can use parseInt or Number for convert version key to integer in your code)

packages/storage-plugin/tests/storage.plugin.spec.ts
splincode: Unfortunately, we cannot allow critical changes.

packages/storage-plugin/src/symbols.ts
kayvanbree: But that would imply that you can use a string, when the comparison is on numbers.

packages/storage-plugin/tests/storage.plugin.spec.ts
kayvanbree: In bug fixes, or also for feature requests?

arturovt: @kayvanbree The semantic versioning assumes that breaking changes must be published with major release. But we can't easily switch to v4.

kayvanbree: Yeah I get that. Are there any plans or breaking changes scheduled for v4?

kayvanbree:
I'm trying to create a minimal reproduction on stackblitz but I can't get the local storage plugin to save anything: https://stackblitz.com/edit/storage-plugin-bug

packages/storage-plugin/tests/storage.plugin.spec.ts
splincode: https://github.com/ngxs/store/issues/827

splincode:
@kayvanbree Are you ready to continue working on the branch?

kayvanbree:
ATM I'm in between jobs, I think I can take a look tomorrow!

markwhitfeld:
Hi @kayvanbree. I hope the transition to the new job has been going smoothly.
The first few weeks are always very busy.
I was wondering if you will get a chance to work on this soon? 
We hope to release the next version before the end of the week and it would be great to get this in.

splincode:
I close because there is no activity

PR #1353, opened 2019-09-28
refactor(store): simplePatch

splincode:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

## What is the new behavior?

This commit updates, where the for-in loop was used, resulting in copying from prototype to own properties, thus consuming more memory. Prior to NodeJS 12 (V8 versions before 7.4) there was an optimization that was improving the situation and since that logic was removed in favor of other optimizations, the situation with memory consumption caused by the for-in loop got worse. This commit adds a check to make sure we copy only own properties over to cloned object.

```
[ ] Yes
[x] No
```


arturovt:
Could you, please, provide benchmarks that it’s really faster on most browsers? I’m onto phone and I cannot check it.
Also you’re talking about nodejs12, but our code is running in browser. Could you also, please, create a spec that shows that for-in consumes a lot of memory? I started googling and didn’t find anything unfortunately:(

splincode:
http://jsben.ch/x4Ig4

image

I understand that these are not huge results, however, we can still help improve performance

arturovt:
image

splincode:
When you're using for (var key in obj) it will loop through the given object + its parent objects' properties on the prototype chain until it reaches the end of the chain. As you want to check only specific object's properties, you need to use hasOwnProperty.

arturovt:
I think this is micro optimization. As you see I just gave you the screen from my phone.
We have to focus mobile devices first, not PCs.

splincode:
@markwhitfeld updated

splincode:
I recently saw in our company how people do this and it terrified me, so I decided to make this pull request

```ts
ctx.patchState(new MyClass('hello world'));
```

packages/store/src/internal/state-operators.ts
```diff
@@ -1,19 +1,17 @@
...
-    if (isPrimitive) {
-      throw new Error('Patching primitives is not supported.');
-    }
-    const newState = { ...(<any>existingState) };
-    for (const k in val) {
-      newState[k] = val[k];
-    }
-    return <T>newState;
+
+    return { ...existingState, ...val };
```
markwhitfeld: Technically this change has the potential to be a breaking change for someone....?

packages/store/tests/state.spec.ts
```diff
@@ -348,20 +348,45 @@ describe('State', () => {
...
+      expect(resultState.y).toEqual(undefined);
+    });
+
     it('should throw exception if value is array', () => {
       try {
         const simple: string[] = ['hello'];
         simplePatch(['world'])(simple);
       } catch (e) {
-        expect(e.message).toEqual('Patching arrays is not supported.');
+        expect(e.message).toEqual(MESSAGES[CODE.PATCHING_ARRAY]());
```
markwhitfeld: Tests should still be explicit about the message. It should not look up the expected value.
```suggestion
        expect(e.message).toEqual('Patching arrays is not supported.');
```

splincode: image

I see that we adhere to the same errors everywhere for verification

markwhitfeld: This is an anti-pattern in tests that we should avoid.
Your expectation should always be constant and not calculated.

packages/store/src/internal/state-operators.ts
splincode: not sure, so rolled back

packages/store/src/internal/state-operators.ts
```diff
@@ -1,19 +1,24 @@
...
-      throw new Error('Patching primitives is not supported.');
-    }
-    const newState = { ...(<any>existingState) };
-    for (const k in val) {
-      newState[k] = val[k];
+
+    const newState: any = { ...(existingState as any) };
+    for (const key in val) {
+      // deep clone for patch compatibility
+      // noinspection JSUnfilteredForInLoop (IDE)
```
arturovt: Could you remove this please?

splincode: my IDEA (WebStorm) swears red constantly

PR #1356, opened 2019-09-29
build(ci): add benchmark

splincode:
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[x] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

## What is the new behavior?

for future refactoring, we will be able to set our limits at which we guarantee how quickly our functions are performed

```
[ ] Yes
[x] No
```

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -169,6 +169,15 @@ jobs:
           command: yarn bundlesize
       # - *persist_coverage // we don't run code coverage for integration tests
 
+  benchmark:
+    <<: *job_defaults
+    steps:
+      - *attach_workspace
+      - run:
+          name: Benchmark
+          command: yarn benchmark
+      # - *persist_coverage // we don't run code coverage for integration tests
+
   upload_coverage:
     <<: *job_defaults
     environment:
@@ -261,6 +270,7 @@ workflows:
             - integration_ng5_tests
             - integration_test_types
             - bundlesize
+            - benchmark
 
       # Publish @dev builds if no tag is specified
       - publish_dev_build_to_npm:
@@ -277,6 +287,7 @@ workflows:
             - integration_ng5_tests
             - integration_test_types
             - bundlesize
+            - benchmark
 
       # Publish package.version @latest when a v.* git tag is present
       - publish_tagged_build_to_npm:
@@ -292,3 +303,4 @@ workflows:
             - integration_ng5_tests
             - integration_test_types
             - bundlesize
+            - benchmark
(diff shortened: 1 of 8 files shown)


splincode:
image

splincode:
Actually, I think I need a docker here.

markwhitfeld:
We would need a measure that would not depend on the machine running the tests.
I don't think that this exists. 

Docker would not solve this because it specifies the operating system, not the actual hardware.
I would love to hear @arturovt 's thoughts on this!

arturovt:
Dear Max and Mark, I’ll be back to the city in 2 days from vacation so I’ll be able to leave a comment here 🤗😝

splincode:
https://github.com/angular/angular/blob/master/modules/benchmarks/e2e_test/old/costs_perf.ts#L17

https://github.com/angular/angular/tree/master/packages/benchpress

We can study this package ```@angular/benchpress```

arturovt:
CircleCI jobs are already running in Docker containers using specified images, thus this won't do the trick.

---

Just pasting here the quote of Donald Knuth:

> The real problem is that programmers have spent far too much time worrying about efficiency in the wrong places and at the wrong times; premature optimization is the root of all evil (or at least most of it) in programming.

Before adding any benchmarks we have to understand WHAT we want to benchmark+optimize and what is actually slow, and how could we gain the performance. The requirements should be developed and the approriate document has to be prepared. It has to consist of:

* Requirements engineering
* Design document
    * What, why and how?
        * some parts of NGXS are slow at such moments...
        * users of our library noticed such and such pitfalls...
    * High-level overview of slow NGXS code
        * let's look at those functions that don't seem to be performant at Chrome/Safari/Mozilla...
* Implementation details
* Summary

---

I wouldn't want  to write benchmarks for every function that used internally like `isObject` or `topologicalSort`. IMHO it's a waste of time. **We have to focus bugs first.**

---

Fix me if I'm wrong. I'm open to discussion.

Also the deepest pitfall is mobile devices. They are way slower then PCs. Unfortunately I haven't found any service that lets run performance tests on mobile devices.

Eventually these benchmarks will depend on CircleCI machines and in addition they will be run inside isolated containers and I wouldn't believe Docker in that case. Docker has its own file/memory management strategies.

> Docker processes compete for all kinds of kernel resources.

So there are too many things to discuss here.

splincode:
what do you think about  @angular/benchpress?

markwhitfeld:
I agree with @arturovt. We are not having perf issues.
My concern with this (although a great idea in concept) is that it will result in fragile tests that do not really add any value.
Can we close this PR as unmerged?

Issue #803, opened 2019-01-30
🐞[BUG]: ErrorHandler triggered twice on one action


piernik:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[x] Bug report  
[ ] Performance issue
[ ] Feature request
[ ] Documentation issue or request
[ ] Support request => https://github.com/ngxs/store/blob/master/CONTRIBUTING.md
[ ] Other... Please describe:
</code></pre>

## Current behavior
I have action that retrives data from backend and throws error.
I run action like this:
```
this.obiektyActions.change(data)
          .subscribe(() => {
             this.changed.emit(id);
        });
```
Action itself is something like this:
```
return this.apiService.change(action.data)
      .pipe(
        catchError((err) => {
// ...
          return throwError(err);
        }),
        tap(resp => {
          // ...
          this.patchCurrentObiekt(ctx, action.obiektId, val);
        })
      );
```

I have my custom `ErrorHandler` and it is invoked twice. As I read it's because observable has two or more subscribers: https://stackoverflow.com/questions/47898101/why-my-global-error-handler-get-invoked-twice-in-my-angular-app

I suppose You are subscribing to action, so I get two subscriptions (and ErrorHandler is triggered twice).

If I change action execution to:
```
this.obiektyActions.change(data)
          .pipe(tap(() => {
             this.changed.emit(id);
        }));
```
Action is still triggered (so You have to subscribe) but on success `this.changed.emit(id);` is never triggered.

Edit:
`this.obiektyActions.change(data)` = `return this.store.dispatch(new ChangeDataAction(data));`
## Expected behavior
ErrorHandler should recive error info only once.

## Minimal reproduction of the problem with instructions
https://stackblitz.com/edit/ngxs-error-throw?file=src/app/app.component.ts

## Environment

<pre><code>
Libs:
- @angular/core version: 7.2.2
- @ngxs/store version: 3.3.4

Browser:
- [ ] Chrome (desktop) version XX
- [ ] Chrome (Android) version XX
- [ ] Chrome (iOS) version XX
- [ ] Firefox version XX
- [ ] Safari (desktop) version XX
- [ ] Safari (iOS) version XX
- [ ] IE version XX
- [ ] Edge version XX
 
For Tooling issues:
- Node version: XX  
- Platform:  

Others:

</code></pre>

splincode:
create stackblitz example

piernik:
[...]

Issue #781, opened 2019-01-18
Error caught after dispatch is still logged


axelboc:
## I'm submitting a...

```
[ ] Regression (a behavior that used to work and stopped working in a new release)
[x] Bug report
[ ] Performance issue
[ ] Feature request
[ ] Documentation issue or request
[ ] Support request => https://github.com/ngxs/store/blob/master/CONTRIBUTING.md
[ ] Other...
```

## Current behavior

I have an action A with an action handler that throws an error (either with `throw` or RxJS' `throwError`). When I dispatch action A, then pipe it with `catchError`, the subscriber's `error` callback is not called as expected, but the error **still gets logged to the console**.

```ts
store
  .dispatch(new ActionThatThrows())
  .pipe(catchError(() => of('foo')))
  .subscribe();
```

## Expected behavior

The error should not be logged to the console, since it is properly caught.

When I remove the `dispatch` and throw the error directly, the error is not logged (as expected):

```ts
throwError(new Error('my oh my'))
  .pipe(catchError(() => of('foo')))
  .subscribe();
```

## Minimal reproduction of the problem with instructions

- https://stackblitz.com/edit/ngxs-simple-uagu8f
- Open the console and click on "Click Me".
- The error is logged, followed by `next foo` and `complete`.

## Environment

```
Libs:
- @angular/core version: 7.2.0
- @ngxs/store version: 3.3.4-dev.master-fa8d5a1

Browser:
- [ ] Chrome (desktop) version XX
- [ ] Chrome (Android) version XX
- [ ] Chrome (iOS) version XX
- [x] Firefox version 64.0.2
- [ ] Safari (desktop) version XX
- [ ] Safari (iOS) version XX
- [ ] IE version XX
- [ ] Edge version XX
```

eranshmil:
This isn't related to NGXS, you should implement an ErrorHandler to catch this kind of exceptions.

axelboc:
My point is that the `ErrorHandler` is not supposed to be called at all, since the error is supposed to have been caught with `catchError`. It works fine without the `dispatch`, which makes me think it's related to NGXS.

eranshmil:
The observable returned from the dispatch isn't the same one as returned in the state.
Please read this comment, that might be related.
https://github.com/ngxs/store/issues/606#issuecomment-435174322

axelboc:
[...]

Issue #463, opened 2018-06-25
🐞[BUG]: Global error handler getting called twice


leksinteractive:
Hi team,

It seems the global error handler is invoked twice when it first catches an error that's thrown by a state action. Any subsequent errors seem to trigger it only once though. This is the behaviour if I re-throw the error in the global error handler like so:

```
public handleError(error: any): void {
      console.log('GlobalErrorHandler', error);
      throw error;
}
```

To make things more interesting, if I don't re-throw the error, the handleError method gets invoked twice on every error (regardless of whether it's the first one or subsequent ones). Any suggestions for getting around this? Also, as in my app I do some redirection in the global error handler eg. for 401 errors, so I wouldn't want to re-throw the error in that case.

Here's a stackblitz: https://github-qfjnen-drrq77.stackblitz.io
And the editable version: https://stackblitz.com/edit/github-qfjnen-drrq77

Cheers,
Aleks

(Loosely relates to: #372)

markwhitfeld:
**Copied from other issue conversation:**
I think that it could be to do with shareReplay (which I think has since also been added to the inner dispatch, within the compose). In PR #376 I opted for a subscribe over a catchError pipe because shareReplay only stores the last replay result of the value in the stream and not an error. Using a shareReplay over a catchError may result in the handler being executed twice.

markwhitfeld:
@leksinteractive Do you want to have a go at reproducing this with a test? 
This scenario seems to be covered with this test: https://github.com/ngxs/store/blob/07de0badfbbd78c1de2f53f08de78445fa5f6b7d/packages/store/tests/dispatch.spec.ts#L22-L58
A repro with a test would make fixing it so much easier :wink:

markwhitfeld:
Hmmm... this may actually have something to do with Zones

leksinteractive:
Thanks for looking into this @markwhitfeld - I'd be happy to put together that test as soon as I find some time. Zones, I think I'm a bit out of my depth there unfortunately :)

amcdnl:
@markwhitfeld - were u able to repo this?

PR #1379, opened 2019-10-10
fix(store): do not re-throw error to the global handler if custom is provided

arturovt:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: #1145 #803 #781 #463 

```
[ ] Yes
[x] No
```

## Other information

:warning: :warning: **The whole PR description is below** :warning: :warning: 

This PR is a fix for old issues that reported about error handlers that were invoked twice. Let's look at the runtime behavior below:

There is a simple `Increment` action and such state:

```ts
@State({
  name: 'counter',
  defaults: 0
})
export class CounterState {
  @Action(Increment)
  increment() {
    throw new EvalError('This is an eval error...');
  }
}
```

And custom error handler looks as follows:

```ts
@Injectable({ providedIn: 'root' })
export class NgxsErrorHandler implements ErrorHandler {
  handleError(error: Error): void {
    console.log(`${NgxsErrorHandler.name}.handleError has been invoked with ${error.name}`);

    // re-throw
    throw error;
  }
}
```

---

What's actually happening there? We dispatch an action. Our action handler throws an `EvalError`. NGXS subscribes under the hood and provides `onError` listener where it does:

```ts
this._errorHandler.handleError(error);
```

NGXS invokes our custom `NgxsErrorHandler.handleError`, it logs to console and re-throws error via `throw error` (the pitfall is here). As this error is re-thrown - it gets caught by the Angular zone. When Angular forks zone it provides the `onHandleError` hook. This looks as follows:

```ts
onHandleError: (delegate: ZoneDelegate, current: Zone, target: Zone, error: any): boolean => {
  delegate.handleError(target, error);
  zone.runOutsideAngular(() => zone.onError.emit(error));
  return false; 
}
```

Zone intercepts all thrown exceptions inside `zone.run` and emits event via `onError.emit(error)`. Is this EventEmitter subscribed anywhere? Yes, when we invoke `bootstrapModule` Angular retrieves `ErrorHandler` under the hood and subscribes to the `onError`:

```ts
const exceptionHandler = moduleRef.injector.get(ErrorHandler, null);
if (!exceptionHandler) {
  throw new Error(
    'No ErrorHandler. Is platform module (BrowserModule) included?'
  );
}

ngZone.runOutsideAngular(() =>
  ngZone.onError.subscribe({
    next: error => {
      exceptionHandler.handleError(error);
    }
  })
);
```

:exclamation: **As you see that's why our message is logged twice, because `handleError` is invoked by NGXS and Angular** :exclamation: 

zone

---

**You may ask me why there are no tests?**

Because runtime behavior differs from testing behavior. The zone, that is used for testing, doesn't intercept errors like at runtime. That's why I cannot add tests. But as you see all tests are passed thus it's not a breaking change. The new behavior looks as follows:

fixed

---

**The same test I've added BUT without `try/catch` block, as you see below the `handleError` does re-throws it to the zone.js** :point_down: 

without

### DIFF
--- a/packages/store/src/internal/dispatcher.ts
+++ b/packages/store/src/internal/dispatcher.ts
@@ -39,7 +39,11 @@ export class InternalDispatcher {
 
     result.subscribe({
       error: error =>
-        this._ngxsExecutionStrategy.leave(() => this._errorHandler.handleError(error))
+        this._ngxsExecutionStrategy.leave(() => {
+          try {
+            this._errorHandler.handleError(error);
+          } catch {}
+        })
     });
 
     return result.pipe(leaveNgxs(this._ngxsExecutionStrategy));
--- a/packages/store/tests/dispatch.spec.ts
+++ b/packages/store/tests/dispatch.spec.ts
@@ -55,7 +55,7 @@ describe('Dispatch', () => {
     expect(observedCalls).toEqual(['handleError(...)', 'observer.error(...)']);
   }));
 
-  fit('should not propagate an unhandled exception', () => {
+  it('should not propagate an unhandled exception', () => {
     // Arrange
     const message = 'This is an eval error...';
 
--- a/packages/store/tests/dispatch.spec.ts
+++ b/packages/store/tests/dispatch.spec.ts
@@ -3,12 +3,8 @@ import { async, fakeAsync, TestBed, tick } from '@angular/core/testing';
 import { of, throwError, timer } from 'rxjs';
 import { delay, skip, tap } from 'rxjs/operators';
 
-import { State } from '../src/decorators/state';
-import { Action } from '../src/decorators/action';
-import { Store } from '../src/store';
-import { NgxsModule } from '../src/module';
-import { StateContext } from '../src/symbols';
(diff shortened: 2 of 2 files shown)


splincode:
Update docs, tests?

arturovt:
> Update docs, tests?

Max, honestly saying I'm very upset that you don't read PR description :cry: :cry: :(

markwhitfeld:
@arturovt, I read your PR description and I would like to frame it and put it up on the wall!
Very nicely done.

arturovt:
Dear Mark and Max, I will try to find a workaround to test this issue. I really want to have tests, believe me, I’m not that person who doesn’t write tests. TDD is our life 😂. This is one of the most nonstandard issues.

markwhitfeld:
If you want to try to write a test then you could create a fake execution strategy and just check that the callback given to the `leave`  call doesn't propogate an unhandled exception.

arturovt:
Updated PR description, see the bottom of it.

Issue #749, opened 2019-01-06
🚀[FEATURE]: new lifecycle API


splincode:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[ ] Performance issue
[x] Feature request
[ ] Documentation issue or request
[ ] Support request => https://github.com/ngxs/store/blob/master/CONTRIBUTING.md
[ ] Other... Please describe:
</code></pre>

## Current behavior

ngxsOnInit() - before APP_INITIALIZER

## Expected behavior

Lifecycle sequence

| Hook  | Purpose and Timing  |
|---|---|
| ngxsOnChanges() | The method receives a SimpleChanges object of current and previous property values. This is convenient if we want to dispatch any additional actions when any fields have changed. Called before ngxsOnInit() and whenever state change. |
|  ngxsOnInit() |  If a state implements the NgxsOnInit interface, its ngxsOnInit method will be invoked after all the states from the state's module definition have been initialized and pushed into the state stream. The states' ngxsOnInit methods are invoked in a topological sorted order going from parent to child. The first parameter is the StateContext where you can get the current state and dispatch actions as usual. |
| ngxsAfterBootstrap() | Respond after Angular root component's  views and child views bootstraped. It is convenient if we need to load the configuration before using some services in the states. |

Transient states

| Hook  | Purpose and Timing  |
|---|---|
| ngxsOnDestroy() | Destroy when destroy component |

PR #1389, opened 2019-10-14
feat(store): implements ngxsOnChanges

splincode:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: https://github.com/ngxs/store/issues/749

## What is the new behavior?

The method receives a NgxsSimpleChanges object of current and previous property values. This is convenient if we want to dispatch any additional actions when any fields have changed. Called whenever state change.

The main motivation is that this is a convenient hook in order to save states on the server, if we need it. We also have the opportunity to call any third-party services directly from the method, without the need to subscribe to changes to the state.

```
[ ] Yes
[x] No
```

### DIFF
--- a/packages/store/src/internal/lifecycle-state-manager.ts
+++ b/packages/store/src/internal/lifecycle-state-manager.ts
@@ -5,7 +5,7 @@ import { filter, mergeMap, tap } from 'rxjs/operators';
 import { StateContextFactory } from './state-context-factory';
 import { InternalStateOperations } from './state-operations';
 import { getStateDiffChanges, MappedStore, StatesAndDefaults } from './internals';
-import { LifecycleHooks, NgxsLifeCycle, NgxsSimpleChanges } from '../symbols';
+import { LifecycleHooks, NgxsLifeCycle, NgxsSimpleChange } from '../symbols';
 
 @Injectable()
 export class LifecycleStateManager {
@@ -50,7 +50,7 @@ export class LifecycleStateManager {
           .getRootStateOperations()
           .getState();
 
-        const firstDiffChanges: NgxsSimpleChanges = getStateDiffChanges(metadata, {
+        const firstDiffChanges: NgxsSimpleChange = getStateDiffChanges(metadata, {
           currentAppState,
           newAppState
         });
--- a/packages/store/src/internal/lifecycle-state-manager.ts
+++ b/packages/store/src/internal/lifecycle-state-manager.ts
@@ -5,7 +5,7 @@ import { filter, mergeMap, tap } from 'rxjs/operators';
 import { StateContextFactory } from './state-context-factory';
 import { InternalStateOperations } from './state-operations';
 import { getStateDiffChanges, MappedStore, StatesAndDefaults } from './internals';
-import { LifecycleHooks, NgxsLifeCycle, NgxsSimpleChange } from '../symbols';
+import { NgxsLifeCycle, NgxsSimpleChange, StateContext } from '../symbols';
 
 @Injectable()
 export class LifecycleStateManager {
@@ -25,25 +25,16 @@ export class LifecycleStateManager {
         mergeMap(() => this.bootstrapper.appBootstrapped$),
         filter(appBootstrapped => !!appBootstrapped)
       )
-      .subscribe(() => {
-        this.invokeBootstrap(results!.states);
-      });
+      .subscribe(() => this.invokeBootstrap(results!.states));
   }
 
   /**
    * Invoke the init function on the states.
    */
   invokeInit(stateMetadatas: MappedStore[]): void {
-    this.invokeOnChanges(stateMetadatas);
-    this.invokeLifecycleHooks(stateMetadatas, LifecycleHooks.NgxsOnInit);
-  }
-
-  /**
-   * Invoke the on changes function on the states.
-   */
-  invokeOnChanges(stateMetadatas: MappedStore[]): void {
     for (const metadata of stateMetadatas) {
       const instance: NgxsLifeCycle = metadata.instance;
+
       if (instance.ngxsOnChanges) {
         const currentAppState: PlainObject = {};
         const newAppState: PlainObject = this.internalStateOperations
@@ -55,7 +46,11 @@ export class LifecycleStateManager {
           newAppState
         });
 
-        instance.ngxsOnChanges!(firstDiffChange);
+        instance.ngxsOnChanges(firstDiffChange);
+      }
+
+      if (instance.ngxsOnInit) {
+        instance.ngxsOnInit(this.getStateContext(metadata));
       }
     }
   }
@@ -64,17 +59,15 @@ export class LifecycleStateManager {
    * Invoke the bootstrap function on the states.
    */
   invokeBootstrap(stateMetadatas: MappedStore[]) {
-    this.invokeLifecycleHooks(stateMetadatas, LifecycleHooks.NgxsAfterBootstrap);
-  }
-
-  private invokeLifecycleHooks(stateMetadatas: MappedStore[], hook: LifecycleHooks): void {
     for (const metadata of stateMetadatas) {
       const instance: NgxsLifeCycle = metadata.instance;
-
-      if (instance[hook]) {
-        const stateContext = this.stateContextFactory.createStateContext(metadata);
-        instance[hook]!(stateContext);
+      if (instance.ngxsAfterBootstrap) {
+        instance.ngxsAfterBootstrap(this.getStateContext(metadata));
       }
     }
+  }
+
+  private getStateContext(metadata: MappedStore): StateContext<any> {
+    return this.stateContextFactory.createStateContext(metadata);
   }
 }
--- a/packages/store/src/internal/lifecycle-state-manager.ts
+++ b/packages/store/src/internal/lifecycle-state-manager.ts
@@ -50,12 +50,12 @@ export class LifecycleStateManager {
           .getRootStateOperations()
           .getState();
 
-        const firstDiffChanges: NgxsSimpleChange = getStateDiffChanges(metadata, {
+        const firstDiffChange: NgxsSimpleChange = getStateDiffChanges(metadata, {
           currentAppState,
           newAppState
         });
 
-        instance.ngxsOnChanges!(firstDiffChanges);
+        instance.ngxsOnChanges!(firstDiffChange);
       }
     }
   }
--- a/packages/store/src/internal/lifecycle-state-manager.ts
+++ b/packages/store/src/internal/lifecycle-state-manager.ts
@@ -52,6 +52,8 @@ export class LifecycleStateManager {
       if (instance.ngxsOnInit) {
         instance.ngxsOnInit(this.getStateContext(metadata));
       }
+
+      metadata.isInitialised = true;
     }
   }
 
--- a/packages/store/src/internal/lifecycle-state-manager.ts
+++ b/packages/store/src/internal/lifecycle-state-manager.ts
@@ -1,11 +1,11 @@
 import { Injectable } from '@angular/core';
-import { NgxsBootstrapper } from '@ngxs/store/internals';
-import { filter, tap, mergeMap } from 'rxjs/operators';
+import { NgxsBootstrapper, PlainObject } from '@ngxs/store/internals';
+import { filter, mergeMap, tap } from 'rxjs/operators';
 
 import { StateContextFactory } from './state-context-factory';
 import { InternalStateOperations } from './state-operations';
-import { MappedStore, StatesAndDefaults } from './internals';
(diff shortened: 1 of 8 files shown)


splincode:
Why pending?

image

splincode:
@markwhitfeld This API unifies the life cycle, thereby making our states universal and similar to Angular components. 

https://angular.io/guide/lifecycle-hooks

Also there are many reasons to add this functionality, so it is necessary in the following use cases:

- *A convenient way to track state changes*

Before

```ts
@State({ })
class MyState {
}
```

```ts
@Component({ })
class MyComponent {
  constructor(store: Store) {
    store.select(MyState).subscribe((newState) => {
       console.log('state is changed', newState);
    })
  }
}
```

One of the problems is, if we do not use the logging plugin (the official example does not use it), then we do not know what the previous state was before our state changed. It's great to have such an opportunity out of the box.

After

```ts
@State({ })
class MyState {
  ngxsOnChanges({  previousValue,  currentValue }: NgxsSimpleChange) {
     console.log('state is changed', previousValue, currentValue);
  }
}
```

Very comfortably! 

Note: Why `NgxsSimpleChange`, again for the sake of versatility.

https://angular.io/api/core/SimpleChange

- *Convenient to synchronize with the server*

In my work when using state, I needed to save state on the server every time the client changed it.


Before

```ts
@State({ })
class MyState {
}
```

```ts
@Component({ })
class MyComponent {
  constructor(store: Store, api: ApiService) {
    store.select(MyState).subscribe(async (newState) => {
       await api.saveToRedis(newState); 
    })
  }
}
```

After

```ts
@State({ })
class MyState {
  constructor(api: ApiService) {}

  async ngxsOnChanges({ currentValue: newState }: NgxsSimpleChange) {
     await api.saveToRedis(newState);
  }
}
```

- *You can write your own custom logger without another plugins*

```ts
@State({ })
class MyState {
  ngxsOnChanges({ previousValue,  currentValue }: NgxsSimpleChange) {
     console.log('prev state', previousValue); // you can add color
     console.log('next state', currentValue);   // you can add color
  }
}
```

Or 

```ts
@State({ })
class MyState {
  constructor(myLogger: MyLoggerService) {}
  ngxsOnChanges(change: NgxsSimpleChange) {
     myLogger.prettyDiffChanges(change);
  }
}
```

This will allow users to independently control logging without complaining about bugs of the logging plugin (#561, #276, #213). 

----------------

Why `ngxsOnChanges(change: NgxsSimpleChange) {}` instead `ngxsOnChanges(change: NgxsSimpleChange, ctx: StateContext<any>) {}`?

One of the main reasons is that users should not manipulate state in this hook! Therefore, we should not motivate them to do this!

If that were so

```
@State({ })
class MyState {
  ngxsOnChanges(_: NgxsSimpleChange, ctx: StateContext<any>) {
     ctx.patchState({  onChange: true }); // Error: Maximum call stack size exceeded error
  }
}
```

Because the hook is called before setting the state, but if we allow the user to do in this method, then this will lead to various kinds of errors.

However, now users can make such mistakes.

```
@State({
   name: 'counter',
   defaults: 0
})
class MyState {

  ngxsOnChanges() {
     this.store.dispatch({ type: 'increment' }); // Error: Maximum call stack size exceeded error
  }

   @Action({ type: 'increment' })
    increment({ setState }: StateContext<number>) {
      setState((state: number) => ++state);
    }
}
```

In the future I will add verification so that users do not make such errors.

packages/store/src/internal/state-context-factory.ts
```diff
@@ -27,6 +27,18 @@ export class StateContextFactory {
 
     function setStateValue(currentAppState: any, newValue: T): any {
       const newAppState = setValue(currentAppState, metadata.depth, newValue);
+
+      if (metadata.instance && metadata.instance.ngxsOnChanges) {
+        const instance: NgxsLifeCycle = metadata.instance;
```
markwhitfeld: If you move the `instance` variable outside the `if` block then it becomes simpler and you won't need the `!` later on the invocation of the `instance.ngxsOnChanges(change)` because TypeScript can now infer it from the check in the condition.
```TS
      const instance: NgxsLifeCycle = metadata.instance;
      if (instance && instance.ngxsOnChanges) {        
```

packages/store/src/internal/state-context-factory.ts
```diff
@@ -27,6 +27,18 @@ export class StateContextFactory {
...
+        instance.isFirstChange = true;
+        const change: NgxsSimpleChange = getStateDiffChanges<T>(metadata, {
+          currentAppState,
+          newAppState
+        });
+
+        instance.ngxsOnChanges!(change);
```
markwhitfeld: See above
```suggestion
        instance.ngxsOnChanges(change);
```

packages/store/src/symbols.ts
```diff
@@ -132,9 +135,8 @@ export interface StoreOptions<T> {
   children?: any[];
 }
 
-export const enum LifecycleHooks {
```
markwhitfeld: Glad to get rid of this enum ;-)

markwhitfeld:
Once @arturovt is happy then we can merge and add to the 3.6 release.

markwhitfeld:
cc @joaqcid I would like to hear your thoughts here too...

packages/store/src/public_api.ts
```diff
@@ -25,7 +25,9 @@ export {
   StateOperator,
   NgxsOnInit,
   NgxsAfterBootstrap,
-  NgxsModuleOptions
+  NgxsOnChanges,
+  NgxsModuleOptions,
+  NgxsSimpleChange
```
markwhitfeld: Add trailing comma. Makes futher changes easier to merge.

packages/store/src/symbols.ts
```diff
@@ -144,6 +146,14 @@ export interface NgxsOnInit {
   ngxsOnInit(ctx?: StateContext<any>): void | any;
 }
 
+/**
+ * On change interface
+ */
+export interface NgxsOnChanges {
+  isFirstChange?: boolean;
```
markwhitfeld: What do you think? Should we make this match the angular interface by calling this prop `firstChange`?
https://angular.io/api/core/SimpleChange

markwhitfeld: I don't see the need for the function that they have on their interface other than encapsulation.

splincode: isFirstChange() - method not needed

markwhitfeld: yeah, but we have named our property(`isFirstChange`) after their method(`isFirstChange`), not after the Angular property(`firstChange`).

packages/store/src/public_api.ts
splincode: Prettier remove before pushed

packages/store/src/symbols.ts
splincode: We can name this property differently, it is necessary to track changes.

markwhitfeld: Yeah, I was just wondering if it was the intention to make it as close to the Angular interface as possible?

packages/store/src/public_api.ts
markwhitfeld: Meh

packages/store/src/internal/state-context-factory.ts
markwhitfeld: @splincode Are you going to do this tweak?

joaqcid:
> cc @joaqcid I would like to hear your thoughts here too...

i like this hook! i think it can be very handy for the scenarios described

packages/store/src/internal/state-context-factory.ts
```diff
@@ -27,6 +27,18 @@ export class StateContextFactory {
 
     function setStateValue(currentAppState: any, newValue: T): any {
       const newAppState = setValue(currentAppState, metadata.depth, newValue);
+
+      if (metadata.instance && metadata.instance.ngxsOnChanges) {
+        const instance: NgxsLifeCycle = metadata.instance;
+        instance.isFirstChange = true;
```
joaqcid: i find this a bit confusing, since we are setting it to true, when its actually gonna be false

splincode: it should always be true after ngxsOnInit

packages/store/src/internal/lifecycle-state-manager.ts
```diff
@@ -25,33 +25,51 @@ export class LifecycleStateManager {
...
 
   /**
    * Invoke the init function on the states.
    */
   invokeInit(stateMetadatas: MappedStore[]): void {
-    this.invokeLifecycleHooks(stateMetadatas, LifecycleHooks.NgxsOnInit);
+    for (const metadata of stateMetadatas) {
+      const instance: NgxsLifeCycle = metadata.instance;
+
+      if (instance.ngxsOnChanges) {
```
markwhitfeld: Can we safely assume that instance is not null here?

packages/store/src/symbols.ts
```diff
@@ -132,9 +135,15 @@ export interface StoreOptions<T> {
...
-  NgxsAfterBootstrap = 'ngxsAfterBootstrap'
+/**
+ * Represents a basic change from a previous to a new value for a single state instance.
+ * Passed as a value in a NgxsSimpleChanges object to the ngxsOnChanges hook.
+ */
+export class NgxsSimpleChange<T = any> {
+  constructor(public previousValue: T, public currentValue: T, public firstChange: boolean) {}
+  public isFirstChange(): boolean {
+    return this.firstChange;
+  }
```
markwhitfeld: Does it need this method?
I feel like the `firstChange` property is good enough. What do you think?

splincode: https://angular.io/api/core/SimpleChange#simplechange

Do we want to repeat this interface?

packages/store/src/internal/lifecycle-state-manager.ts
splincode: Yes

image

packages/store/src/symbols.ts
markwhitfeld: Not necessarily, but we can take inspiration from it.
I think it is all good, but that method is just a bit strange to have there if you already expose a prop.

packages/store/src/internal/lifecycle-state-manager.ts
markwhitfeld: 👍

packages/store/src/symbols.ts
splincode: done

packages/store/src/internal/state-context-factory.ts
splincode: done

packages/store/src/symbols.ts
```diff
@@ -132,9 +135,12 @@ export interface StoreOptions<T> {
...
 
-export const enum LifecycleHooks {
-  NgxsOnInit = 'ngxsOnInit',
-  NgxsAfterBootstrap = 'ngxsAfterBootstrap'
+/**
+ * Represents a basic change from a previous to a new value for a single state instance.
+ * Passed as a value in a NgxsSimpleChanges object to the ngxsOnChanges hook.
+ */
+export class NgxsSimpleChange<T = any> {
+  constructor(public previousValue: T, public currentValue: T, public firstChange: boolean) {}
```
markwhitfeld: Just in case someone feels tempted to mess with these properties ;-)
```suggestion
  constructor(public readonly previousValue: T, public readonly currentValue: T, public readonly firstChange: boolean) {}
```

packages/store/src/symbols.ts
```diff
@@ -144,6 +150,13 @@ export interface NgxsOnInit {
   ngxsOnInit(ctx?: StateContext<any>): void | any;
 }
 
+/**
+ * On change interface
+ */
+export interface NgxsOnChanges {
+  ngxsOnChanges(change?: NgxsSimpleChange): void;
```
markwhitfeld: The '?' in this method declaration would mean that this method could be invoked without a value for the `change` parameter. This is not the case because it is always provided by the framework.
Please remove the `?`.

splincode: I do not need, in this case, an argument

```
class State implements NgxsOnChanges {
  public ngxsOnChanges(): void {
     this.myService.trigger();
  }
}
```

markwhitfeld: Yes, but that is not what the question mark means. You can always leave out parameters in your implementation and TS will be happy. Try it out.

packages/store/src/symbols.ts
splincode: done

packages/store/src/symbols.ts
splincode: done

markwhitfeld:
@arturovt Would you like to review?

Issue #1375, opened 2019-10-09
🐞[BUG]: State operators typing issue


bvolkj:
### Affected Package

 The issue is caused by package @ngxs/3.5.1

### Is this a regression?

 No (only tested 3.5.0)

### Description

 
When using the typescript strict compiler option there are Typescript compilation Errors when using certain state operators.

## 🔬 Minimal Reproduction
As it isn't possible to edit tsconfig files in stackblitz, I created a repo to reproduce the issue
https://github.com/bvolkj/ngxs-typing/tree/master/test

As additional information, this is what I did:

- set Typescript compilerOption strict to `true`
tsconfig.json:
 ```
 {
  "compileOnSave": false,
  "compilerOptions": {
    "strict": true,
    "baseUrl": "./",
...
``` 
- create a state with an action that uses any of these state operators: removeItem, append, or insertItem. I used the following example code: https://www.ngxs.io/advanced/operators#advanced-example

## 🔥 Exception or Error
When starting the application with `ng serve` the following errors occur:
<pre><code>

src/app/modules/profile/store/zoo.state.ts:36:17 - error TS2322: Type '(existing: string[]) => string[]' is not assignable to type 'string[] | StateOperator<string[]> | undefined'.
  Type '(existing: string[]) => string[]' is not assignable to type 'StateOperator<string[]>'.
    Types of parameters 'existing' and 'existing' are incompatible.
      The type 'readonly string[]' is 'readonly' and cannot be assigned to the mutable type 'string[]'.

36         zebras: append([payload])
                   ~~~~~~~~~~~~~~~~~

  src/app/modules/profile/store/zoo.state.ts:36:17
    36         zebras: append([payload])
                       ~~~~~~~~~~~~~~~~~
    Did you mean to call this expression?
src/app/modules/profile/store/zoo.state.ts:45:17 - error TS2322: Type '(existing: string[]) => string[]' is not assignable to type 'string[] | StateOperator<string[]> | undefined'.
  Type '(existing: string[]) => string[]' is not assignable to type 'StateOperator<string[]>'.
    Types of parameters 'existing' and 'existing' are incompatible.
      The type 'readonly string[]' is 'readonly' and cannot be assigned to the mutable type 'string[]'.

45         pandas: removeItem<string>(name => name === payload)
                   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

  src/app/modules/profile/store/zoo.state.ts:45:17
[...]

PR #1394, opened 2019-10-17
fix(store): improved typings for state operators

splincode:
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[x] Other... Please describe:
```

Issue Number: https://github.com/ngxs/store/issues/1375

image

image

## What is the new behavior?

This test needs to work correctly later, this test confirms the problem now.

```
[ ] Yes
[x] No
```

### DIFF
--- a/packages/store/types/tests/state-operator.lint.ts
+++ b/packages/store/types/tests/state-operator.lint.ts
@@ -1,0 +1,76 @@
+import { State, Action, StateContext } from '@ngxs/store';
+import { patch, append, removeItem, insertItem, updateItem } from '@ngxs/store/operators';
+
+export interface AnimalsStateModel {
+  zebras: string[];
+  pandas: string[];
+}
+
+export class AddZebra {
+  static readonly type = '[Animals] Add zebra';
+
+  constructor(public payload: string) {}
+}
+
+export class RemovePanda {
+  static readonly type = '[Animals] Remove panda';
+
+  constructor(public payload: string) {}
+}
+
+export class Test {
+  static readonly type = '[Animals] test';
+
+  constructor(public payload: string) {}
+}
+
+export class ChangePandaName {
+  static readonly type = '[Animals] Change panda name';
+
+  constructor(public payload: { name: string; newName: string }) {}
+}
+
+@State<AnimalsStateModel>({
+  name: 'animals',
+  defaults: {
+    zebras: ['Jimmy', 'Jake', 'Alan'],
+    pandas: ['Michael', 'John']
+  }
+})
+export class AnimalsState {
+  @Action(AddZebra)
+  addZebra(ctx: StateContext<AnimalsStateModel>, { payload }: AddZebra) {
+    ctx.setState(
+      patch({
+        zebras: append([payload]) // $ExpectError
+      })
+    );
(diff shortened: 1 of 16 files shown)


packages/store/types/tests/state-operator.lint.ts
```diff
@@ -0,0 +1,76 @@
...
+        pandas: insertItem<string>(payload) // $ExpectError
+      })
+    );
+  }
+
+  @Action(ChangePandaName)
+  changePandaName(ctx: StateContext<AnimalsStateModel>, { payload }: ChangePandaName) {
+    ctx.setState(
+      patch({
+        pandas: updateItem<string>(name => name === payload.name, payload.newName)
```
markwhitfeld: So, is this operator fine?

packages/store/types/tsconfig.json
```diff
@@ -20,9 +20,9 @@
     "baseUrl": ".",
     "paths": {
       "@angular/*": ["../../../node_modules/@angular/*"],
-      "@ngxs/store/operators": ["../../../packages/store/operators/src"],
-      "@ngxs/store/internals": ["../../../packages/store/internals/src"],
-      "@ngxs/*": ["../../../packages/*"]
+      "@ngxs/store/operators": ["../../../@ngxs/store/operators/index.d.ts"],
+      "@ngxs/store/internals": ["../../../@ngxs/store/internals/index.d.ts"],
```
markwhitfeld: Is this how we simulate the `strict` compile options?
This does introduce the dependency that these tests can only be run after the build is done. Are the scripts adjusted accordingly?

packages/store/types/tests/state-operator.lint.ts
splincode: No, this test shows that we have a compilation error.

packages/store/types/tsconfig.json
splincode: Unfortunately, otherwise we cannot verify the reality of the work

packages/store/types/tests/state-operator.lint.ts
markwhitfeld: But there is no `// $ExpectError` next to this one, but there is next to the others....
Does that mean that this one is ok?
I do think that we should be documenting the ones that are ok as well as the broken ones.

packages/store/types/tsconfig.json
markwhitfeld: This does introduce the dependency that these tests can only be run after the build is done. Are the scripts adjusted accordingly?

splincode: These tests should be run after assembly! Since they guarantee that the real API is built under TS

packages/store/types/tests/state-operator.lint.ts
splincode: Oh, sorry, this good work :)

splincode:
@markwhitfeld @arturovt this pull request fixed error
https://github.com/ngxs/store/issues/1375

I also checked, the error is easily reproduced in version 3.5.1 (critical bug with operators)

Issue #1323, opened 2019-09-13
🚀[FEATURE]: Expose StateFactory and StateContextFactory for ngxs-data plugin


splincode:
### Relevant Package
 This feature request is for @ngxs-labs/data

### Description

Problem with UMD modules (Éµn as StateFactory)

![image](https://user-images.githubusercontent.com/12021443/64888521-e7cbff80-d673-11e9-9481-894485f0e1b6.png)

![image](https://user-images.githubusercontent.com/12021443/64888461-c23ef600-d673-11e9-8fc3-63f3d7ead6cf.png)

### Describe the solution you'd like

Expose StateFactory, StateContextFactory for plugins API.

### Describe alternatives you've considered
 Have you considered any alternative solutions or workarounds?

markwhitfeld:
Hi @splincode I really do not want to expose the internals of NGXS for plugins. This can compromise the fundamentals of the framework and cause instability in the plugins if we change these internals.

I will respond to your PR #1325 with a potential option here (Exposing an InjectionToken under the internals subpackage).

PR #1410, opened 2019-10-24
fix(store): ensure state instance in metadata

splincode:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

https://github.com/ngxs-labs/data/blob/master/src/lib/decorators/action/action.ts#L58

image

Now, to get the state instance, I needed to use the stateFactory reference

Issue Number: https://github.com/ngxs/store/issues/1323

In fact, it turns out that you could provide a link to the state instance

## What is the new behavior?

```ts
// INSTEAD
import { ... } from '@ngxs/...';
const mapped: MappedStore = NgxsDataAccessor.ensureMappedState(repository.stateMeta);
const stateInstance: Type<unknown> = mapped.instance;

// USE
import { getStoreMetadata } from '@ngxs/store';
const stateInstance: Type<unknown> = getStoreMetadata(StateClassTarget).instance;
```

```
[ ] Yes
[x] No
```
I really need this in the release, as it provides simplified access to states and their methods

### DIFF
--- a/packages/store/src/internal/state-factory.ts
+++ b/packages/store/src/internal/state-factory.ts
@@ -104,10 +104,13 @@ export class StateFactory {
     for (const name of sortedStates) {
       const stateClass: StateClassInternal = nameGraph[name];
       const depth: string = depths[name];
+
+      // external state metadata (ExampleState['NGXS_META'])
       const meta: MetaDataModel = stateClass[META_KEY]!;
 
       this.addRuntimeInfoToMeta(meta, depth);
 
+      // internal state metadata (only use in state factory)
       const stateMap: MappedStore = {
         name,
         depth,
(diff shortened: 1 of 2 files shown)


packages/store/src/internal/state-factory.ts
```diff
@@ -116,6 +116,8 @@ export class StateFactory {
         defaults: StateFactory.cloneDefaults(meta.defaults)
       };
 
+      meta.instance = stateMap.instance;
```
markwhitfeld: Oh no! This `instance` property was supposed to be removed from the `MetaDataModel` a very long time ago.
It was added by a mistake by Danny in this commit: https://github.com/ngxs/store/commit/35f3e149fe3b266ef1aaea61f7f5f67c7af2cb2a
It is not used and should be removed. I'll add a PR to remove it.
It should not be confused with the `instance` property on the `MappedStore` objects which are maintained within the `StateFactory`.

Having this on the type violates encapsulation so badly. The instance must remain within the objects managed by the Angular Injector.

Issue #1329, opened 2019-09-16
🚀[FEATURE]: Custom deserializer


davidmorissette:
### Relevant Package

 This feature request is for @ngxs/storage-plugin

### Description
 The plugin currently uses a global deserializer which calls "JSON.parse" and returns a generic JavaScript object. When the state model is a class (instead of being an interface or a type), the current implementation does not provide a way to convert the generic object into an concrete instance.

### Describe the solution you'd like
 A new options value "serialization" should provide the ability to define a collection of key/deserializer so we can inject our own implementation:

```
NgxsStoragePluginModule.forRoot({
      serialization: [
        {
          key: 'zoo',
          onAfterDeserializer: (state) => {
            // return concrete instance of the model
          }
        }
      ]
    })
```

arturovt:
@davidmorissette would you want to do a PR? We would be happy :)

splincode:
@markwhitfeld @arturovt what do you think about? 

```ts
import { NgxsSerializable } from '@ngxs/storage-plugin';

class AppState implements NgxsSerializable<AppStateModel> {
  public serialize(state): string {
    return JSON.stringify(state); // or custom stringify
  }

   public deserialize(value: string): AppStateModel {
    return JSON.parse(value); // or custom serial
  }
}
```

```ts
interface NgxsSerializable<T> {
 serialize?(state: T): string;
 deserialize?(value: string): T;
}
```

davidmorissette:
@arturovt https://github.com/ngxs/store/pull/1447

PR #1447, opened 2019-11-12
feat(storage-plugin): add serialization strategy

davidmorissette:
## PR Checklist

- [X ] The commit message follows our guidelines: https://github.com/ngxs/store/blob/master/CONTRIBUTING.md#commit
- [ X] Tests for the changes have been added (for bug fixes / features)
- [ X] Docs have been added / updated (for bug fixes / features)

## PR Type

```
[ ] Bugfix
[X] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

There is no per-state serialization.

Issue Number: N/A

## What is the new behavior?

Per-state serialization offers full control on how states are serialize/deserialize.

```
[ ] Yes
[X] No
```

### DIFF
--- a/packages/storage-plugin/src/storage.plugin.ts
+++ b/packages/storage-plugin/src/storage.plugin.ts
@@ -68,8 +68,8 @@ export class NgxsStoragePlugin implements NgxsPlugin {
           }
 
           this.executeSerializationStrategies(key, strategy => {
-            if (strategy.deserialize) {
-              val = strategy.deserialize(val);
+            if (strategy.onAfterDeserialize) {
+              val = strategy.onAfterDeserialize(val);
             }
           });
 
@@ -93,8 +93,8 @@ export class NgxsStoragePlugin implements NgxsPlugin {
             }
 
             this.executeSerializationStrategies(key, strategy => {
-              if (strategy.serialize) {
-                val = strategy.serialize(val);
+              if (strategy.onBeforeSerialize) {
+                val = strategy.onBeforeSerialize(val);
               }
             });
 
--- a/packages/storage-plugin/src/storage.plugin.ts
+++ b/packages/storage-plugin/src/storage.plugin.ts
@@ -38,7 +38,8 @@ export class NgxsStoragePlugin implements NgxsPlugin {
     const keys = this._options.key as string[];
     const matches = actionMatcher(event);
     const isInitAction = matches(InitState) || matches(UpdateState);
-    let hasMigrationOrSerialization = false;
+    let hasMigration = false;
+    let hasSerialization = false;
 
     if (isInitAction) {
       for (const key of keys) {
@@ -62,14 +63,14 @@ export class NgxsStoragePlugin implements NgxsPlugin {
               const keyMatch = (!strategy.key && isMaster) || strategy.key === key;
               if (versionMatch && keyMatch) {
                 val = strategy.migrate(val);
-                hasMigrationOrSerialization = true;
+                hasMigration = true;
               }
             });
           }
 
           this.executeSerializationStrategies(key, strategy => {
             if (strategy.onBeforeSerialize) {
-              hasMigrationOrSerialization = true;
+              hasSerialization = true;
             }
 
             if (strategy.onAfterDeserialize) {
@@ -88,7 +89,11 @@ export class NgxsStoragePlugin implements NgxsPlugin {
 
     return next(state, event).pipe(
       tap(nextState => {
-        if (!isInitAction || (isInitAction && hasMigrationOrSerialization)) {
+        if (
+          !isInitAction ||
+          (isInitAction && hasMigration) ||
+          (isInitAction && hasSerialization)
+        ) {
           for (const key of keys) {
             let val = nextState;
 
--- a/packages/storage-plugin/src/storage.plugin.ts
+++ b/packages/storage-plugin/src/storage.plugin.ts
@@ -14,6 +14,7 @@ import { tap } from 'rxjs/operators';
 import {
   StorageEngine,
   NgxsStoragePluginOptions,
+  NgxsStoragePluginSerializationOptions,
   STORAGE_ENGINE,
   NGXS_STORAGE_PLUGIN_OPTIONS
 } from './symbols';
@@ -66,16 +67,11 @@ export class NgxsStoragePlugin implements NgxsPlugin {
             });
           }
 
-          if (this._options.serialization) {
-            this._options.serialization.forEach(strategy => {
-              if (
-                strategy.deserialize &&
-                ((!strategy.key && isMaster) || strategy.key === key)
-              ) {
-                val = strategy.deserialize(val);
-              }
-            });
-          }
+          this.executeSerializationStrategies(key, strategy => {
+            if (strategy.deserialize) {
+              val = strategy.deserialize(val);
+            }
+          });
 
           if (!isMaster) {
             state = setValue(state, key!, val);
@@ -90,23 +86,17 @@ export class NgxsStoragePlugin implements NgxsPlugin {
       tap(nextState => {
         if (!isInitAction || (isInitAction && hasMigration)) {
           for (const key of keys) {
-            const isMaster = key === DEFAULT_STATE_KEY;
             let val = nextState;
 
-            if (!isMaster) {
+            if (key !== DEFAULT_STATE_KEY) {
               val = getValue(nextState, key!);
             }
 
-            if (this._options.serialization) {
-              this._options.serialization.forEach(strategy => {
-                if (
-                  strategy.serialize &&
-                  ((!strategy.key && isMaster) || strategy.key === key)
-                ) {
-                  val = strategy.serialize(val);
-                }
-              });
-            }
+            this.executeSerializationStrategies(key, strategy => {
+              if (strategy.serialize) {
+                val = strategy.serialize(val);
+              }
+            });
 
             try {
               this._engine.setItem(key!, this._options.serialize!(val));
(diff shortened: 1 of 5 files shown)


splincode:
What do you think about
https://github.com/ngxs/store/issues/1329#issuecomment-552579788

davidmorissette:
@splincode I figured that putting the serialization within the options class of the plugin was a better way to follow the strategy pattern already in place (same one as for the migrations). Also, when many state classes share the same model they would have to implement the interface over again. A clean way could be to use the pattern I put in place but to reference static serializer/deserializer methods defined on the respective model classes.

packages/storage-plugin/src/storage.plugin.ts
```diff
@@ -97,4 +110,17 @@ export class NgxsStoragePlugin implements NgxsPlugin {
       })
     );
   }
+
+  private executeSerializationStrategies(
+    key: string,
+    action: (strategy: NgxsStoragePluginSerializationOptions) => void
```
splincode: maybe extract to new type

```ts
type SerializeAction = (strategy: NgxsStoragePluginSerializationOptions) => void;

private executeSerializationStrategies(key: string, action: SerializeAction): void {

}
```

markwhitfeld: @splincode I don't think it is necessary to extract this because it is found in only one place (in this declaration) and it is internal.

packages/storage-plugin/tests/storage.plugin.spec.ts
```diff
@@ -462,4 +473,41 @@ describe('NgxsStoragePlugin', () => {
...
+    TestBed.configureTestingModule({
+      imports: [
+        NgxsModule.forRoot([ConcreteTypeState]),
+        NgxsStoragePluginModule.forRoot({
+          key: 'concreteType',
+          serialization: [
+            {
+              key: 'concreteType',
+              onAfterDeserialize: (obj: any) => {
+                return <CounterStateModel>{
```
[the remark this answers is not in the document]
markwhitfeld: Agreed, a test with both directions is needed too.

packages/storage-plugin/src/symbols.ts
```diff
@@ -7,6 +7,23 @@ export const enum StorageOption {
   SessionStorage
 }
 
+export interface NgxsStoragePluginSerializationOptions {
```
markwhitfeld: @davidmorissette I'm just wondering if it would be possible to add any additional typings here?
Not sure how this would work when expressed in the array, but let's hope that it 'Just Works™️ '.
We could default it to `any`... maybe that would make the array happy.
Something like this (comments excluded for berevity):
```ts
export interface NgxsStoragePluginSerializationOptions<TStateModel = any, TStoredModel = any> {
  key?: string;
  onBeforeSerialize?: (obj: TStateModel) => TStoredModel;
  onAfterDeserialize?: (obj: TStoredModel) => TStateModel;
}
```

splincode: @davidmorissette ping

davidmorissette: @markwhitfeld As far as I know, it is not possible to implement an array of different generics with TypeScript. Please correct me if I am wrong. Plus, I don't see any pattern of reusability with these methods and I am not so sure generics are a good fit there. Migrations don't make usage of them.

davidmorissette:
@markwhitfeld In my opinion, one advantage of using TypeScript is class-based programming. The library itself supports putting instances in the store out of the box. The ability to deserialize objects back to their original types is important since it aligns the plugin with the library itself. Thank you!

splincode:
@davidmorissette

> In my opinion, one advantage of using TypeScript is class-based programming. The library itself supports putting instances in the store out of the box. 

You are confusing the data type of compile time and runtime.

```ts
class A {
  public a: number = 5;
}

class B extends A {
  public b: number = 10;
}

const b: B = new B(); 
console.log(b); // { a: 5, b: 10 }
console.log(b instanceof B) // true
```

but

```ts
const b: B = new B(); 
const b2: B = { ...b, b: 20 };
console.log(b); // { a: 5, b: 20 }
console.log(b instanceof B) // false
```

Any situation that can happen with objects can lead to various side effects. Because operations on objects in Redux are mainly done on pojo.

markwhitfeld:
@davidmorissette Just continuing to discuss before we commit to adding this to the API...
What do you think about the approach of allowing for a state to declare the way that it should be serialized/deserialized? (something like https://github.com/ngxs/store/issues/1329#issuecomment-552579788)

I guess the question would be if this logic would be the responsibility of the storage options or the state class. Also, would this logic be used anywhere else? ... like the dev tools plugin...?

davidmorissette:
@markwhitfeld See my existing reply regarding your question: https://github.com/ngxs/store/pull/1447#issuecomment-553250084
Moving the pre-serialization/post-deserialization to the state would also take away the possibility of using no key (@@STATE). But my main argument is that migrations have been defined around the storage options (There is no ngxsMigration interface implemented on the state) and for consistency reasons it is better to follow the same approach. What do you think?

markwhitfeld:
@arturovt I would love to hear your thoughts on this one.

markwhitfeld:
Sorry to do this so late in the discussion but I have just realized that this could probably be done as a plugin which provides a custom serializer and deserializer. This would not require a single ngxs core API change (first prize). I'm thinking something like this:
```ts
// your plugin would have a createSerializationStrategy function like the one below 
// It would return an object with this type of interface:
interface SerializationStrategy {
  serialize?(obj: any): string;
  deserialize?(obj: any): any;
}

// In your app you would then create your custom strategy:

const serializationStrategy = createSerializationStrategy([
  {
    key: 'zoo',
    onBeforeSerialize: state => {
      return { zebra: { name: state.zebra.name.trim() } };
    },
    onAfterDeserialize: state => {
      return <AnimalStateModel>{
        zebra: new Zebra(state.zebra.name)
      };
    }
  }
]);

// Then in your module, you would use it...

@NgModule({
  imports: [
    NgxsModule.forRoot([]),
      serialize: serializationStrategy.serialize,
      deserialize: serializationStrategy.deserialize,
    })
  ]
})
export class MyModule {}
```
@davidmorissette @arturovt @splincode What do you think?

davidmorissette:
@markwhitfeld I like the idea of integrating the functionality inside of a serialization plugin. However,  "onBeforeSerialize/onAfterDeserialization" is not where the serialization/deserialization actually happens, but mostly act as an interceptor. Would it be the plugin's responsibility to execute "onBeforeSerialize/onAfterDeserialization" everytime serialize/deserialize is called?

markwhitfeld:
@davidmorissette Yes, that is what I was thinking.

davidmorissette:
@markwhitfeld Since serialize/deserialize are agnostic of the "key" value, no logic can be implemented (these functions only deal with the value, not the key). I also thought about putting the logic inside of a custom engine, but then it would compete with the call of serialize/deserialize by the storage plugin. One way to work around this could be to pass the key as extra parameter on serialize/deserialize in order to avoid a breaking change.

davidmorissette:
@markwhitfeld @arturovt @splincode Do you think it is better to proceed with the initial pull request or is it better to modify the serialize/deserialize functions and pass a 'key' parameter in order to put all the functionality inside of a serialization plugin?

markwhitfeld:
@davidmorissette Sorry for the delay in getting back to you. I have been on holiday and have fallen a bit behind with notifications.
What I was thinking is that the `createSerializationStrategy` would do the following (pseudo code):

- Serialize(obj: any):
  - Run each transformation against the relative property of `obj`:
    - Get the value of the property found at `key` on `obj`
    - run `onBeforeSerialize` on this value
    - Set the value returned back to the `key` property on `obj` (in an immutable update way)
  - Run `JSON.stringify` on the new value.
- Deserialize(json: string)
  - Run `JSON.parse` on the new `json` value to get `obj`.
  - Run each transformation against the relative property of `obj`:
    - Get the value of the property found at `key` on `obj`
    - run `onAfterDeserialize` on this value
    - Set the value returned back to the `key` property on `obj` (in an immutable update way)
  - Return this transformed object from the function

What do you think?

davidmorissette:
@markwhitfeld "Get the value of the property found at key on obj" => How do you map the key with the obj (or the obj with the key)? The serialize/deserialize functions don't pass the key parameter and the obj alone does not know nothing about the key. Can you explain or provide pseudo code for that part?

markwhitfeld:
@davidmorissette the `key` I am referring to here is the one in the transformations configuration:
image
Am I missing something?

markwhitfeld:
Ah! I see what I was missing!
You were talking about the keys that are passed into the storage plugin options to persist specific parts of the state. Yes, the serialise and deserialise methods would then have to take another parameter for `key`. This would enable us to do this as a plugin.

davidmorissette:
@markwhitfeld Exactly! Beside my pull request, passing the key could be a good improvement to the storage plugin in general since it would allow other people to leverage serialize/deserialize in a more flexible way.

I would prefer to see the key to be the first parameter (more standard key/value pair order) but I understand that it would break existing code.

markwhitfeld:
@davidmorissette I think that it actually makes more sense as the second parameter as a `key` property of a `context`  parameter object so that other information could be passed too. I see this context object as being more secondary information (as opposed to the primary information of the data being serialized or deserialized), so this makes it consistent with other serialization/deserialisation methods too.

davidmorissette:
@markwhitfeld I agree with you. I will work on a new PR in the next coming days and I will keep you updated.

davidmorissette:
@markwhitfeld Please have a look to PR #1513

Issue #1223, opened 2019-08-09
🐞[BUG]: Can't resolve all parameters of state 


workfel:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[x] Bug report  
[ ] Performance issue
[ ] Feature request
[ ] Documentation issue or request
[ ] Support request => https://github.com/ngxs/store/blob/master/CONTRIBUTING.md
[ ] Other... Please describe:
</code></pre>

## Current behavior

**!!! Before read this, i'm using Angular 8 with ivy compiler !!!**

When i have a any service on DI in the constructor of State i  have always this error . ` Can't resolve all parameters for NavigatorState: (?).`

This is my app.module
```ts
@NgModule({
  declarations: [
    AppComponent,
  ],
  imports: [
   .....,
    // Ngxs
    NgxsModule.forRoot([
      LayoutState,
      NavigatorState
    ], {
      developmentMode: !environment.production,
    }),
.....
```

The layoutState not have a construction with DI

```ts
@State<LayoutStateModel>({
  name: 'layout',
  defaults: defaultLayoutStateModel
})
export class LayoutState {

}
```

And the State who i am blocked

```ts
@State<NavigatorStateModel>({
  name: 'navigator',
  defaults: {
    items: []
  }
})
export class NavigatorState {

  constructor(private readonly _router: Router) {
  }

  @Action(Navigate)
  add(ctx: StateContext<NavigatorStateModel>, { payload }: Navigate) {
    this._router.navigate([payload]);
  }
}
```

If you want my `angularCompilerOptions`

```json
  "angularCompilerOptions": {
    "fullTemplateTypeCheck": true,
    "strictInjectionParameters": true,
    "enableIvy": true
  }
```

<pre><code>
Libs:
- @angular/core version: 8.2.0
- @angular/compiler-cli version: 8.2.0
- @ngxs/store version: ^3.5.0-dev.master-07d305f

For Tooling issues:
- Node version: 10.16.0
- Platform: Mac

</code></pre>

splincode:
Related: https://github.com/ngxs/store/issues/1165

splincode:
@workfel please add repo for reproduce

splincode:
https://github.com/ngxs/store/pull/1474

tskweres:
How can I fix this? Using latest version of Angular and:

@ngxs/devtools-plugin": "^3.6.2",
@ngxs/store": "^3.6.2",

[...]

Issue #1165, opened 2019-07-20
🐞[BUG]: Ivy build error


C1E1S1T:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ x] Bug report  
[ ] Performance issue
[ ] Feature request
[ ] Documentation issue or request
[ ] Support request => https://github.com/ngxs/store/blob/master/CONTRIBUTING.md
[ ] Other... 
</code></pre>

## Current behavior

Angular cli can not compiled with Ivy (ng serve works):
node --max-old-space-size=8192 ./node_modules/@angular/cli/bin/ng build --prod
```
ERROR in ./src/app/authenticated-users/inventory/inventory.module.ts 49:21-41
"export 'NgxsFeatureModule' (imported as 'i1') was not found in '@ngxs/store'
ERROR in ./src/app/authenticated-users/cashier/cashier.module.ts 49:24-44
"export 'NgxsFeatureModule' (imported as 'i1') was not found in '@ngxs/store'
ERROR in ./src/store/store.module.ts 31:79-96
"export 'NgxsRootModule' (imported as 'i1') was not found in '@ngxs/store'
```

## Expected behavior

Angular cli can compiled with Ivy

## Environment

<pre><code>
Libs:
- @angular/core version: ^8.2.0-next.1
- @ngxs/store version: ^3.3.2

Browser:
- [ ] Chrome (desktop) version XX
- [ ] Chrome (Android) version XX
- [ ] Chrome (iOS) version XX
- [ ] Firefox version XX
- [ ] Safari (desktop) version XX
- [ ] Safari (iOS) version XX
- [ ] IE version XX
- [ ] Edge version XX
- [x] Opera 62.0.3331.72
 
For Tooling issues:
- Node version:  v12.6.0.  
- Platform: Linux Ubuntu 16.4

Others:

Azure
</code></pre>

splincode:
Show stackblitz or repo example

markwhitfeld:
This issue has been reported to the angular team: https://github.com/angular/angular/issues/31791

PS. This is not really possible to reproduce in StackBlitz so @arturovt made a repo here:
https://github.com/arturovt/ivy-ngxs-repro

ErazerBrecht:
We are currently trying Ivy out in our codebase and for now it seems that ngxs 3.4.X works with Ivy.

arturovt:
[...]

PR #1459, opened 2019-11-24
fix(router-plugin): build compatibility for NGXS Router plugin with Ivy

splincode:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: https://github.com/ngxs/store/issues/1223, https://github.com/ngxs/store/issues/1165

## What is the new behavior?

image

image

```
[ ] Yes
[x] No
```

### DIFF
--- a/docs/plugins/router.md
+++ b/docs/plugins/router.md
@@ -31,7 +31,6 @@ Add the `NgxsRouterPluginModule` plugin to your root app module:
 ```ts
 import { NgxsModule } from '@ngxs/store';
 import { NgxsRouterPluginModule } from '@ngxs/router-plugin';
-import { AppRoutingModule } from './app-routing.module';
 
 @NgModule({
   imports: [NgxsModule.forRoot([]), NgxsRouterPluginModule.forRoot()]
--- a/docs/plugins/router.md
+++ b/docs/plugins/router.md
@@ -34,11 +34,7 @@ import { NgxsRouterPluginModule } from '@ngxs/router-plugin';
 import { AppRoutingModule } from './app-routing.module';
 
 @NgModule({
-  imports: [
-    AppRoutingModule, // 1. Init RouterModule.forRoot([ROUTES])
-    NgxsModule.forRoot([AppState]), // 2. Init NGXS Store with states
-    NgxsRouterPluginModule.forRoot() // 3. Init NGXS Router plugin
-  ]
+  imports: [NgxsModule.forRoot([]), NgxsRouterPluginModule.forRoot()]
 })
 export class AppModule {}
 ```
--- a/docs/plugins/router.md
+++ b/docs/plugins/router.md
@@ -31,9 +31,14 @@ Add the `NgxsRouterPluginModule` plugin to your root app module:
 ```ts
 import { NgxsModule } from '@ngxs/store';
 import { NgxsRouterPluginModule } from '@ngxs/router-plugin';
+import { AppRoutingModule } from './app-routing.module';
 
 @NgModule({
-  imports: [NgxsModule.forRoot([]), NgxsRouterPluginModule.forRoot()]
+  imports: [
+    AppRoutingModule, // 1. Init RouterModule.forRoot([ROUTES])
+    NgxsModule.forRoot([AppState]), // 2. Init NGXS Store with states
+    NgxsRouterPluginModule.forRoot() // 3. Init NGXS Router plugin
+  ]
 })
 export class AppModule {}
 ```
--- a/packages/router-plugin/src/router.module.ts
+++ b/packages/router-plugin/src/router.module.ts
@@ -4,15 +4,14 @@ import { NgxsModule } from '@ngxs/store';
 import { RouterState } from './router.state';
 import { DefaultRouterStateSerializer, RouterStateSerializer } from './serializer';
 
-@NgModule()
+@NgModule({
+  imports: [NgxsModule.forFeature([RouterState])]
+})
 export class NgxsRouterPluginModule {
-  static forRoot(): ModuleWithProviders[] {
-    return [
-      NgxsModule.forFeature([RouterState]),
-      {
-        ngModule: NgxsRouterPluginModule,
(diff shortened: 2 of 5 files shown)


docs/plugins/router.md
```diff
@@ -31,9 +31,14 @@ Add the `NgxsRouterPluginModule` plugin to your root app module:
...
 import { NgxsModule } from '@ngxs/store';
 import { NgxsRouterPluginModule } from '@ngxs/router-plugin';
+import { AppRoutingModule } from './app-routing.module';
 
 @NgModule({
-  imports: [NgxsModule.forRoot([]), NgxsRouterPluginModule.forRoot()]
+  imports: [
+    AppRoutingModule, // 1. Init RouterModule.forRoot([ROUTES])
+    NgxsModule.forRoot([AppState]), // 2. Init NGXS Store with states
+    NgxsRouterPluginModule.forRoot() // 3. Init NGXS Router plugin
```
markwhitfeld: With the `1.`, `2.` and `3.` are we saying that this order matters? If so that we should probably be explicit about it.
Otherwise the numbers are not necessary,

packages/router-plugin/src/router.module.ts
```diff
@@ -4,15 +4,14 @@ import { NgxsModule } from '@ngxs/store';
...
-      {
-        ngModule: NgxsRouterPluginModule,
-        providers: [{ provide: RouterStateSerializer, useClass: DefaultRouterStateSerializer }]
-      }
-    ];
+  static forRoot(): ModuleWithProviders {
+    return {
+      ngModule: NgxsRouterPluginModule,
+      providers: [{ provide: RouterStateSerializer, useClass: DefaultRouterStateSerializer }]
+    };
```
markwhitfeld: I'm trying to figure out what fixes the issue?
Was it moving the `NgxsModule.forFeature` to the top?

splincode: Ivy automatically infer type `ModuleWithProviders<T>` by type from `ngModule`

```ts
return {
      ngModule: NgxsRouterPluginModule,
      providers: [{ provide: RouterStateSerializer, useClass: DefaultRouterStateSerializer }]
};
```

Also by Angular Guide forRoot need return `ModuleWithProviders` instead `ModuleWithProviders[]`

docs/plugins/router.md
splincode: Yes order matters

markwhitfeld: Ok, please could you add a line above the code snippet stating that the order matters. ie. what needs to be before what.

Do you know why it is order sensitive?

packages/router-plugin/src/router.module.ts
markwhitfeld: ah, ok. Thanks!

markwhitfeld: I wonder if there are any other plugins we need to fix this for?
... any in the labs? @joaqcid maybe your `actions-executing` because it also has its own state.

packages/router-plugin/src/router.module.ts
```diff
@@ -4,15 +4,14 @@ import { NgxsModule } from '@ngxs/store';
...
 export class NgxsRouterPluginModule {
-  static forRoot(): ModuleWithProviders[] {
-    return [
-      NgxsModule.forFeature([RouterState]),
-      {
-        ngModule: NgxsRouterPluginModule,
-        providers: [{ provide: RouterStateSerializer, useClass: DefaultRouterStateSerializer }]
-      }
-    ];
+  static forRoot(): ModuleWithProviders {
```
arturovt: > It is mandatory to use the generic ModuleWithProviders<T> type to indicate the Angular module type.
```suggestion
  static forRoot(): ModuleWithProviders<put type here> {
```

packages/router-plugin/src/router.module.ts
splincode: @markwhitfeld https://github.com/ngxs-labs/actions-executing/blob/master/src/lib/actions-executing.module.ts

packages/router-plugin/src/router.module.ts
splincode: I can't put type, because Angular 5 build failed

docs/plugins/router.md
splincode: Because then you can get the router status incorrectly in NGXS Router plugin

packages/router-plugin/src/router.module.ts
arturovt: This is required by Ivy, f*** this Angular 5 :joy: :joy:

splincode: CI failed if added

docs/plugins/router.md
markwhitfeld: So, is it important to have the Router plugin after the Routing module for this reason. But it should be after the `NgxsModule.forRoot` for a different reason?

It would be great if we could remove these strange order dependencies somehow though.

packages/router-plugin/src/router.module.ts
markwhitfeld: Not sure if any of the others hold their own state.

packages/router-plugin/src/router.module.ts
markwhitfeld: Hmmm... I'm feeling more and more like we need to drop support for it in this release.

markwhitfeld: Let's drop NG5 support in a separate PR and then add this generic type there.
We should include the reasons for dropping support in that PR.

packages/router-plugin/src/router.module.ts
splincode: > Not sure if any of the others hold their own state.

What do you think?

docs/plugins/router.md
markwhitfeld: @splincode If the order matters, why did you now remove the details about the order that they must be in?

PR #1461, opened 2019-11-25
refactor(store): reduce @ngxs/store/internals bundlesize

splincode:
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[x] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

## What is the new behavior?

Reduced 0.06Kb

### DIFF
--- a/packages/store/src/internal/internals.ts
+++ b/packages/store/src/internal/internals.ts
@@ -381,6 +381,6 @@ export function getStateDiffChanges<T>(
   return new NgxsSimpleChange(previousValue, currentValue, !mappedStore.isInitialised);
 }
 
-export function mergeObjects(a: object, b: object): object {
-  return { ...(a || {}), ...(b || {}) };
+export function merge(a: object, b: object): object {
+  return { ...a, ...b };
 }
--- a/packages/store/src/internal/internals.ts
+++ b/packages/store/src/internal/internals.ts
@@ -380,3 +380,7 @@ export function getStateDiffChanges<T>(
   const currentValue: T = getValue(diff.newAppState, mappedStore.depth);
   return new NgxsSimpleChange(previousValue, currentValue, !mappedStore.isInitialised);
 }
+
+export function mergeObjects(a: object, b: object): object {
+  return { ...(a || {}), ...(b || {}) };
+}
--- a/bundlesize.config.json
+++ b/bundlesize.config.json
@@ -2,23 +2,23 @@
   "files": [
     {
       "path": "./@ngxs/store/fesm2015/ngxs-store-internals.js",
-      "maxSize": "3 kB"
+      "maxSize": "1.72KB"
     },
     {
       "path": "./@ngxs/store/fesm2015/ngxs-store-operators.js",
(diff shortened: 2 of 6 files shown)


packages/store/src/internal/internals.ts
```diff
@@ -380,3 +380,7 @@ export function getStateDiffChanges<T>(
   const currentValue: T = getValue(diff.newAppState, mappedStore.depth);
   return new NgxsSimpleChange(previousValue, currentValue, !mappedStore.isInitialised);
 }
+
+export function mergeObjects(a: object, b: object): object {
+  return { ...(a || {}), ...(b || {}) };
```
arturovt:
```suggestion
  return { ...a, ...b };
```

packages/store/src/internal/internals.ts
```diff
@@ -380,3 +380,7 @@ export function getStateDiffChanges<T>(
   const currentValue: T = getValue(diff.newAppState, mappedStore.depth);
   return new NgxsSimpleChange(previousValue, currentValue, !mappedStore.isInitialised);
 }
+
+export function mergeObjects(a: object, b: object): object {
```
arturovt:
```suggestion
export function merge(a: object, b: object): object {
```

arturovt:
This might be a refactor PR and not a fix.

markwhitfeld:
Did we reduce the bundle by 6kB or by 0.06kB with this change?
I'm wondering if it is worth it?

splincode:
@markwhitfeld 0.06kB
but improved readability and treeshaking :)

markwhitfeld:
There doesn't really seem to be any real value to this refactor.
Was the merge function used by anything other than the store lib?

I'm actually wondering what the benefit of having the merge function was in the first place...
The only benefit would be the fact that all the extra code that ES2017 and earlier transpilation adds  (ie. `Object.assign(Object.assign({}, a), b)`) would now only be found in one place. Once you start looking at this it is a slippery slope because the normal spread syntax can be optimized by the V8 engine (I think) so then if you optimize for old you end up sacrificing the new.

splincode:
@artuvotr

arturovt:
Hey, Mr. Mark and Mr. Max :)

> the normal spread syntax can be optimized by the V8 engine

No. The TC39 spec says that spread operator (specifically for objects) is syntactic sugar over "copy data properties" from one object to the second one and evaluate assignment expression.

Spread for objects is a part of ES2018 spec which is slightly supported by modern browsers because it also has other features.

As we can't target ES2018 then this still will be transformed into `Object.assign` by the TS compiler.

> I'm actually wondering what the benefit of having the merge function was in the first place...

Yeah. I think we shouldn't have the `merge` function. I'm sure we can do this directly there:
```js
{ ...this._config.defaultsState, ...initialStateValue }
```

arturovt:
I actually liked this PR as I approved it :slightly_smiling_face: Just thought that we could get rid of `merge` and move destruction there :)

PR #1462, opened 2019-11-25
refactor(store): replace `for-of` with oldie for loops

arturovt:
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: N/A

```
[ ] Yes
[x] No
```

## Other information

`for-of` is actually used to iterate over ANY iterable (map, set, etc...). TS compiler emits `__value` helper when targetting ES5, which supports ES2015 iteration protocol.

See how much code is emitted:
---
image
---
image

There are a lot of `try-catch-finally` calls. We wouldn't want them when we just wanna loop over a single array. It seems like the bundlesize grew (from 18.67 to 18.74), but no. This is because the bundlesize package checks es2015 files and not es5.

I just checked the size of `@ngxs/store/bundles/ngxs-store.umd.min.js` (**it's minified**). It was 32KB before this refactoring and became 30KB after.

ES5 is still important for old browsers. *And if the user uses old browser then the bundlesize really matters and those unnecessary emitted `try-catch-finally` blocks can slow the code execution on older browsers*.

### DIFF
--- a/bundlesize.config.json
+++ b/bundlesize.config.json
@@ -2,102 +2,122 @@
   "files": [
     {
       "path": "./@ngxs/store/fesm2015/ngxs-store-internals.js",
-      "maxSize": "6.22kB",
+      "package": "@ngxs/store/internals",
+      "maxSize": "6.220kB",
       "compression": "none"
     },
     {
       "path": "./@ngxs/store/fesm5/ngxs-store-internals.js",
-      "maxSize": "7.18kB",
+      "package": "@ngxs/store/internals",
+      "maxSize": "7.180kB",
       "compression": "none"
     },
     {
       "path": "./@ngxs/store/fesm2015/ngxs-store-operators.js",
-      "maxSize": "9.68kB",
+      "package": "@ngxs/store/operators",
+      "maxSize": "9.680kB",
       "compression": "none"
     },
     {
       "path": "./@ngxs/store/fesm5/ngxs-store-operators.js",
-      "maxSize": "9.82kB",
+      "package": "@ngxs/store/operators",
+      "maxSize": "9.820kB",
       "compression": "none"
     },
     {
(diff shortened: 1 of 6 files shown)


splincode:
@markwhitfeld In total (taking into account the difference of the two packages) savings in `0.5kB`. I think this is also a good result.

arturovt:
Don't worry guys, I will reopen it with more changes a little bit later!

arturovt:
Max, please, do not push to other members PRs w/o negotiation. I had to cherry pick your last commit to retrieve those changes locally, also I ain't able to squash :(

splincode:
@arturovt can be merge?

arturovt:
> @arturovt can be merge?

Don't know, do we need Mark's approve?

markwhitfeld:
I don't think that we should make the change to check against the uncompressed sizes in this PR. It makes the effect on build sizes more difficult to understand within this PR.

Also, are we wanting to smake our modern bundle size larger (and the code less modern) to make it smaller for old browsers?

arturovt:
> I don't think that we should make the change to check against the uncompressed sizes in this PR. It makes the effect on build sizes more difficult to understand within this PR.
> 
> Also, are we wanting to smake our modern bundle size larger (and the code less modern) to make it smaller for old browsers?

Basically, this change will not affect modern browsers (swapping from `for-of` to a single `for` loop). But this will affect older browsers.

markwhitfeld:
Honestly, I feel like this is a micro-optimisation that has little value.
Close it if you agree.

arturovt:
> Honestly, I feel like this is a micro-optimisation that has little value.
> Close it if you agree.

Agreed.

Issue #1467, opened 2019-11-28
🚀[FEATURE]: Support passing type to store.reset


ahnpnl:
### Relevant Package

 This feature request is for @ngxs/store

### Description
`reset` method in `Store` doesn't support passing type. Reset now is expecting always `any` type.

### Describe the problem you are trying to solve
I usually use `store.reset` in my unit test like

```
store.reset({
     animal: <default_state_value>
})
```

However, this is not type strict.

### Describe the solution you'd like
Support passing type to `reset`. For example:
```
store.reset<Animal>({
      animal: <default_state_value> // This <default_state_value> should follow type Animal
})
```

### Describe alternatives you've considered
I don't have any workarounds.

splincode:
reset uses the `setState` which will return `void` in v4.

by @arturovt 

> Will tell you frankly I've never seen that someone had to use the `reset` method result. I don't say that other people don't do it, just meant that people in my team have never done it.
This case is very specific for your project, I guess that could be kept out of the core framework. You always can do something like:

```ts
const reset = <T = any>(state: T) => store.reset(state);
```

PR #1490, opened 2019-12-05
feat(store): support passing type to `store.reset` method

splincode:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: https://github.com/ngxs/store/issues/1467

## What is the new behavior?

Now it is safer to get the type from the method

```
[ ] Yes
[x] No
```

### DIFF
--- a/packages/store/src/store.ts
+++ b/packages/store/src/store.ts
@@ -99,7 +99,7 @@ export class Store {
    * Reset the state to a specific point in time. This method is useful
    * for plugin's who need to modify the state directly or unit testing.
    */
-  reset(state: any) {
+  reset<T = any>(state: T): T {
     return this._internalStateOperations.getRootStateOperations().setState(state);
   }
 
--- a/packages/store/types/tests/dispatch.lint.ts
+++ b/packages/store/types/tests/dispatch.lint.ts
@@ -92,7 +92,7 @@ describe('[TEST]: Action Types', () => {
   it('should be correct type base API', () => {
     assertType(() => store.snapshot()); // $ExpectType any
     assertType(() => store.subscribe()); // $ExpectType Subscription
-    assertType(() => store.reset({})); // $ExpectType any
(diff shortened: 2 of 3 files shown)


splincode:
```
This method was deliberately designed this way. It is not recommended to have too much knowledge of the structure of your root state therefore we used any.
```

Why? For example, in my tests it is very convenient to return a type that I myself determined. 

Now in tests I have to do so

```ts
const newState = store.reset(newAppState) as AppState;
// it is necessary for IDEA autocomplete and infer expect type
expect(newState.deep.value).toEqual('val');
```

packages/store/src/store.ts
```diff
@@ -99,7 +99,7 @@ export class Store {
    * Reset the state to a specific point in time. This method is useful
    * for plugin's who need to modify the state directly or unit testing.
    */
-  reset(state: any) {
+  reset<T = any>(state: T): T {
```
arturovt: `reset` uses the `setState` which will return `void` in v4.

arturovt:
> ```
> This method was deliberately designed this way. It is not recommended to have too much knowledge of the structure of your root state therefore we used any.
> ```
> 
> Why? For example, in my tests it is very convenient to return a type that I myself determined.
> 
> Now in tests I have to do so
> 
> ```ts
> const newState = store.reset(newAppState) as AppState;
> // it is necessary for IDEA autocomplete and infer expect type
> expect(newState.deep.value).toEqual('val');
> ```

Will tell you frankly I've never seen that someone using the `reset` method result. I don't say that other people don't do it, just meant that people in my team have never done it.

This case is very specific for your project, I guess that could be kept out of the core framework. You always can do something like:
```ts
const reset = <T = any>(state: T) => store.reset(state);
```

Issue #1504, opened 2019-12-20
🐞[BUG]: NGXS 3.5+ causing pollution across unrelated tests


johnmcase:
### Affected Package

The issue is caused by package @ngxs/store

### Is this a regression?

Yes, the previous version in which this bug was not present was: 3.4.x

### Description

There is some kind of pollution in the store happening across unit tests.  

In the reproduction provided I have 2 states, `ConfigState` and `ContactState`.  ContactState has a dependency on ConfigState in its `ngxsOnInit()` method

one for `BasicComponent` and one for `SomeService`.  If the Component test runs first then the Service test fails.  If the Service test runs first then both pass.  You can change the order of the tests in the `src/main-testing.ts` file

Also, If the Component test imports the `ConfigState` then both tests pass regardless of order.  That spec file is `src/app/components/basic/basic.component.spec.ts`

When the SomeService test is failing, the dispatch action in the beforeEach() appears to be completely ignored.  When the test is passing that dispatch is being handled. 

I think the most correct behavior should be that the BasicComponent test should always fail when it doesn't import ConfigState due to the dependency ContactState has on ConfigState.  The SomeService test should never fail.

This has caused a huge headache on my project after upgrading to Angular 8 (and thus to NGXS 3.5).  We started seeing all kinds of non-deterministic test failures that were happening because jasmine runs the tests in random order.  I've had to work around this by using a sledge hammer and forcing every test that imports any state to import ALL of the states which not ideal at all.

This is probably related to #1180 but this is a very different expression of the problem.

## 🔬 Minimal Reproduction

https://stackblitz.com/edit/ngxs-non-deterministic-tests-ihsuyd

## 🔥 Exception or Error
None

## Environment

<pre><code>
Libs:
- @angular/core version: 8.2.14
- @ngxs/store version: >=3.5.0  still an issue on 3.6.0

Browser:
- [x ] Chrome (desktop) version XX
- [ ] Chrome (Android) version XX
- [ ] Chrome (iOS) version XX
- [ ] Firefox version XX
- [ ] Safari (desktop) version XX
- [ ] Safari (iOS) version XX
- [ ] IE version XX
- [ ] Edge version XX
 
For Tooling issues:
- Node version: 10.16.3
- Platform:  Windows

Others:

</code></pre>

josiahdahl:
[...]

Issue #1180, opened 2019-07-24
🐞[BUG]: No reset of NgxsModule.forRoot after test run


amjmhs:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[x] Bug report  
[ ] Performance issue
[ ] Feature request
[ ] Documentation issue or request
[ ] Support request => https://github.com/ngxs/store/blob/master/CONTRIBUTING.md
[ ] Other... Please describe:
</code></pre>

## Current behavior
I want to test a component that has a `@Select()` on a state.
```typescript
@Component({
  selector: 'my-user',
  templateUrl: './user.component.html',
  styleUrls: ['./user.component.scss'],
})
export class UserComponent {
  /** select current user */
  @Select(UserState)
  user$: Observable<UserDto>;
```

At my test, I prefill the store with: 
```typescript
 beforeEach(async(() => {
    TestBed.configureTestingModule({
      declarations: [UserComponent],
      imports: [
       ...
      ],
    }).compileComponents();

    const store = TestBed.get(Store);
    store.reset({
        user: {
            id: 3,
            name: 'MyUser',
        },
    });
``` 
Obviously, this will fail, because I forget the TestBed-Import of the `NgxsModule` with the according state:
```
      imports: [
        NgxsModule.forRoot([UserState], {
          developmentMode: true,
        }),
      ],
```

But, if there is any other test that runs before the `UserComponent`-test and imports this `UserState` then my `UserComponent`-test will succeed.

At the moment, we have over 500 tests running in random order and from time to time there are failures because someone forget to provide the neccessary state at `NgxsModule.forRoot()`. But with an other random order, this test will succeed again.

## Expected behavior
`NgxsModule.forRoot()` should be isolated for each TestBed.

## Environment

<pre><code>
Libs:
- @angular/core version: 7.2.15
- @ngxs/store version: 3.3.4

Browser:
- [x] Chrome (desktop) version 75.0.3770 (Windows 10.0.0) 
- [ ] Chrome (Android) version XX
- [ ] Chrome (iOS) version XX
- [ ] Firefox version XX
- [ ] Safari (desktop) version XX
- [ ] Safari (iOS) version XX
- [ ] IE version XX
- [ ] Edge version XX
 
For Tooling issues:
- Node version: v12.6.0  
- Platform: Windows 10

Others:

</code></pre>

splincode:
Please add repo for reproduce

[...]

PR #1508, opened 2019-12-26
fix: selectors should be deterministic based on store being used

markwhitfeld:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Currently, when the ngxs module is loaded for some states it updates the metadata of those states with the path for those states within the state tree. This is a problem because the metadata should be immutable to prevent one test from affecting the next. Although we cannot remove the behavior with the `path` because it is being used by some plugins, but we can decouple the selectors from this property so that they become deterministic based on the store that invoked them.

Issue Number: #1504, #1180 

## What is the new behavior?

Selectors now use the state paths as determined from the store that invoked them. This results in the selectors becoming deterministic and avoids the testing issues as described in the above issues.

```
[ ] Yes
[x] No
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 $ npm install @ngxs/store@dev
 ```
 
+- Fix: Selectors should be deterministic based on store being used [#1508](https://github.com/ngxs/store/pull/1508)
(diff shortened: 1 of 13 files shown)


markwhitfeld:
I would like to wait for @arturovt 's review before merging.

PR #1509, opened 2019-12-31
fix: add support for using state tokens in sub states

markwhitfeld:
## PR Checklist

## PR Type

```
[X] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: N/A

Currently, there exists an issue with the State Token construct that, if it is used with a sub-state, the selector cannot locate the state within the state tree. This is due to the fact that, when it was implemented, there didn't exist any deterministic construct to determine the location of a state by its' name. PR #1508 added this construct that we are now able to take advantage of.

## What is the new behavior?

The selector function for the State Token uses the runtime information passed from the store to determine the location of the State that the token represents.

There is an opportunity to optimize this further, but that optimization will be implemented in a subsequent PR and is not needed for this fix.

```
[ ] Yes
[X] No
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ $ npm install @ngxs/store@dev
 ```
 
 - Fix: Selectors should be deterministic based on store being used [#1508](https://github.com/ngxs/store/pull/1508)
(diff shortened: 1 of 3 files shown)


markwhitfeld:
Great, I would like to get @arturovt's review before merging this.

Issue #1392, opened 2019-10-15
🐞[BUG]: Incorrect Order of Interacting Selectors failure not obvious


r-LaForge:
### Affected Package
The issue is caused by package @ngxs/store.

### Is this a regression?
~I don't think so. I think this has always been the behaviour.~
EDIT:
No, I think ordering has always influenced the runtime behaviour. However, I think the actual behaviour is unknown.
It's described in the [docs](https://www.ngxs.io/concepts/select#the-order-of-interacting-selectors) as a topic that may be helpful for those users who keep their selectors in separate classes.

### Description
When declaring `Selectors` on your `State` class our-of-order you get strange behaviour rather than an error / anything helpful.

For example, the below reproduction shows how the first argument is set to the entire state when selectors are out of order. Having a clearer error message could help a developer pinpoint the problem sooner.

I'm not sure how easy this would be to change, or if this would be better classified as a `feature request`. If changes are needed, please let me know and I can update this.

## 🔬 Minimal Reproduction
Minimal reproduction can be found here:
https://stackblitz.com/edit/selector-ordering-issue

Simply open Chrome dev tools and the debugger will stop in the Selector, which will have invalid state as the first argument.

## Environment

<pre><code>
Libs:
@angular/common8.2.6
@angular/compiler8.2.6
@angular/core8.2.6
@angular/forms8.2.6
@angular/platform-browser8.2.6
@angular/platform-browser-dynamic8.2.6
@angular/router8.2.6
@ngxs/store3.5.1
core-js2.6.9
rxjs6.5.3
zone.js0.9.1

arturovt:
The order of selectors section is described in our [docs](https://www.ngxs.io/concepts/select#the-order-of-interacting-selectors).

r-LaForge:
Thanks for the link, I'm aware of that.
It doesn't mean it won't confuse, and doesn't mean it couldn't be handled with an error rather than seemingly undefined behaviour.

arturovt:
I think we can move that discussion to the v4 list of breaking changes issue. Because currently that would really be a breaking change if start throwing some errors.

markwhitfeld:
I wonder if we could potentially fix this order dependency or at least detect it and log something to the console in development mode?

markwhitfeld:
@r-LaForge I have merged a PR that should fix this issue. Please could you test with the `@ngxs/store@dev` package and confirm if it is fixed for you or not.

r-LaForge:
[...]

Issue #1340, opened 2019-09-20
📚[DOCS]: Order of @Selector() functions matters with joined selectors and setting injectContainerState: false


rwlnd:
### Description

Hi, I'm not really sure if this is `injectContainerState` related or maybe I'm just pointing out the obvious here but I just spent quiiite some time figuring out why my **joined selectors** did not resolve the correct data.

I wanted to optimize a joined `@Selector()` because it was being recalculated on EVERY state change which was completely unnecessary. So, I came across the `injectContainerState : false` section in the documentation ( https://www.ngxs.io/concepts/select#joining-selectors ) and decided to give it a go. Heck, _like that I'll even be ngxs-4 compliant already_ I was thinking. So I put `injectContainerState : false` on my configuration and refactored my selectors to only inject the state if necessary and inject other selectors.

I noticed that in the joined selector I always got the FULL state injected and not a specific slice. Which caused some unwanted errors of course.

Fast forward after quite some investigation, and trying to recreate the error in a stackblitz for the github issues section I found out that the weird behavior was because the selectors that I was injecting were declared AFTER the joined selector.

Moving the injected selector above the joined selector made the selectors act as expected.

So, as already said, maybe I'm stating the obvious here but, I think it would be helpful to have a **warning about this behavior in the documentation**. Theoretically, ordering your selectors based on dependency makes sense but I think that this will be an issue a lot of current ngxs using people will run into when they refactor towards the new default behavior of not injecting the state by default as the first parameter. While refactoring selector functions that relied on the state being injected you will not immediately think about re-arranging your selector functions.

tldr; It would be nice to add awareness of the importance of the the order of @Selector functions when using joined selectors.

## 🔬 Minimal Reproduction

I recreated the behavior in stackblitz: https://stackblitz.com/edit/angular-pax1fo
In the solution, moving the joined selector in `article.state.ts` below the 2 other selector will make this solution work as expected.

## 🌍  Your Environment
"@ngxs/store": "3.5.0"

I really hope this will at least help some people who will run into this behavior in the meanwhile.

arturovt:
Would you like to contribute and make a PR?

arturovt:
I have read your issue more thoughtfully and I have to say that the order of selectors does matter and it’s already mentioned in our docs. See https://www.ngxs.io/concepts/select#the-order-of-interacting-selectors

rwlnd:
[...]

PR #1242, opened 2019-08-17
docs(store): enhance docs about the order of selectors

arturovt:
## PR Checklist
Please check if your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://github.com/ngxs/store/blob/master/CONTRIBUTING.md#commit
- [ ] Tests for the changes have been added (for bug fixes / features)
- [x] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

```
[ ] Bugfix
[ ] Feature
[ ] Code style update (formatting, local variables)
[ ] Refactoring (no functional changes, no api changes)
[ ] Build related changes
[ ] CI related changes
[x] Documentation content changes
[ ] Other... Please describe:
```

## What is the current behavior?

Issue Number: N/A

## Does this PR introduce a breaking change?
```
[ ] Yes
[ ] No
```

## Other information
This PR improves the documentation with information about how to order selectors. This was originally described in the #1143 issue.

Issue #1143, opened 2019-07-02
Selectors should not have a specific order in a query class


t-denis:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[x] Bug report  
[ ] Performance issue
[ ] Feature request
[ ] Documentation issue or request
[ ] Support request => https://github.com/ngxs/store/blob/master/CONTRIBUTING.md
[ ] Other... Please describe:
</code></pre>

## Current behavior

A selector should be declared after it's dependencies in a query class.

## Expected behavior

Any developer-preferred selectors order.
Or an exception or, at least, a warning in the console that indicates that there is an issue with a hint on how to fix it.

## Minimal reproduction of the problem with instructions

https://stackblitz.com/edit/angular-qczwnr?file=src/app/store/counter.query.ts

## What is the motivation / use case for changing the behavior?

Happy ngxs users

## Environment

<pre><code>
Libs:
- @angular/core version: 8.0.3
- @ngxs/store version: 3.4.3

Browser:
- [x] Chrome (desktop) version XX
- [ ] Chrome (Android) version XX
- [ ] Chrome (iOS) version XX
- [ ] Firefox version XX
- [ ] Safari (desktop) version XX
- [ ] Safari (iOS) version XX
- [ ] IE version XX
- [ ] Edge version XX
 
For Tooling issues:
- Node version: XX  
- Platform:  

Others:

</code></pre>

arturovt:
@splincode @markwhitfeld any thoughts?

arturovt:
I'm confirming this is not an issue.
![image](https://user-images.githubusercontent.com/7337691/63117251-0e7b1580-bfa4-11e9-8fce-e2a9e887fc52.png)

This is not related to the NGXS itself but rather to TypeScript decorators behavior, decorators are invoked after the class declaration, so basically you're trying to access non-decorated methods yet with `@Selector`.

![image](https://user-images.githubusercontent.com/7337691/63117436-77fb2400-bfa4-11e9-9b3b-6ba77b554ce1.png)

t-denis:
Is there a way for library to check if the accessed method is decorated or not?
Maybe mention this behavior in a documentation?

arturovt:
@t-denis I understand your agitation. I'm discussing that with my NGXS teammate.

arturovt:
@t-denis I will make a PR to mention this.

markwhitfeld:
[...]

PR #1514, opened 2020-01-10
fix: selectors should not be declaration order sensitive

markwhitfeld:
## PR Checklist

## PR Type

```
[X] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: #1340, #1392, #1242, #1143 

There has been a long-standing issue with selectors where they don't work correctly if declared in an arbitrary order. The user was forced to declare referenced selectors in a class before their usage. This was even added to the docs because a solution was not possible... until now...

## What is the new behavior?
With the new internal structure of selectors and how they are given a runtime context by the store at the last possible moment the selectors could be improved to remove this order dependency. 

The one key trick to get this working was to add the selector metadata to the original function and set the prototype of the memoized function to be the original function. This allowed both the memoized function and the original function to share the exact same property descriptor for the metadata. This resolved the order dependency because the metadata becomes available from either the original or the memoized function.

```
[ ] Yes
[X] No
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 $ npm install @ngxs/store@dev
 ```
 
+- Fix: Selectors should not be declaration order sensitive [#1514](https://github.com/ngxs/store/pull/1514)
 - Fix: Selectors should be deterministic based on store being used [#1508](https://github.com/ngxs/store/pull/1508)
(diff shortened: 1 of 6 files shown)


markwhitfeld:
Will wait for @arturovt 's review before merging

Issue #759, opened 2019-01-09
🐞[BUG]: Dispatching an empty array from an @Action() sillences all future errors


thinkingmedia:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[x] Bug report  
[ ] Performance issue
[ ] Feature request
[ ] Documentation issue or request
[ ] Support request => https://github.com/ngxs/store/blob/master/CONTRIBUTING.md
[ ] Other... Please describe:
</code></pre>

## Current behavior

I ran into the problem of dispatching an action, and the subscribe method was not being fired. There was no error and the observable completed.

```
    this.store.dispatch(new Action()).subscribe(()=> console.log('next'));
```

The above example did not log anything to the console, and there was no error message logged. When I used callbacks for the next, error and done callbacks I noticed that only the done callback was executed.

```
    this.store.dispatch(new Action()).subscribe(
          ()=> console.log('this is not logged'),
          ()=> console.log('this is not logged'),
          ()=> console.log('done is logged')
    );
```

## Expected behavior

The subscriber should always be executed once unless there is an error. In which case, the default error handler will log it or you provide your own error handler.

```
    this.store.dispatch(new Action()).subscribe(()=> console.log('next'));
```

## Minimal reproduction of the problem with instructions

I have been able to reproduce the problem by having an action handler dispatch an empty array. Any action that is handled afterwards will have all thrown errors silenced.

Here is a link to a stackblitz that reproduces the problem (open the console and see that no error message is logged when you click the problem).

https://stackblitz.com/edit/angular-kzkt8z

Here is source that reproduces the problem:

```
import { Action, Selector, State, StateContext, Store } from '@ngxs/store';
import { Component } from '@angular/core';

export class ActionError { static type = 'error'};
export class ActionEmptyArray { static type = 'two'};

@State<{}>({
  name: 'app',
  defaults: {}
})
export class AppState {
  @Action(ActionError)
  public actionError() {
      throw new Error('should be shown in the console');
  }

  @Action(ActionEmptyArray)
  public actionTwo(ctx: StateContext<{}>) {
      return ctx.dispatch([]);
  }
}

@Component({
  selector: 'my-app',
  templateUrl: './app.component.html',
  styleUrls: [ './app.component.css' ]
})
export class AppComponent  {
  public constructor(public store: Store) {}

  public click() {
    this.store.dispatch([new ActionEmptyArray(), new ActionError()])
      .subscribe(()=>{
[...]

PR #1521, opened 2020-01-29
fix(store): handle empty array dispatch edge case

splincode:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: https://github.com/ngxs/store/issues/759

## What is the new behavior?

Warning output

```
[ ] Yes
[x] No
```


arturovt:
Maybe pipe with `defaultIfEmpty` if there is an empty array of actions?

splincode:
@markwhitfeld @arturovt please review again

packages/store/src/internal/dispatcher.ts
```diff
@@ -32,7 +33,7 @@ export class InternalDispatcher {
   /**
    * Dispatches event(s).
    */
-  dispatch(actionOrActions: any | any[]): Observable<any> {
+  dispatch(actionOrActions: ActionType | ActionType[]): Observable<any> {
```
arturovt: Should it be a part of this PR? Since it's a public API.

packages/store/src/internal/dispatcher.ts
```diff
@@ -49,15 +50,18 @@ export class InternalDispatcher {
     return result.pipe(leaveNgxs(this._ngxsExecutionStrategy));
   }
 
-  private dispatchByEvents(actionOrActions: any | any[]): Observable<any> {
+  private dispatchByEvents(actionOrActions: ActionType | ActionType[]): Observable<any> {
     if (Array.isArray(actionOrActions)) {
-      return forkJoin(actionOrActions.map(a => this.dispatchSingle(a)));
+      const isEmpty: boolean = actionOrActions.length === 0;
+      return isEmpty
```
arturovt: You don't need to check array length and return different observables.

```suggestion
      return forkJoin(actionOrActions.map(a => this.dispatchSingle(a))).pipe(defaultIfEmpty(this._stateStream.getValue()))
```

splincode: I tried to do this (with using defaultIfEmpty), but then all the threads unsubscribe and the tests fall.

this.dispatch([ new ActionEmptyArray(), new ActionN1(), new ActionN2() ])

In the first case, if our action returns dispatch with empty array, then the rest of the actions will not be executed due to unsubscribing

splincode: @markwhitfeld what do you think?

splincode: forkJoin([ ]).pipe( defaultIfEmpty(this._stateStream.getValue())

equals to

EMPTY.pipe( map(() => this._stateStream.getValue()) )

But we need to avoid unsubscribing

packages/store/src/internal/dispatcher.ts
splincode: sure, I will be revert it

packages/store/src/internal/dispatcher.ts
arturovt: > forkJoin([ ]).pipe( defaultIfEmpty(this._stateStream.getValue())

> equals to

> EMPTY.pipe( map(() => this._stateStream.getValue()) )

It's not true, run it in the console and you will see the difference.

> all the threads unsubscribe

What threads?

> But we need to avoid unsubscribing

What?

arturovt: Why do you lie? :disappointed: 

image

arturovt:
Also, could you move `dispatching-empty.spec.ts` to `issues/dispatching-empty.spec.ts` and add a comment before `describe` (with link to the github issue)? Thus we will know in the future what this test is related to.

markwhitfeld:
> Also, could you move `dispatching-empty.spec.ts` to `issues/dispatching-empty.spec.ts` and add a comment before `describe` (with link to the github issue)? Thus we will know in the future what this test is related to.

I like this idea! Or even  `issues/issue-759-dispatching-empty.spec.ts`

packages/store/src/internal/dispatcher.ts
markwhitfeld: I think that I prefer @splincode 's code because it makes the scenario that it is handling a bit more explicit. To be clear, I prefer this:
```ts
      const isEmpty: boolean = actionOrActions.length === 0;
      return isEmpty
        ? of(this._stateStream.getValue())
        : forkJoin(actionOrActions.map((action: ActionType) => this.dispatchSingle(action)));
```
It may be more code but I think that it is better to be explicit here. I would never expect `forkJoin` to return an empty observable in any other situation.

arturovt: As you wish.
`isEmpty` with ternary seems to be an overhead for me :smile:

markwhitfeld: @arturovt Maybe a guard clause would be better?
```ts
      if (actionOrActions.length === 0) return of(this._stateStream.getValue());
      return forkJoin(actionOrActions.map((action: ActionType) => this.dispatchSingle(action)));
```

splincode: It seems to me that such code is really better if the tests pass

```ts
forkJoin(actionOrActions.map((action: ActionType) => this.dispatchSingle(action))).pipe(
   defaultIfEmpty(this._stateStream.getValue()) // returning when forkJoin([ ])
);
```

splincode: I apparently didn’t check the code and Arthur was right

arturovt: Thank you for checking, Max!

packages/store/tests/dispatch.spec.ts
```diff
@@ -691,7 +691,7 @@ describe('Dispatch', () => {
           next: () => (nextCalled = true)
```
arturovt: also change `it should have a next value`

packages/store/tests/dispatching-empty.spec.ts
```diff
@@ -0,0 +1,181 @@
...
+    static type = 'three';
+  }
+
+  @State<{}>({
+    name: 'app',
+    defaults: {}
+  })
+  class AppState {
+    @Action(ActionError)
+    public actionError() {
```
arturovt: and remove `public` from methods :smile:

splincode:
@markwhitfeld @arturovt please review again

splincode:
@arturovt sorry, `defaultIfEmpty` is breaking change

image

splincode:
Blocked: https://github.com/ngxs/store/pull/1526

PR #1527, opened 2020-02-04
chore: prettier format

splincode:
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

## What is the new behavior?

```ts
$ yarn format
```

### DIFF
--- a/docs/advanced/mapped-sub-states.md
+++ b/docs/advanced/mapped-sub-states.md
@@ -20,21 +20,15 @@ interface Animal {
 })
 export class ZooState {
   static pandas(age: string) {
-    return createSelector(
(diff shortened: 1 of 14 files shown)


markwhitfeld:
@splincode I reverted this commit because the code became much less readable after this change.
We need to investigate if there is any way to control the way that prettier formats rxjs `.pipe(` expressions in particular. They should always be split over multiple lines.

At first I was just going to revert the docs changes (because readability is 1st prize), but then I just decided to revert the whole thing because each case the code became less readable.

Issue #1565, opened 2020-04-14
🚀[FEATURE]: Ignored Actions in Logger Plugin


armanozak:
### Relevant Package
This feature request is for [@ngxs/logger-plugin](https://github.com/ngxs/store/tree/master/packages/logger-plugin)

### Description
The logger plugin does not have an ignored actions list option. It either logs every action or does not log any action at all.

### Describe the problem you are trying to solve
Some actions are simply fired too often or not really the actions you, as developer, want to focus on or debug. When they are logged too, the console becomes cumbersome.

### Describe the solution you'd like
Passing a list of actions to plugin options which will be used for filtering some actions out before logging. The filtering function should probably be memoized in order to avoid overhead (although there may be some eventually).

### Describe alternatives you've considered
I have considered passing a custom logger via options, which is currently available, but that is not really practical.

Another option is whitelisting instead, but it does not seem to be as practical as using a blacklist in this case. Additionally, backwards compatibility requires all actions to be logged if a list is not provided and that alone may be confusing.

markwhitfeld:
I think that we should add an optional `filter` to the [options](https://github.com/ngxs/store/blob/0f5a23f4936d35ea84addcd76d9b5dc1b23dd700/packages/logger-plugin/src/symbols.ts#L3).
The `filter` would have a normal predicate signature: `(action: any) => boolean`.
If it returns true then the action would be logged.
If the filter is missing then all actions would be logged (as long it is not `disabled`).

What do you think?

armanozak:
I think this is a much better idea than just getting a list of actions to ignore. 🙂 

Nevertheless, some people will have difficulty in creating performant functions to filter actions. So, I suggest exporting predefined memoized functions for simple filters . For example:

```typescript
NgxsLoggerPluginModule.forRoot({
  filter: LogFilters.Blacklist([SomeAction, SomeOtherAction])  // will not log given actions
})
```
-- or --

```typescript
NgxsLoggerPluginModule.forRoot({
  filter: LogFilters.Whitelist([AnotherAction])  // will only log given actions
})
```

`LogFilters.Blacklist` and `LogFilters.Whitelist` are higher order functions here.

BTW, we can also consider implementing the strategy pattern.

joaqcid:
what about blacklisting at the `@Action` level ?

we could add a `log: false` to the `ActionsConfig` and control which actions to exclude from logging

markwhitfeld:
[...]

PR #1571, opened 2020-04-24
feat(logger-plugin): Action Filter in Logger Plugin

armanozak:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: #1565

## What is the new behavior?

Described in the docs.

```
[ ] Yes
[x] No
```

### DIFF
--- a/packages/logger-plugin/src/logger.module.ts
+++ b/packages/logger-plugin/src/logger.module.ts
@@ -1,8 +1,7 @@
-import { NgModule, ModuleWithProviders, InjectionToken } from '@angular/core';
+import { InjectionToken, ModuleWithProviders, NgModule } from '@angular/core';
 import { NGXS_PLUGINS } from '@ngxs/store';
-
-import { NgxsLoggerPluginOptions, NGXS_LOGGER_PLUGIN_OPTIONS } from './symbols';
 import { NgxsLoggerPlugin } from './logger.plugin';
+import { NgxsLoggerPluginOptions, NGXS_LOGGER_PLUGIN_OPTIONS } from './symbols';
 
 export const USER_OPTIONS = new InjectionToken('LOGGER_USER_OPTIONS');
 
@@ -10,7 +9,8 @@ export function loggerOptionsFactory(options: NgxsLoggerPluginOptions) {
   const defaultLoggerOptions: NgxsLoggerPluginOptions = {
     logger: console,
     collapsed: false,
-    disabled: false
+    disabled: false,
+    filter: () => true
   };
 
   return {
--- a/packages/logger-plugin/src/logger.plugin.ts
+++ b/packages/logger-plugin/src/logger.plugin.ts
@@ -1,11 +1,9 @@
-import { Injectable, Inject, Injector } from '@angular/core';
-import { tap, catchError } from 'rxjs/operators';
-
-import { NgxsPlugin, NgxsNextPluginFn, Store } from '@ngxs/store';
-
-import { NGXS_LOGGER_PLUGIN_OPTIONS, NgxsLoggerPluginOptions } from './symbols';
+import { Inject, Injectable, Injector } from '@angular/core';
+import { NgxsNextPluginFn, NgxsPlugin, Store } from '@ngxs/store';
+import { catchError, tap } from 'rxjs/operators';
 import { ActionLogger } from './action-logger';
 import { LogWriter } from './log-writer';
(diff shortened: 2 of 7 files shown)


packages/logger-plugin/src/logger.module.ts
```diff
@@ -1,16 +1,16 @@
...
 
 export const USER_OPTIONS = new InjectionToken('LOGGER_USER_OPTIONS');
 
 export function loggerOptionsFactory(options: NgxsLoggerPluginOptions) {
   const defaultLoggerOptions: NgxsLoggerPluginOptions = {
     logger: console,
     collapsed: false,
-    disabled: false
+    disabled: false,
+    filter: () => true
```
splincode, line 13: I think it would be breaking change, I think it is not necessary

packages/logger-plugin/src/logger.plugin.ts
```diff
@@ -18,7 +16,7 @@ export class NgxsLoggerPlugin implements NgxsPlugin {
   ) {}
 
   handle(state: any, event: any, next: NgxsNextPluginFn) {
-    if (this._options.disabled) {
+    if (this._options.disabled || !this._options.filter!(event, state)) {
```
splincode, line 19: `this._options.filter!` 
- I think you need to check for null

armanozak: But it cannot be `null`. The `loggerOptionsFactory` is passing in a default predicate.

packages/logger-plugin/src/logger.module.ts
armanozak: This is not a breaking change, because the option is optional.

packages/logger-plugin/src/symbols.ts
```diff
@@ -9,6 +9,9 @@ export interface NgxsLoggerPluginOptions {
 
   /** Disable the logger. Useful for prod mode. */
   disabled?: boolean;
+
+  /** Predicate for actions to be the logged. Takes action and state snapshot as parameters */
+  filter?: (action: any, state: any) => boolean;
```
splincode, line 14:
```ts
filter?: <T extends PlainObject>(action: ActionType, state: T) => boolean;
```

splincode:
Update bundlesize.json please

markwhitfeld:
@armanozak Great change.
At first I was a bit weary of passing the state through as the second param, but after reading your motivation I think that it makes sense. Nicely done!
Thank you for covering all bases including the announcement article. We will be able to release this soon!

Please just update the max size on the bundle size to the new values:
```
 FAIL  ./@ngxs/logger-plugin/fesm2015/ngxs-logger-plugin.js: 10.91KB > maxSize 10.61KB (no compression) 
 FAIL  ./@ngxs/logger-plugin/fesm5/ngxs-logger-plugin.js: 12.61KB > maxSize 12.29KB (no compression) 
```
ie. update the fesm2015 version to 10.91KB and the fesm5 one to 12.61KB
Thanks!

packages/logger-plugin/src/symbols.ts
armanozak: It would have caused a warning in Ivy if I did that, because `PlainObject` is an internal type of the store package.

armanozak:
I have updated the bundle sizes as you both have suggested @markwhitfeld and @splincode.

packages/logger-plugin/src/symbols.ts
splincode: ok

```ts
filter?: (action: ActionType, state: any) => boolean;
```

armanozak: `action` here is not the constructor, but the instance. Therefore, `ActionType` would be misleading. People would think they could access static `type` property (like `action.type`), but they cannot and they will get `undefined` instead of the type.

splincode:
@arturovt please, take part in the review

packages/logger-plugin/src/symbols.ts
markwhitfeld: I agree that `ActionType` would not work here, because an action's type could be available directly on the object or as a static property. We would not want to mislead either way.
The user would have to use the `getActionTypeFromInstance` method exposed from `@ngxs/store` in order to safely get the type if that is the information they are after.
I think that we have to stick with the `any` type for both parameters here.

markwhitfeld:
@arturovt I would love to get your review on this.

PR #1596, opened 2020-05-18
build: compile using Angular 9

splincode:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[x] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

```
[ ] Yes
[x] No
```


angular.json
```diff
@@ -14,13 +14,6 @@
             "tsConfig": "tsconfig.lib.json"
           }
         },
-        "test": {
```
markwhitfeld, line 17: Why did you remove the test configuration?

package.json
```diff
@@ -77,8 +77,8 @@
     "build:integration": "ng run integration:build:production",
     "build:integration:ssr": "yarn build:integration && ng run integration:server:production",
     "// - TESTING": "Test all modules #requires yarn build:packages",
-    "test": "ng test",
```
markwhitfeld, line 80: Why are we moving away from `ng test`?

integration/main.server.ts
```diff
@@ -1 +1,12 @@
...
+import { environment } from './environments/environment';
+
+if (environment.production) {
+  enableProdMode();
+}
+
 export { AppServerModule } from './app/app.server.module';
+export { ngExpressEngine } from '@nguniversal/express-engine';
+
+export { renderModule, renderModuleFactory } from '@angular/platform-server';
```
markwhitfeld, line 12: What is the reason for these exports and running the `enableProdMode();` in this file now?
Also, some of these exports are not used and can be removed.

integration/server.ts
```diff
@@ -1,48 +1,79 @@
...
+// '__non_webpack_require__' is a proxy to Node 'require'
+// The below code is to ensure that the server is run only when not requiring the bundle.
+// eslint-disable-next-line
+declare const __non_webpack_require__: NodeRequire;
+// eslint-disable-next-line
+const mainModule = __non_webpack_require__.main;
+const moduleFilename = (mainModule && mainModule.filename) || '';
+if (moduleFilename === __filename || moduleFilename.includes('iisnode')) {
+  run();
+}
```
markwhitfeld, line 77: What is the need for the code above?
I don't see server.ts imported anywhere...
Or, is this from a generic template that you found (in that case you should leave it matching the generic template)

integration/server.ts
```diff
@@ -1,48 +1,79 @@
...
 
-app.listen(PORT, () => {
-  console.log(`Express server is running and listening at http://localhost:${PORT}!`);
-});
+export * from './main.server';
```
markwhitfeld, line 79: Same question here, I don't see it imported anywhere...?

packages/hmr-plugin/src/tests/hmr-mock.ts
```diff
@@ -82,17 +81,10 @@ export class AppMockModuleNoHmrLifeCycle {
...
 function createRootNode(selector = 'app-root'): void {
-  const document = TestBed.get(DOCUMENT);
-  const adapter: DomAdapter = new BrowserDomAdapter();
-
-  const root = adapter.firstChild(
-    adapter.content(adapter.createTemplate(`<${selector}></${selector}>`))
-  );
-
-  const oldRoots = adapter.querySelectorAll(document, selector);
-  oldRoots.forEach(oldRoot => adapter.remove(oldRoot));
```
markwhitfeld, line 93: So, we don't need to remove the old roots anymore?

packages/store/types/tsconfig.json
```diff
@@ -19,7 +19,7 @@
     "typeRoots": ["node_modules/@types"],
     "baseUrl": ".",
     "paths": {
-      "@angular/*": ["../../../node_modules/@angular/*"],
+      "@angular/core": ["../../../node_modules/@angular/core/core.d.ts"],
```
markwhitfeld, line 22: Interesting, what is the reason for this change?

tsconfig.build.json
```diff
@@ -13,6 +13,7 @@
     }
   },
   "angularCompilerOptions": {
+    "enableIvy": false,
```
markwhitfeld, line 16: Is this needed so that our lib works for non-ivy too?

splincode: Yes, turn enableIvy only need after upgrade to Angular 10
 you will need to enable this and then, the library will be supported only by Angular 9 and Angular 10, and then the NGXS library will be supported only by Angular 9 and Angular 10

packages/store/types/tsconfig.json
splincode: It turns out that in reality the NGXS library uses only @angular/core in source code (use injectable, ngModule). Therefore, there is no need to specify any ng libraries. The most surprising surprise was that Angular now compiles its library files differently. They used to generate index.d.ts in the output. Therefore, everything worked for us. Now core.d.ts is generated (if it @angular/core), and for example testing.d.ts (if it @angular/core/testing).

packages/hmr-plugin/src/tests/hmr-mock.ts
splincode: Yes, now Angular itself destroys the root.

integration/server.ts
splincode: It needed for Angular Ivy and SSR + ViewEngine
Otherwise, it just doesn't start and doesn't work.

Official example:
https://github.com/angular/universal/blob/master/integration/hapi-engine-ve/server.ts#L69

Another examples:
https://github.com/Angular-RU/angular-universal-starter/blob/master/server.ts#L189
https://www.ganatan.com/tutorials/server-side-rendering-with-angular-universal

integration/server.ts
splincode: See my comment below:
All new projects now look like this 

https://github.com/angular/universal/blob/master/integration/hapi-engine-ve/server.ts#L69
https://github.com/Angular-RU/angular-universal-starter/blob/master/server.ts#L189

integration/main.server.ts
splincode: yes, you are right, enableProd not needed
but without the following imports, Angular SSR does not start

```ts
export { AppServerModule } from './app/app.server.module';

export { renderModule, renderModuleFactory } from '@angular/platform-server';
```

package.json
splincode: Because `@angular-builders/jest` is not supported with Jest 26 and correctly Angular CLI 9, 
so you can just call the command:

```
$ jest --config jest.config.js 
```

angular.json
splincode: Because `@angular-builders/jest` is not supported with Jest 26 and correctly Angular CLI 9, 
so you can just call the command:

```
$ jest --config jest.config.js 
```

integrations/ivy/jest.config.js
```diff
@@ -3,5 +3,10 @@ const path = require('path');
 module.exports = {
   displayName: 'ngxs-ivy',
   rootDir: path.resolve(),
-  testMatch: ['<rootDir>/src/**/**/*.spec.ts']
+  testMatch: ['<rootDir>/src/**/**/*.spec.ts'],
```
splincode, line 6: It should be need for `angular-builders@9`

package.json
splincode: good catch

angular.json
splincode: good catch

PR #1599, opened 2020-05-20
style: added generic type state in Store implementation

mazulatas:
This PR adds typing to the Store


splincode:
Also you need add dtslint spec files for check correct typings

markwhitfeld:
Similar discussion here:
https://github.com/ngxs/store/pull/649#discussion_r232485038

PR #1604, opened 2020-05-26
feat(form-plugin): add reset form action

ozanturhan:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: N/A

```
[ ] Yes
[x] No
```

## Other information
ResetForm action added for form plugin.

### DIFF
--- a/docs/plugins/form.md
+++ b/docs/plugins/form.md
@@ -130,7 +130,7 @@ The form plugin comes with the following `actions` out of the box:
 - `SetFormEnabled(path)` - Set the form to enabled
 - `SetFormDirty(path)` - Set the form to dirty (shortcut for `UpdateFormDirty`)
 - `SetFormPristine(path)` - Set the form to pristine (shortcut for `UpdateFormDirty`)
-- `ResetForm({ path, value? })` - Reset the form with value or without value.
+- `ResetForm({ path, value? })` - Reset the form with or without the form value.
 
 ### Updating Specific Form Properties
 
--- a/docs/plugins/form.md
+++ b/docs/plugins/form.md
@@ -130,6 +130,7 @@ The form plugin comes with the following `actions` out of the box:
 - `SetFormEnabled(path)` - Set the form to enabled
 - `SetFormDirty(path)` - Set the form to dirty (shortcut for `UpdateFormDirty`)
 - `SetFormPristine(path)` - Set the form to pristine (shortcut for `UpdateFormDirty`)
+- `ResetForm({ path, value?, propertyPath? })` - Reset the form with value or without value (or optionally an inner property value).
 
 ### Updating Specific Form Properties
 
@@ -183,17 +184,17 @@ The state contains information about the new novel name and its authors. Let's c
 })
 export class NewNovelComponent {
   newNovelForm: FormGroup;
-  
+
   constructor(private fb: FormBuilder) {
-		this.newNovelForm = this.fb.group({
-			novelName: 'Zenith',
-			authors: this.fb.array([
-				this.fb.group({
-					name: 'Sasha Alsberg',
-				}),
-			]),
-		});
-	}
+    this.newNovelForm = this.fb.group({
+      novelName: 'Zenith',
+      authors: this.fb.array([
+        this.fb.group({
+          name: 'Sasha Alsberg'
+        })
+      ])
+    });
+  }
 
   onSubmit() {
     //
--- a/docs/plugins/form.md
+++ b/docs/plugins/form.md
@@ -130,7 +130,7 @@ The form plugin comes with the following `actions` out of the box:
 - `SetFormEnabled(path)` - Set the form to enabled
 - `SetFormDirty(path)` - Set the form to dirty (shortcut for `UpdateFormDirty`)
 - `SetFormPristine(path)` - Set the form to pristine (shortcut for `UpdateFormDirty`)
-- `ResetForm({ path, value?, propertyPath? })` - Reset the form with value or without value (or optionally an inner property value).
+- `ResetForm({ path, value? })` - Reset the form with value or without value.
 
 ### Updating Specific Form Properties
 
--- a/packages/form-plugin/src/actions.ts
+++ b/packages/form-plugin/src/actions.ts
@@ -88,5 +88,5 @@ export class ResetForm {
     // NOTE: Not necessary to declare the type in this way in your code. See https://github.com/ngxs/store/pull/644#issuecomment-436003138
     return '[Forms] Reset Form';
   }
-  constructor(public payload: { path: string; value?: any; propertyPath?: string }) {}
+  constructor(public payload: { path: string; value?: any }) {}
 }
--- a/packages/form-plugin/src/actions.ts
+++ b/packages/form-plugin/src/actions.ts
@@ -82,3 +82,11 @@ export class SetFormEnabled {
   }
   constructor(public payload: string) {}
 }
+
+export class ResetForm {
+  static get type() {
+    // NOTE: Not necessary to declare the type in this way in your code. See https://github.com/ngxs/store/pull/644#issuecomment-436003138
+    return '[Forms] Reset Form';
+  }
+  constructor(public payload: { path: string; value?: any; propertyPath?: string }) {}
+}
--- a/packages/form-plugin/src/directive.ts
+++ b/packages/form-plugin/src/directive.ts
@@ -42,6 +42,7 @@ export class FormDirective implements OnInit, OnDestroy {
       )
       .subscribe(({ payload: { value } }: ResetForm) => {
         this.form.reset(value);
+        this.updateFormStateWithRawValue(true);
         this._cd.markForCheck();
       });
 
@@ -103,27 +104,7 @@ export class FormDirective implements OnInit, OnDestroy {
     });
 
     this._formGroupDirective.valueChanges!.pipe(this.debounceChange()).subscribe(() => {
-      const value = this._formGroupDirective.control.getRawValue();
-      this._updating = true;
-      this._store
-        .dispatch([
-          new UpdateFormValue({
-            path: this.path,
-            value
-          }),
-          new UpdateFormDirty({
-            path: this.path,
-            dirty: this._formGroupDirective.dirty
-          }),
-          new UpdateFormErrors({
-            path: this.path,
-            errors: this._formGroupDirective.errors
-          })
-        ])
-        .subscribe({
-          error: () => (this._updating = false),
-          complete: () => (this._updating = false)
-        });
+      this.updateFormStateWithRawValue();
     });
 
     this._formGroupDirective
@@ -138,6 +119,41 @@ export class FormDirective implements OnInit, OnDestroy {
       });
   }
 
+  updateFormStateWithRawValue(withFormStatus?: boolean) {
+    if (this._updating) return;
+
+    const value = this._formGroupDirective.control.getRawValue();
+
+    const actions: any[] = [
+      new UpdateFormValue({
+        path: this.path,
+        value
+      }),
+      new UpdateFormDirty({
+        path: this.path,
+        dirty: this._formGroupDirective.dirty
+      }),
+      new UpdateFormErrors({
+        path: this.path,
+        errors: this._formGroupDirective.errors
+      })
+    ];
+
+    if (withFormStatus) {
+      actions.push(
+        new UpdateFormStatus({
+          path: this.path,
+          status: this._formGroupDirective.status
+        })
+      );
+    }
+
+    this._updating = true;
+    this._store.dispatch(actions).subscribe({
+      error: () => (this._updating = false),
+      complete: () => (this._updating = false)
+    });
+  }
   ngOnDestroy() {
     this._destroy$.next();
     this._destroy$.complete();
--- a/packages/form-plugin/src/directive.ts
+++ b/packages/form-plugin/src/directive.ts
@@ -1,14 +1,15 @@
-import { Directive, Input, OnInit, OnDestroy, ChangeDetectorRef } from '@angular/core';
-import { FormGroupDirective, FormGroup } from '@angular/forms';
-import { Store, getValue } from '@ngxs/store';
-import { Subject, Observable } from 'rxjs';
-import { takeUntil, debounceTime, distinctUntilChanged } from 'rxjs/operators';
+import { ChangeDetectorRef, Directive, Input, OnDestroy, OnInit } from '@angular/core';
+import { FormGroup, FormGroupDirective } from '@angular/forms';
+import { Actions, getValue, ofActionDispatched, Store } from '@ngxs/store';
+import { Observable, Subject } from 'rxjs';
+import { debounceTime, distinctUntilChanged, filter, takeUntil } from 'rxjs/operators';
 import {
-  UpdateFormStatus,
-  UpdateFormValue,
+  ResetForm,
+  UpdateForm,
   UpdateFormDirty,
   UpdateFormErrors,
-  UpdateForm
+  UpdateFormStatus,
+  UpdateFormValue
 } from './actions';
 
 @Directive({ selector: '[ngxsForm]' })
@@ -26,12 +27,24 @@ export class FormDirective implements OnInit, OnDestroy {
   private _updating = false;
 
   constructor(
+    private _actions$: Actions,
     private _store: Store,
     private _formGroupDirective: FormGroupDirective,
     private _cd: ChangeDetectorRef
   ) {}
 
   ngOnInit() {
+    this._actions$
+      .pipe(
+        ofActionDispatched(ResetForm),
+        filter((action: ResetForm) => action.payload.path === this.path),
+        takeUntil(this._destroy$)
+      )
+      .subscribe(({ payload: { value } }: ResetForm) => {
+        this.form.reset(value);
+        this._cd.markForCheck();
+      });
+
     this.getStateStream(`${this.path}.model`).subscribe(model => {
       if (this._updating || !model) {
         return;
@@ -114,10 +127,7 @@ export class FormDirective implements OnInit, OnDestroy {
     });
 
     this._formGroupDirective
-      .statusChanges!.pipe(
-        distinctUntilChanged(),
-        this.debounceChange()
-      )
+      .statusChanges!.pipe(distinctUntilChanged(), this.debounceChange())
       .subscribe((status: string) => {
         this._store.dispatch(
           new UpdateFormStatus({
@@ -152,10 +162,7 @@ export class FormDirective implements OnInit, OnDestroy {
     return skipDebounceTime
       ? (change: Observable<any>) => change.pipe(takeUntil(this._destroy$))
       : (change: Observable<any>) =>
-          change.pipe(
-            debounceTime(this.debounce),
-            takeUntil(this._destroy$)
-          );
+          change.pipe(debounceTime(this.debounce), takeUntil(this._destroy$));
   }
 
   private get form(): FormGroup {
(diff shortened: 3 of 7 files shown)


splincode:
@ozanturhan update docs please, also

ozanturhan:
Hello @splincode I updated form.md is it enough?

ozanturhan:
@markwhitfeld Thanks a lot for your feedback.

I want to reset the form on the related page whenever the parameter changes in the resolver. I'am using the "UpdateFormValue" and "UpdateFormDirty" actions for this. But i couldn't reset form without passing all form members. You can compare the following two usages.

An example from our project:

```ts
  resolve() {
    return this.getSearchConfig().pipe(
      switchMap(() =>
        this.store.dispatch([
          new UpdateFormValue({
            value: {
              type: null,
              status: null,
              categoryId: null,
              coverage: null,
              commitmentEndDate: null,
              phoneNo: null,
              firstName: null,
              lastName: null,
              deviceType: null,
              brand: null,
              deviceCharModel: null,
              deviceCharCapacity: null,
              deviceCharColor: null,
            },
            path: 'ServiceState.searchForm',
          }),
          new UpdateFormDirty({ dirty: false, path: 'ServiceState.searchForm' }),
        ]),
      ),
      mapTo(null),
    );
  }
```
Example with ResetForm action:
```ts
  resolve() {
    return this.getSearchConfig().pipe(
      switchMap(() =>
        this.store.dispatch(
          new ResetForm({ path: 'ServiceState.searchForm' }),
        ),
      ),
      mapTo(null),
    );
  }
```

packages/form-plugin/src/directive.ts
```diff
@@ -26,12 +27,24 @@ export class FormDirective implements OnInit, OnDestroy {
...
 
   ngOnInit() {
+    this._actions$
+      .pipe(
+        ofActionDispatched(ResetForm),
+        filter((action: ResetForm) => action.payload.path === this.path),
+        takeUntil(this._destroy$)
+      )
+      .subscribe(({ payload: { value } }: ResetForm) => {
+        this.form.reset(value);
```
markwhitfeld, line 44: Should we be using `form.reset(value)` here or `form.resetForm(value)`, which will reset the validations too?

packages/form-plugin/tests/form.plugin.spec.ts
```diff
@@ -645,5 +667,123 @@ describe('NgxsFormPlugin', () => {
...
+          NgxsFormPluginModule.forRoot()
+        ],
+        declarations: [MockComponent]
+      });
+
+      const store = getStore();
+      const fixture = TestBed.createComponent(MockComponent);
+
+      fixture.detectChanges();
+
```
markwhitfeld, line 714: Could you move the `// Act` onto this line, everything before this is setup.

packages/form-plugin/tests/form.plugin.spec.ts
```diff
@@ -645,5 +667,123 @@ describe('NgxsFormPlugin', () => {
...
+      store.dispatch(new ResetForm({ path: 'todos.todosForm' }));
+
+      fixture.detectChanges();
+
```
markwhitfeld, line 718: Also please place the `// Arrange` on this line, because the `input` variable is only required for the assertions and is not directly part of the action we are testing.

packages/form-plugin/tests/form.plugin.spec.ts
```diff
@@ -645,5 +667,123 @@ describe('NgxsFormPlugin', () => {
...
+          NgxsFormPluginModule.forRoot()
+        ],
+        declarations: [MockComponent]
+      });
+
+      const store = getStore();
+      const fixture = TestBed.createComponent(MockComponent);
+
+      fixture.detectChanges();
+
```
markwhitfeld, line 771: Could you move the `// Act` onto this line, everything before this is setup?

packages/form-plugin/tests/form.plugin.spec.ts
```diff
@@ -645,5 +667,123 @@ describe('NgxsFormPlugin', () => {
...
+      store.dispatch(
+        new ResetForm({ path: 'todos.todosForm', value: { text: 'Buy some tea' } })
+      );
+
+      fixture.detectChanges();
+
```
markwhitfeld, line 777: Also please place the `// Arrange` on this line, because the `input` variable is only required for the assertions.

packages/form-plugin/src/actions.ts
```diff
@@ -82,3 +82,11 @@ export class SetFormEnabled {
   }
   constructor(public payload: string) {}
 }
+
+export class ResetForm {
+  static get type() {
+    // NOTE: Not necessary to declare the type in this way in your code. See https://github.com/ngxs/store/pull/644#issuecomment-436003138
+    return '[Forms] Reset Form';
+  }
+  constructor(public payload: { path: string; value?: any; propertyPath?: string }) {}
```
markwhitfeld, line 91: The `propertyPath` is unused, please remove it.

packages/form-plugin/src/form.plugin.ts
```diff
@@ -24,13 +26,18 @@ export class NgxsFormPlugin implements NgxsPlugin {
...
+    if (type === UpdateFormValue.type || type === UpdateForm.type || type === ResetForm.type) {
       const { value } = event.payload;
       const payloadValue = Array.isArray(value) ? value.slice() : { ...value };
       const path = this.joinPathWithPropertyPath(event);
       nextState = setValue(nextState, path, payloadValue);
     }
 
+    if (type === ResetForm.type) {
+      const model = getValue(nextState, `${event.payload.path}.model`);
+      nextState = setValue(nextState, `${event.payload.path}`, { model: model });
```
markwhitfeld, line 38: What is the intention of this code here?
To wipe out the other properties for the form?
It would wipe out the `status`, `errors`, `dirty` and `disabled` properties. Not sure if this is what we want.
Surely all of the properties will be reset when the `form.reset` happens, or maybe the `form.resetForm` would make this not needed anymore. Maybe after making that change remove this and see if you get the behaviour you are intending.

packages/form-plugin/src/actions.ts
ozanturhan: Thanks for your comment. But this property is used in form.plugin.ts "joinPathWithPropertyPath" function. You can check line 29 in form.plugn.ts. We have a condition:

```typescript
if (type === UpdateFormValue.type || type === UpdateForm.type || type === ResetForm.type) {
      const { value } = event.payload;
      const payloadValue = Array.isArray(value) ? value.slice() : { ...value };
      const path = this.joinPathWithPropertyPath(event);
      nextState = setValue(nextState, path, payloadValue);
    }
```

This action can be used with property value like "UpdateFormValue" and "UpdateForm" actions.

If you think this is not necessary, I can remove it.

packages/form-plugin/src/directive.ts
ozanturhan: We should use form.reset because If we don't use it form values dirty and errors returns previous state. Because directive have a debounce functionality.

Can you check line 98 in directive.ts.

If we use RestAction without form reset
- state will be reset
- observable will spread a value because of the debounce
- UpdateFormValue etc. will be called with form raw value
```typescript
const value = this._formGroupDirective.control.getRawValue();
```

packages/form-plugin/src/form.plugin.ts
ozanturhan: You alright. I have to work on it a little bit. We can update form state after from.reset called in directive like debounce change subscriptions.

I think we create new function for update state and use it in debounce subscription and after form reset.

```typescript
  updateFormStateWithRawValue() {
    const value = this._formGroupDirective.control.getRawValue();
    this._updating = true;
    this._store
      .dispatch([
        new UpdateFormValue({
          path: this.path,
          value
        }),
        new UpdateFormDirty({
          path: this.path,
          dirty: this._formGroupDirective.dirty
        }),
        new UpdateFormErrors({
          path: this.path,
          errors: this._formGroupDirective.errors
        })
      ])
      .subscribe({
        error: () => (this._updating = false),
        complete: () => (this._updating = false)
      });
  }
```
Debounce Change
```typescript
this._formGroupDirective.valueChanges!.pipe(this.debounceChange()).subscribe(() => {
      this.updateFormStateWithRawValue();
});
```
After Form Reset
```typescript
    this._actions$
      .pipe(
        ofActionDispatched(ResetForm),
        filter((action: ResetForm) => action.payload.path === this.path),
        takeUntil(this._destroy$)
      )
      .subscribe(({ payload: { value } }: ResetForm) => {
        this.form.reset(value);
        this.updateFormStateWithRawValue();
        this._cd.markForCheck();
      });
```

How about that?

ozanturhan: This code worked correctly. I think it was better like this. I will commit my changes if it is suitable for you.

packages/form-plugin/tests/form.plugin.spec.ts
```diff
@@ -207,6 +208,27 @@ describe('NgxsFormPlugin', () => {
...
+    store.dispatch([
+      new UpdateForm({
+        errors: {},
+        dirty: true,
+        status: 'INVALID',
+        value: { name: 'Ozan Turhan' },
+        path: 'student.studentForm'
+      }),
+      new ResetForm({ path: 'student.studentForm' })
+    ]);
```
markwhitfeld, line 224: In order to test what the interactions would be after some time (due to debounce), please could you await this observable before doing the `selectSnapshot`.
```suggestion
    await store.dispatch([
      new UpdateForm({
        errors: {},
        dirty: true,
        status: 'INVALID',
        value: { name: 'Ozan Turhan' },
        path: 'student.studentForm'
      }),
      new ResetForm({ path: 'student.studentForm' })
    ]).toPromise();
```

markwhitfeld:
If you could answer these questions with tests in this PR, I would be really impressed ;-)

ozanturhan:
@markwhitfeld Thank you for your comments. I will be working on this weekend.

ozanturhan:
Hi @markwhitfeld,

I have a question. I couldn't decide what it should be.

What should be expectation in below test case?
```ts
    await store.dispatch([
      new UpdateForm({
        errors: { something: true },
        dirty: true,
        status: 'INVALID',
        value: { name: 'Ozan Turhan' },
        path: 'student.studentForm'
      }),
      new ResetForm({ path: 'student.studentForm' })
    ]).toPromise();
```

When we run test without form, should these fields to be reset (errors, dirty, status etc.) ? I think it shouldn't be change. Because these parameters related to form. 

For example in this test case ResetForm action changes only model. Because there is no form also ngxsForm directive.

## Student Form Before Reset:
```
{
  dirty: true,
  errors: { something: true  },
  model: {
      name: "Ozan Turhan",
  },
  status: "INVALID",
}
```

## Student Form After Reset:
```
{
  dirty: true,
  errors: { something: true  },
  model: {},
  status: "INVALID",
}
```

But if we are run test with runtime behavior, also if we have a form with ngxsForm directive studentForm state will be like below (after reset form action called):

```
{
  dirty: false,
  errors: {},
  model: {
      name: null
  },
  status: "VALID",
}
```
Example Form:
```ts
      @Component({
        template: `
          <form [formGroup]="form" ngxsForm="todos.todosForm">
            <input formControlName="text" /> <button type="submit">Add todo</button>
          </form>
        `
      })
      class MockComponent {
        public form = new FormGroup({
          text: new FormControl()
        });
      }
```

I think runtime behavior is true. But I'm not sure about the first case (without ngxsDirective).

After your anwser i will be start to writing to the tests.

If you can help me to make decision for that, I would be glad.

Regards

packages/form-plugin/tests/form.plugin.spec.ts
arturovt: You can add `[ngxsFormDebounce]="-1"` as binding. I prefer synchronous code over asynchronous if it's possible. It's easier to test it.

ozanturhan: Hi @arturovt in this case we don't have a form as you can see:

```ts
  it('should reset form', fakeAsync(async() => {
    // Arrange & act
    const store = getStore();

    await store.dispatch([
      new UpdateForm({
        errors: {},
        dirty: true,
        status: 'INVALID',
        value: { name: 'Ozan Turhan' },
        path: 'student.studentForm'
      }),
      new ResetForm({ path: 'student.studentForm'})
    ]).toPromise();

    const form = store.selectSnapshot(StudentState.getStudentForm);

    // Assert
    expect(form).toEqual({});
  }));
```

so we can't use `ngxsFormDebounce` in this state. But i can write test with form and `[ngxsFormDebounce]="-1"` together if is necessary. (with runtime behavior)

arturovt: Oh. I see. Then you can use single `async/await` w/o `fakeAsync`.

markwhitfeld:
Apologies for the delay in getting back to you!

This is a very interesting case that you raise.
My gut feel is that if you are not bound with an `ngxsForm` then the model and everything else should be reset. ie. not dirty, not touched, no errors, reset status, etc.

Conceptually, when you dispatch actions there should be no observable difference in the state if something is bound to that state or not. In this case there is a subtle difference, which essentially is that if you are bound to a form then the model resets to an object with empty properties (as described in the form) instead of an empty object (`{}`). I am willing to live with this because there is no way for the state to know what properties are on the form if it has never been bound.
Do you agree?

markwhitfeld:
PS. With regard to the debounce. I think that this action should result in an immediate change to the state and the form, and that any pending changes that are being debounced should be ignored.
For example... let's say that we had a debounce of 500ms and we type into one of the inputs.
The model in the state would only be updated after 500ms as expected.
But, if we were to dispatch the `ResetForm` action before that 500ms was up then the input change should never reach the state.

If we do not address this then we could have this strange scenario (500ms debounce):
- user types input
- after 200ms, `ResetForm` is dispatched by some code
- form model and all other props are rest
- at 500ms the new value arrives in the state!
- this would not be expected

ozanturhan:
@splincode Thanks a lot for your support. 

> Apologies for the delay in getting back to you!
> 
> This is a very interesting case that you raise.
> My gut feel is that if you are not bound with an `ngxsForm` then the model and everything else should be reset. ie. not dirty, not touched, no errors, reset status, etc.
> 
> Conceptually, when you dispatch actions there should be no observable difference in the state if something is bound to that state or not. In this case there is a subtle difference, which essentially is that if you are bound to a form then the model resets to an object with empty properties (as described in the form) instead of an empty object (`{}`). I am willing to live with this because there is no way for the state to know what properties are on the form if it has never been bound.
> Do you agree?

Yes i agree. Form state should be reset without bounding to `ngxsForm`. Reset action will reset dirty, status etc. 

But if we have ngsxForm state will be reseting by form raw values. 

I got it. I covered these cases with unit tests. I am going to commit these tests.



> PS. With regard to the debounce. I think that this action should result in an immediate change to the state and the form, and that any pending changes that are being debounced should be ignored.
> For example... let's say that we had a debounce of 500ms and we type into one of the inputs.
> The model in the state would only be updated after 500ms as expected.
> But, if we were to dispatch the `ResetForm` action before that 500ms was up then the input change should never reach the state.
> 
> If we do not address this then we could have this strange scenario (500ms debounce):
> 
> * user types input
> * after 200ms, `ResetForm` is dispatched by some code
> * form model and all other props are rest
> * at 500ms the new value arrives in the state!
> * this would not be expected

Also i added new test case for this.

packages/form-plugin/src/actions.ts
markwhitfeld: Yes, please do remove the `propertyPath`. I think that a reset conceptually applies to the whole form therefore giving it a value just for a single internal property does not make sense. Giving it a value to reset the whole form to does make a lot of sense.

markwhitfeld: Yes, please do remove the propertyPath. I think that a reset conceptually applies to the whole form therefore giving it a value just for a single internal property does not make sense. Giving it a value to reset the whole form to does make a lot of sense.

ozanturhan:
> Please remove the `propertyPath` because, as stated below, I don't think it makes sense in this case.
> Do you agree?
> Are there any other aspects you feel that we still need to discuss/cover?
> 
> PS. I really appreciate and find it really encouraging that you have engaged so well about getting this feature implemented. Often people drop a PR and are not willing to discuss or put extra effort in. You have really gone the extra mile! Your contribution is definitely a highlight in the often thankless task of open source work

Okay, you are right. i will remove it tomorrow. I don't think there's anything left to discuss.

Thank you very much for your positive comments.

Thank you for your interest and support in this PR. This will be my second contribution to an open source project. It was a very good experience for me. Also your comments were useful and I really learned something.

markwhitfeld:
Thank you so much for this work. I would like to release this soon in the next version. 
Would you mind adding one more thing to this PR?
Could you add a section to the announcement article with a brief overview of this feature.
Edit this file: `publication/announcing_ngxs_next/article.md`

ozanturhan:
@markwhitfeld of course i will add as soon as possible.

docs/plugins/form.md
```diff
@@ -130,6 +130,7 @@ The form plugin comes with the following `actions` out of the box:
 - `SetFormEnabled(path)` - Set the form to enabled
 - `SetFormDirty(path)` - Set the form to dirty (shortcut for `UpdateFormDirty`)
 - `SetFormPristine(path)` - Set the form to pristine (shortcut for `UpdateFormDirty`)
+- `ResetForm({ path, value? })` - Reset the form with value or without value.
```
markwhitfeld, line 133:
```suggestion
- `ResetForm({ path, value? })` - Reset the form with or without the form value.
```

publication/announcing_ngxs_next/article.md
```diff
@@ -114,6 +114,52 @@ You can define your own logic before or after the state get serialized or deseri
...
 ```
 
+### Form Plugin
+
+- Feature: Reset Form Action [#1604](github.com/ngxs/store/pull/1604)
+
+You can reset form with `ResetForm` action.
+
+- This action reset your form and related form state.
+- Form status, dirty, values etc. will be reset by related form values after calling this action.
```
markwhitfeld, line 124:
```suggestion
You can reset the form with the `ResetForm` action.

- This action resets the form and related form state.
- The form status, dirty, values, etc. will be reset after calling this action.
```

publication/announcing_ngxs_next/article.md
```diff
@@ -114,6 +114,52 @@ You can define your own logic before or after the state get serialized or deseri
...
+  resetForm() {
+    this.store.dispatch(new ResetForm({ path: 'exampleState.form' }));
+  }
+
+  resetFormWithValue() {
+    this.store.dispatch(
+      new ResetForm({
+        path: 'exampleState.form',
+        value: {
+          text: 'Defaul Text',
```
markwhitfeld, line 155:
```suggestion
          text: 'Default Text',
```

packages/form-plugin/src/form.plugin.ts
markwhitfeld: Is this discussion resolved? I think that this was covered by later changes and tests?

ozanturhan: Yes this discussion resolved. This is required and this is covered with tests. Because if we are reset form state without form and ngxsForm directive this code block resets our state.

I asked this case, if i understood correctly this is your answer:

> My gut feel is that if you are not bound with an `ngxsForm` then the model and everything else should be reset. ie. not dirty, not touched, no errors, reset status, etc.

nickroberts:
This is great! We are currently using the form plugin for several different forms on a single page, and resetting them is a huge pain.

Any idea when the next release will be? I was hoping this was in the latest dev version, and I was going to use that, but it doesn't look like it is, as that version was from May

Keep up the great work!

Issue #1551, opened 2020-03-15
🐞[BUG]: Incorrect definition of payload in logging


splincode:
### Affected Package

 The issue is caused by package @ngxs/logger-plugin

### Description

https://github.com/ngxs/store/blob/master/packages/logger-plugin/src/action-logger.ts#L40

## 🔬 Minimal Reproduction

```ts
export class RemoveTodoAction {
  public static type: string = '@todo.removeTodo(idx)';
  constructor(public idx: number) {}
}
```

```ts
this.store.dispatch(new RemoveTodoAction( 1 ));
```

![image](https://user-images.githubusercontent.com/12021443/76709191-bd83f900-670d-11ea-8824-1a6fd0ec4b5c.png)

```ts
this.store.dispatch(new RemoveTodoAction( 0 ));
```

![image](https://user-images.githubusercontent.com/12021443/76709203-ddb3b800-670d-11ea-8881-aceddf29a3ab.png)

markwhitfeld:
Good news, this has been addressed in NGXS v3.7 which was released last week.
( see https://dev.to/ngxs/announcing-ngxs-3-7-1kbk for release notes )

I will close this issue. Please let us know if it is not resolved.

PR #1617, opened 2020-06-14
fix(logger-plugin): filter out only `undefined` payloads

arturovt:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: #1551 

```
[ ] Yes
[x] No
```

### DIFF
--- a/packages/logger-plugin/tests/helpers/setup.ts
+++ b/packages/logger-plugin/tests/helpers/setup.ts
@@ -1,0 +1,30 @@
+import { ErrorHandler } from '@angular/core';
+import { TestBed } from '@angular/core/testing';
+import { NgxsModule, Store } from '@ngxs/store';
+import { StateClass } from '@ngxs/store/internals';
+
+import { LoggerSpy } from './logger-spy';
+import { NoopErrorHandler } from '../../../store/tests/helpers/utils';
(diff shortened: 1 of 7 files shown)


packages/logger-plugin/tests/helpers/setup.ts
```diff
@@ -0,0 +1,30 @@
+import { ErrorHandler } from '@angular/core';
+import { TestBed } from '@angular/core/testing';
+import { NgxsModule, Store } from '@ngxs/store';
+import { StateClass } from '@ngxs/store/internals';
+
+import { LoggerSpy } from './logger-spy';
+import { NoopErrorHandler } from '../../../store/tests/helpers/utils';
+import { NgxsLoggerPluginModule, NgxsLoggerPluginOptions } from '../../';
+
+export function setup(states: StateClass[], opts?: NgxsLoggerPluginOptions) {
```
markwhitfeld, line 10: The method name `setup` is very common to use within tests, so could we change this name to be more specific to the type of setup so that it does not conflict with the convention (and potentially cause conflicts with the `setup` method in tests where we would want to build on top of this)?
Obviously the usages will also have to change with this rename.
```suggestion
export function setupWithLogger(states: StateClass[], opts?: NgxsLoggerPluginOptions) {
```

PR #1622, opened 2020-06-16
docs(optimizing-selectors): create page

troydietz:
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: N/A

```
[ ] Yes
[x] No
```

### DIFF
--- a/docs/advanced/optimizing-selectors.md
+++ b/docs/advanced/optimizing-selectors.md
@@ -54,8 +54,7 @@ export class SomeState {
   }
 }
 
-@Injectable()
-export class SomeSelectorService {
+export class SomeSelectors {
   @Selector([SomeState.getData])
   static getViewData(data: Data[]) {
     return data.map(d => expensiveFunction(d));
--- a/docs/advanced/optimizing-selectors.md
+++ b/docs/advanced/optimizing-selectors.md
@@ -36,7 +36,7 @@ static getData(state: SomeStateModel) {
 
 @Selector([SomeState.getData])
 static getViewData(data: Data[]) {
-   return data.map(d => expensiveFunction(d));
+  return data.map(d => expensiveFunction(d));
 }
 ```
 
--- a/docs/advanced/optimizing-selectors.md
+++ b/docs/advanced/optimizing-selectors.md
@@ -54,7 +54,7 @@ export class SomeState {
   }
 }
 
-export class SomeSelectors {
+export class SomeStateQueries {
   @Selector([SomeState.getData])
   static getViewData(data: Data[]) {
     return data.map(d => expensiveFunction(d));
--- a/docs/advanced/optimizing-selectors.md
+++ b/docs/advanced/optimizing-selectors.md
@@ -26,12 +26,12 @@ static getViewData(state: SomeStateModel) {
 
 Selectors defined in state classes implicitly have `state` injected as their first argument. The above selector will be recalculated every time the user types into the input component. Since `state` could update rapidly when a user types, the expensive selector will needlessly recalculate even though it does not care about the `name` property of `state` changing. This selector does not take advantage of memoization.
 
-One way to solve this problem is to turn off the `injectContainerState` selector [option](options.md). This option does not apply to selectors declared outside of state classes. This option only applies to selectors declared within state classes by preventing state being implicitly injected as the first argument. This requires you to explicitly specify all arguments using the `@Selector([...])` decorator _to selectors that are declared in state classes_. For example, we create two selectors in our state class:
+One way to solve this problem is to turn off the `injectContainerState` selector [option](options.md) at root, state, or selector level. By default (in NGXS v3), the state is implicitly injected as the first argument for composite selectors _defined within state classes_. Turning off this setting prevents the container state from being injected as the first argument. This requires you to explicitly specify all arguments when you use the `@Selector([...])` decorator. Any parameterless `@Selector()` decorators will still inject the state as an implicit argument. Note that this option does not apply to selectors declared _outside of state classes_ (because there is no container state to inject). For example, we create two selectors in our state class:
 
 ```ts
 @Selector([SomeState])
 static getData(state: SomeStateModel) {
-  return state.data;
+   return state.data;
 }
 
 @Selector([SomeState.getData])
--- a/docs/advanced/optimizing-selectors.md
+++ b/docs/advanced/optimizing-selectors.md
@@ -26,7 +26,7 @@ static getViewData(state: SomeStateModel) {
 
 Selectors defined in state classes implicitly have `state` injected as their first argument. The above selector will be recalculated every time the user types into the input component. Since `state` could update rapidly when a user types, the expensive selector will needlessly recalculate even though it does not care about the `name` property of `state` changing. This selector does not take advantage of memoization.
 
-One way to solve this problem is to turn off the `injectContainerState` selector option. This enables you to explicitly specify arguments using the `@Selector` decorator. For example:
+One way to solve this problem is to turn off the `injectContainerState` selector [option](options.md). This option does not apply to selectors declared outside of state classes. This option only applies to selectors declared within state classes by preventing state being implicitly injected as the first argument. This requires you to explicitly specify all arguments using the `@Selector([...])` decorator _to selectors that are declared in state classes_. For example, we create two selectors in our state class:
 
 ```ts
 @Selector([SomeState])
@@ -42,7 +42,7 @@ static getViewData(data: Data[]) {
 
 This `getViewData` selector will not be recalculated when a user types into the input component. This selector targets the specific property of `state` it cares about as its argument by leveraging an additional selector. When the `name` property of state changes, the `getViewData` arguments _do not change_. Memoization is taken advantage of.
 
-An alternative solution to turning off the selector option is to create selectors outside of your state classes.
+An alternative solution to turning off the selector option is to create a [meta selector](../concepts/select#meta-selectors). For example, we declare one selector in our state class and declare another selector outside of our state class:
 
 ```ts
 @State({...})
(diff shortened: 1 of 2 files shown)


troydietz:
The force push is a consequence of me squashing an additional typo commit into the original.

docs/advanced/optimizing-selectors.md
```diff
@@ -0,0 +1,110 @@
...
+```ts
+@Selector()
+static getViewData(state: SomeStateModel) {
+   return state.data.map(d => expensiveFunction(d));
+}
+```
+
+Selectors defined in state classes implicitly have `state` injected as their first argument. The above selector will be recalculated every time the user types into the input component. Since `state` could update rapidly when a user types, the expensive selector will needlessly recalculate even though it does not care about the `name` property of `state` changing. This selector does not take advantage of memoization.
+
+One way to solve this problem is to turn off the `injectContainerState` selector option. This enables you to explicitly specify arguments using the `@Selector` decorator. For example:
```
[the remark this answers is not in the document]
troydietz: Good call.  That additional information will definitely make things clearer.

docs/advanced/optimizing-selectors.md
```diff
@@ -0,0 +1,110 @@
...
+
+@Selector([SomeState.getData])
+static getViewData(data: Data[]) {
+   return data.map(d => expensiveFunction(d));
+}
+```
+
+This `getViewData` selector will not be recalculated when a user types into the input component. This selector targets the specific property of `state` it cares about as its argument by leveraging an additional selector. When the `name` property of state changes, the `getViewData` arguments _do not change_. Memoization is taken advantage of.
+
+An alternative solution to turning off the selector option is to create selectors outside of your state classes.
```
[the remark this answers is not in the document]
troydietz: Agreed 👍🏽

docs/advanced/optimizing-selectors.md
```diff
@@ -0,0 +1,110 @@
...
+And you have this selector:
+
+```ts
+@Selector([SelectedDataState])
+isDataSelected(state: SelectedDataStateModel) {
+  return (id: number) => state.selectedIds.includes(id);
+}
+```
+
+The selector returns a function, which accepts an id as an argument and returns a boolean indicating whether or not this id is selected. In this example, we want to render a list of checkboxes.
```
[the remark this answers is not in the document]
troydietz: Agreed 👍🏽

docs/advanced/optimizing-selectors.md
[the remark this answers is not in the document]
troydietz: After some more thought, I think its more correct to convey the following (I'm still working out the verbiage):

This option does not affect the behavior of selectors declared outside of state classes.
This option only affects selectors declared within state classes by preventing state being implicitly injected as the first argument.  This requires you to explicitly specify all arguments using the `@Selector([...])` decorator _that are declared in state classes_.

Motivation:
> Maybe one aside note here is that this setting does not apply to parameterless selectors eg. @Selector().

Since it's possible to have parameterless selectors in state classes., e.g.
```ts
@State({ ... })
@Injectable()
class State {
``` 
Turning the option off, those selectors could still benefit from removing state

docs/advanced/optimizing-selectors.md
```diff
@@ -0,0 +1,110 @@
...
+
+When users check or uncheck an item, `state.selectedIds` is updated, therefore the `isDataSelected` selector is recalculated and the list must re-render. Every time the list re-renders, `isDataSelected` is invoked `data.length` number of times. Because the selector implementation uses [Array.includes](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes). This template renders in `O(n^2)` time complexity - **Ugh!**. One magnitude of `n` for the length of data , another for `state.selectedIds.length`.
+
+Here's one way to improve performance in that example:
+
+```ts
+@Selector([SelectedDataState])
+isDataSelected(state: SelectedDataStateModel) {
+  const selectedIds = new Set(state.selectedIds);
+  return (id: number) => selectedIds.has(id);
```
bbarry, line 106: I think a comment here would be useful to explicitly point out the fact that you are returning a closure over the `selectedIds` variable. It is not immediately obvious why this is better.

troydietz: You’re absolutely correct.  I will make note of this.  Thanks for the feedback!

troydietz:
Thank you, I appreciate the feedback everyone.  I've addressed the above comments with my latest commit.  Please let me know what you think.

docs/advanced/optimizing-selectors.md
```diff
@@ -0,0 +1,112 @@
...
+@Injectable()
+export class SomeState {
+  @Selector()
+  static getData(state: SomeStateModel) {
+    return state.data;
+  }
+}
+
+@Injectable()
+export class SomeSelectorService {
```
arturovt, line 58: I wouldn't call it service and it's definitely shouldn't have the injectable decorator since only static methods will be used. I remember we had examples in our docs with a class that has exposed selectors and its name was suffixed with "Selectors". I would like just to keep it consistent :slightly_smiling_face:

docs/advanced/optimizing-selectors.md
```diff
@@ -0,0 +1,112 @@
...
+```
+
+Selectors defined in state classes implicitly have `state` injected as their first argument. The above selector will be recalculated every time the user types into the input component. Since `state` could update rapidly when a user types, the expensive selector will needlessly recalculate even though it does not care about the `name` property of `state` changing. This selector does not take advantage of memoization.
+
+One way to solve this problem is to turn off the `injectContainerState` selector [option](options.md). This option does not apply to selectors declared outside of state classes. This option only applies to selectors declared within state classes by preventing state being implicitly injected as the first argument. This requires you to explicitly specify all arguments using the `@Selector([...])` decorator _to selectors that are declared in state classes_. For example, we create two selectors in our state class:
+
+```ts
+@Selector([SomeState])
+static getData(state: SomeStateModel) {
+   return state.data;
```
arturovt, line 34:
```suggestion
  return state.data;
```

docs/advanced/optimizing-selectors.md
```diff
@@ -0,0 +1,112 @@
...
+}
+
+@Selector([SomeState.getData])
+static getViewData(data: Data[]) {
+   return data.map(d => expensiveFunction(d));
```
arturovt, line 39:
```suggestion
  return data.map(d => expensiveFunction(d));
```

docs/advanced/optimizing-selectors.md
troydietz: Excellent catch 🎣  Thank you!

I will remove `@Injectable()` and rename this class to `SomeSelectors` to remain consistent with the other class names in that example.

docs/advanced/optimizing-selectors.md
```diff
@@ -0,0 +1,112 @@
...
+@State({...})
+@Injectable()
+export class SomeState {
+  @Selector()
+  static getData(state: SomeStateModel) {
+    return state.data;
+  }
+}
+
+@Injectable()
```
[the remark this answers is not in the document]
troydietz: @arturovt beat you too it 😄  You guys are good.  The most recent commit should have this resolved.

docs/advanced/optimizing-selectors.md
```diff
@@ -0,0 +1,112 @@
...
+
+Selectors defined in state classes implicitly have `state` injected as their first argument. The above selector will be recalculated every time the user types into the input component. Since `state` could update rapidly when a user types, the expensive selector will needlessly recalculate even though it does not care about the `name` property of `state` changing. This selector does not take advantage of memoization.
+
+One way to solve this problem is to turn off the `injectContainerState` selector [option](options.md). This option does not apply to selectors declared outside of state classes. This option only applies to selectors declared within state classes by preventing state being implicitly injected as the first argument. This requires you to explicitly specify all arguments using the `@Selector([...])` decorator _to selectors that are declared in state classes_. For example, we create two selectors in our state class:
+
+```ts
+@Selector([SomeState])
+static getData(state: SomeStateModel) {
+   return state.data;
+}
```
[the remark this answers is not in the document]
troydietz: Sounds good to me.  Thank you for writing that up!  Stay tuned for an update

troydietz:
I believe the most recent commit addresses all outstanding comments and change requests.  Please don't hesitate to provide anymore feedback.  Thank you for your time.

docs/advanced/optimizing-selectors.md
markwhitfeld: PS. I actually prefer to name these classes with the `Queries` suffix.
For example `TodoQueries`

docs/advanced/optimizing-selectors.md
```diff
@@ -0,0 +1,111 @@
...
+@State({...})
+@Injectable()
+export class SomeState {
+  @Selector()
+  static getData(state: SomeStateModel) {
+    return state.data;
+  }
+}
+
+export class SomeSelectors {
```
markwhitfeld, line 57: I really like this convention to suffix with `Queries` for these meta selectors.
What do you think?
```suggestion
export class SomeStateQueries {
```

troydietz: @arturovt also requested to change this class name (to what it is now).  I'm not sure there is a convention on the ngxs website.

On my team we do:

```export class ProductsSelectors```

markwhitfeld: Hmmm... it is really a matter of preference.
I feel like the word "selector" gets really overused all over my codebase so I opted for something similar that can carry the meaning of the Q from CQRS ;-)
I don't feel strongly either way! 🤷‍♂️

troydietz: That definitely makes sense 👍

troydietz:
It appears this merge commit did not pass the one of your integration tests. https://github.com/ngxs/store/commit/7b45724344e797f9b9d25d6823ae4dc3d7140bcf. If there's anything I can do to help, please let me know.

markwhitfeld:
@troydietz looks like just a circleci workspace failure.
I have restarted it, and it should be good 😉

Issue #1532, opened 2020-02-10
🐞[BUG]: Cannot assign @Select to private property


twjacobsen:
### Affected Package

 The issue is caused by package @ngxs/5.6.2

### Is this a regression?

 Yes, it worked just fine in 3.4.3

### Description

 A clear and concise description of the problem...
Using the `@Select` decorator to assign to private properties of another class results in this error:
```
Argument of type 'HelloComponent' is not assignable to parameter of type 'ComponentClass<any> & Record<"_name", Observable<string>>'.
  Type 'HelloComponent' is not assignable to type 'Record<"_name", Observable<string>>'.
    Property '_name' is private in type 'HelloComponent' but not in type 'Record<"_name", Observable<string>>'.
```

## 🔬 Minimal Reproduction

 
https://stackblitz.com/edit/ngxs-at-select-private-property?file=src%2Fapp%2Fhello.component.ts

## Environment

<pre><code>
Libs:
- @angular/core version: 8.2.4
- @ngxs/store version: 5.6.2
- Typescript version: 3.7.5

splincode:
please create stackblitz or guthub repo with reproduce

twjacobsen:
There is a link under "Minimal reproduction"... IDK why it won't make it an actual link..

markwhitfeld:
@splincode I believe that this would have been caused by the extra typings you added in this commit:
https://github.com/ngxs/store/pull/1436/commits/ca7de82b3032057743dbd1cb4e50a5cf934011d1
Are you able to improve the types to handle private fields?

splincode:
Sorry, I read the description poorly and did not look at the link

splincode:
Sorry, I read the description poorly and did not look at the link.

@markwhitfeld I will investigate

steve-todorov:
I'm getting this issue as well. In our case we have an `AbstractAuthorityGuard` which has a "default" implementation of `CanActivate` and allows for other `Guard`s to extend the abstract class and do minor additional checks. Sort of like this oversimplified example:

```
@Injectable({
    providedIn: 'root'
})
export abstract class AbstractAuthorityGuard implements CanActivate {
    @Select(SessionState.authorities)
    protected authorities$: Observable<UserAuthority[]>;

    canActivate(): Observable<boolean> {
        return this.authorities$;
    }
}
```

steve-todorov:
Hey guys, any update on this? :)

splincode:
[...]

PR #1623, opened 2020-06-17
fix(store): remove SelectType<T> implementation due to cannot assign @Select to private/protected property

splincode:
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: #1532 

```
[ ] Yes
[ ] No
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ $ npm install @ngxs/store@dev
 - Feature: Logger Plugin - Add filter for Logger Plugin [#1571](https://github.com/ngxs/store/pull/1571)
 - Performance: Logger Plugin - Plugin should lazy inject the store once [#1550](https://github.com/ngxs/store/pull/1550)
 - Fix: `ofAction*` methods should prevent passing anything except of `ActionType` [#1616](https://github.com/ngxs/store/pull/1616)
-- Fix: remove SelectType<T> implementation due to cannot assign @Select to private/protected property [#1623](https://github.com/ngxs/store/pull/1623)
+- Fix: Remove the recent `@Select` type safety check due to issues with private/protected properties [#1623](https://github.com/ngxs/store/pull/1623)
 - Fix: Actions are not canceled when any `Observable` returned by any handler is completed without emitting [#1615](https://github.com/ngxs/store/pull/1615)
 - Fix: Router Plugin - Update state after route successfully activates [#1606](https://github.com/ngxs/store/pull/1606)
 - Fix: HMR Plugin - Show error when use Angular Ivy with JIT mode [#1607](https://github.com/ngxs/store/pull/1607)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ $ npm install @ngxs/store@dev
 - Feature: Logger Plugin - Add filter for Logger Plugin [#1571](https://github.com/ngxs/store/pull/1571)
 - Performance: Logger Plugin - Plugin should lazy inject the store once [#1550](https://github.com/ngxs/store/pull/1550)
 - Fix: `ofAction*` methods should prevent passing anything except of `ActionType` [#1616](https://github.com/ngxs/store/pull/1616)
+- Fix: remove implement SelectType<T> due to cannot assign @Select to private/protected property [#1623](https://github.com/ngxs/store/pull/1623)
 - Fix: Actions are not canceled when any `Observable` returned by any handler is completed without emitting [#1615](https://github.com/ngxs/store/pull/1615)
 - Fix: Router Plugin - Update state after route successfully activates [#1606](https://github.com/ngxs/store/pull/1606)
 - Fix: HMR Plugin - Show error when use Angular Ivy with JIT mode [#1607](https://github.com/ngxs/store/pull/1607)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ $ npm install @ngxs/store@dev
 - Feature: Logger Plugin - Add filter for Logger Plugin [#1571](https://github.com/ngxs/store/pull/1571)
 - Performance: Logger Plugin - Plugin should lazy inject the store once [#1550](https://github.com/ngxs/store/pull/1550)
 - Fix: `ofAction*` methods should prevent passing anything except of `ActionType` [#1616](https://github.com/ngxs/store/pull/1616)
-- Fix: remove implement SelectType<T> due to cannot assign @Select to private/protected property [#1623](https://github.com/ngxs/store/pull/1623)
+- Fix: remove SelectType<T> implementation due to cannot assign @Select to private/protected property [#1623](https://github.com/ngxs/store/pull/1623)
 - Fix: Actions are not canceled when any `Observable` returned by any handler is completed without emitting [#1615](https://github.com/ngxs/store/pull/1615)
 - Fix: Router Plugin - Update state after route successfully activates [#1606](https://github.com/ngxs/store/pull/1606)
 - Fix: HMR Plugin - Show error when use Angular Ivy with JIT mode [#1607](https://github.com/ngxs/store/pull/1607)
--- a/packages/store/types/tests/selection.lint.ts
+++ b/packages/store/types/tests/selection.lint.ts
@@ -145,7 +145,7 @@ describe('[TEST]: Action Types', () => {
     }
   });
 
-  it('issues/1532', () => {
+  it('should support protected and private methods (https://github.com/ngxs/store/issues/1532)', () => {
     const TODOS_TOKEN = new StateToken<string[]>('todos');
 
     @State({ name: TODOS_TOKEN })
--- a/packages/store/types/tests/selection.lint.ts
+++ b/packages/store/types/tests/selection.lint.ts
@@ -170,7 +170,7 @@ describe('[TEST]: Action Types', () => {
 
     @Component({ selector: 'app' })
     class AppComponent {
-      // $ExpectType (state: string[]) => string[]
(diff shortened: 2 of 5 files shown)


arturovt:
It fails because of `keyof` operator which tries to yield public property names in that construct that was added in this PR. AFAIK `keyof` doesn't yield private/protected property names.

But Max is more knowledgeable in that typing stuff so decisions are up to you.

splincode:
@arturovt @markwhitfeld This is a big problem)) I have never used private/protected selectors. And we never had unit tests with private fields, so I thought that was normal behavior and therefore we introduced this in v3.6. While I do not really know how to fix this without breaking the behavior of the current version 3.6. What do you think can be fixed with this?

```ts
It fails because of keyof operator which tries to yield public property names 
```

Do you know how to pull out private/protected fields?

https://github.com/ngxs/store/blob/master/packages/store/src/decorators/select/symbols.ts#L48

arturovt:
As TypeScript team says: **no way** :smile:

splincode:
There is an offer to leave the `select` only for public fields)

arturovt:
Btw what is the purpose of that type and can't we leave it as any?

P.S. just curious.

splincode:
```
Btw what is the purpose of that type
```

- Auto inferable type from selector function

Example:

```ts
@State<Model>({
  name: 'my',
  default: {
     path: 'hello'
  }
})
class MyState {
  @Selector()
  public static getState(state) {
    return state.path;
  }
}
```

// MyState.getState() -> auto inferable (state: Model) => string

```ts
class MyComponent {
   @Select(MyState.getState) public path;  // Magic: auto inferable this.path: Observable<string>
}
```

splincode:
```
can't we leave it as any
```

If move to any (v3.5)

```
class MyComponent {
   @Select(MyState.getState) public path;  // this.path: any
}
```

arturovt:
I understand your point. But this seems like it has been a breaking change for users that might have private properties. Am I wrong?

splincode:
```
But this seems like it has been a breaking change
```

Yes, but we have 3.6.0 for more than six months as a version in production)

arturovt:
There is at least 1 guy that faced breaking changes :rofl:

splincode:
image
we can add to the description that this was a critical change

arturovt:
I'm actually not pushing things with this. As I said all decisions are up to you.

splincode:
@markwhitfeld @arturovt I just don’t know how to fix this problem without breaking typing back (any)

arturovt:
> There is at least 1 guy that faced breaking changes

image

Or even more.. At least there can be a quick fix, I mean just remove `private` and it'll be working.

splincode:
```
I mean just remove private and it'll be working.
```

everyone who encountered this in my company did just that

arturovt:
There is also a guy that says that he's having large codebase and it blocks him from upgrading.

Cannot advise anything in that case... I'm not a type expert :(

splincode:
@markwhitfeld what we do? downgrade to any type?

markwhitfeld:
Are we able to provide a fallback that will permit the private fields but then if the field is public then it will infer the type?
This is similar to https://github.com/DefinitelyTyped/DefinitelyTyped/issues/14811

splincode:
```
Are we able to provide a fallback that will permit the private fields but then if the field is public then it will infer the type?
```

@markwhitfeld  tried, I didn’t succeed

arturovt:
@splincode what do you think if we leave it as any? Personally, I think that your feature was great and I clearly understand its purpose. But we have to think about our users... :disappointed: 
Maybe we could add that feature for v4?

markwhitfeld:
I think that we need to firstly address the regression, and if that means that we move back to `any` for now, then so be it!
We could investigate how to make this type safety work again and push it in another PR if we find a way.

markwhitfeld:
PS. I thought that this construct only gave you the ability to detect if the types were mismatched (type safety). As far as I know, with TypeScript, a decorator cannot be used to infer a type of a property... unless this changed in a recent version.

splincode:
@markwhitfeld sorry, you are right

splincode:
@arturovt @markwhitfeld review again please

kuncevic:
Hey guys, how about to consider of using schematic to fix up edge cases (breaking changes) like that automatically?

markwhitfeld:
@kuncevic Nice idea, but I think that it is totally valid for someone to have a private observable. I don't think it would be right to adjust people's code to make it no longer private. Unless you were thinking about another approach?

bbarry:
> 
> 
> @kuncevic Nice idea, but I think that it is totally valid for someone to have a private observable. I don't think it would be right to adjust people's code to make it no longer private. Unless you were thinking about another approach?

We declare all our store observables in the constructor in order to fully apply tsc strict compiler options:

```typescript
class MyComponent {
   private path: Observable<string>;
   constructor(store: Store) {
      this.path = store.select(MyState.getState);
   }
}
```

Perhaps that migration is a migration that could be done without changing the member accessibility.

splincode:
@markwhitfeld updated

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ $ npm install @ngxs/store@dev
 - Feature: Logger Plugin - Add filter for Logger Plugin [#1571](https://github.com/ngxs/store/pull/1571)
 - Performance: Logger Plugin - Plugin should lazy inject the store once [#1550](https://github.com/ngxs/store/pull/1550)
 - Fix: `ofAction*` methods should prevent passing anything except of `ActionType` [#1616](https://github.com/ngxs/store/pull/1616)
+- Fix: remove implement SelectType<T> due to cannot assign @Select to private/protected property [#1623](https://github.com/ngxs/store/pull/1623)
```
arturovt, line 13:
```suggestion
- Fix: remove SelectType<T> implementation due to cannot assign @Select to private/protected property [#1623](https://github.com/ngxs/store/pull/1623)
```

splincode: done

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ $ npm install @ngxs/store@dev
 - Feature: Logger Plugin - Add filter for Logger Plugin [#1571](https://github.com/ngxs/store/pull/1571)
 - Performance: Logger Plugin - Plugin should lazy inject the store once [#1550](https://github.com/ngxs/store/pull/1550)
 - Fix: `ofAction*` methods should prevent passing anything except of `ActionType` [#1616](https://github.com/ngxs/store/pull/1616)
+- Fix: remove SelectType<T> implementation due to cannot assign @Select to private/protected property [#1623](https://github.com/ngxs/store/pull/1623)
```
markwhitfeld, line 13:
```suggestion
- Fix: Remove the recent `@Select` type safety check due to issues with private/protected properties [#1623](https://github.com/ngxs/store/pull/1623)
```

packages/store/types/tests/selection.lint.ts
```diff
@@ -138,4 +144,43 @@ describe('[TEST]: Action Types', () => {
       }
     }
   });
+
+  it('issues/1532', () => {
```
markwhitfeld, line 148:
```suggestion
  it('should support protected and private methods (https://github.com/ngxs/store/issues/1532)', () => {
```

dmrickey:
I guess I've never tried assigning a select statement to a private property, but I'm just now jumping on the 3.7.x bandwagon from 3.6.2 and this is a heartbreaking change. I can see why someone might want to, but I would much rather have the typecheck back than be able to assign to private props this way.

markwhitfeld:
@dmrickey I will be working on a better, more type-safe alternative to `@Select` very soon, so all is not lost here 👍 
This is the basic idea for syntax:
```TS
isLoggedIn$ = select(AuthState.isLoggedIn);
```
no deocrator, less code, more compatible, and type safe
or potentially:
```TS
isLoggedIn$ = new StoreObservable(AuthState.isLoggedIn);
```
What do you think?

dmrickey:
Yeah both of those look like good options (though I personally like the first option better, it seems a bit weird to me to be "selecting it into a constructor"). How will this work with memoized selectors (or are memoized selectors outside of 3.6.2 already typesafe)?

dmrickey:
Are typesafe selectors in 3.7.x now or is that not back in yet?

markwhitfeld:
@dmrickey Unfortunately not, but we have some really great type safety coming to `createSelector` in v3.8, some really great state slicing utilities (#1653) and potentially some other `@Selector` typing improvements (#1765). 

So, no improvements on the `@Select` decorator as yet, but would you like to try out the experimental utility I made?
ie.: `isLoggedIn$ = new StoreObservable(AuthState.isLoggedIn);`
I would love to get some feedback.

dmrickey:
I can take a look (not sure where the experimental repo is, a quick google didn't turn it up). Does it only work with memoized selectors?

markwhitfeld:
@dmrickey Have a look at this stackblitz:
https://stackblitz.com/edit/ngxs-select-dispatch-utils-v0?file=src%2Fapp%2Fapp.component.ts

What do you think of the `select` and `dispatch` utilities?
Note: these have not yet been tested in production, but I don't see any reason why they would fail.

markwhitfeld:
@dmrickey Ping. I would love to hear what you think.

dmrickey:
> @dmrickey Ping. I would love to hear what you think.

Sorry I never got back to you. I've taken a look and my only question is do all selects specifically need a memoized selector? Can we not select from the state with a syntax similar to the 3.6 (`@Select((x: FooState) => x.foo.bar): Observable<Bar>;`)?

dmrickey:
What's the status of this with the current release?

PR #1624, opened 2020-06-20
perf: do not create `new Observable()` every time when subscribing to `ivyEnabledInDevMode$`

arturovt:
I have noticed that I've added `asObservable()` on the replay subject which creates `new Observable` every time the `ivyEnabledInDevMode` is invoked. That shouldn't be like that since it can slow down things on large codebases and dozens of states.


arturovt:
Yep, I haven't added anything.

Issue #1603, opened 2020-05-24
🚀[FEATURE]: Throw an error if an Action's static type field is undefined.


spockoyno:
### Relevant Package

 This feature request is for @ngxs/store.

### Description
 This would prevent bugs due to mistyped action definitions like 
 `export class ChangedInputs{
 static readonly type: '[Inputs] updated user inputs'
  constructor(public value: InputModel) {}
}`
which should have `type =` instead. Neither the IDE nor TypeScript flags this as error. In my case, this  caused state's action methods to catch the wrong actions and update Models with their payloads. I thought it was a bug in NGXS, see #1601, then after a few hours found a solution in  [SO question](https://stackoverflow.com/questions/59363163/ngxs-dispatch-run-all-action-for-state). Conceivably, it may take a while to realise there is a problem in the first place. 

### Describe the solution you'd like
 Seemingly the most straightforward solution is to check an action's type field and throw an error if it's undefined. Surely, a  `static readonly` field should be assigned only once at the class definition. 

### Describe alternatives you've considered
 Open a  Webstorm issue to flag this typo as error (might take years before it's implemented though). Also, consider using  [](https://github.com/ngxs-labs/data) in a future project, so such issues can't appear. It might be also helpful to mention this typo in the docs.

splincode:
You can send PR also)

markwhitfeld:
Good news, this has been addressed in NGXS v3.7 which was released last week.
( see https://dev.to/ngxs/announcing-ngxs-3-7-1kbk for release notes )

I will close this issue. Please let us know if it is not resolved.

PR #1625, opened 2020-06-21
feat: throw an error when actions do not have a static type property

arturovt:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: #1603 

```
[ ] Yes
[x] No
```

### DIFF
--- a/packages/store/src/internal/dispatcher.ts
+++ b/packages/store/src/internal/dispatcher.ts
@@ -33,7 +33,7 @@ export class InternalDispatcher {
   /**
    * Dispatches event(s).
    */
-  dispatch(actionOrActions: any): Observable<any> {
+  dispatch(actionOrActions: any | any[]): Observable<any> {
     const result = this._ngxsExecutionStrategy.enter(() =>
       this.dispatchByEvents(actionOrActions)
     );
@@ -50,7 +50,7 @@ export class InternalDispatcher {
     return result.pipe(leaveNgxs(this._ngxsExecutionStrategy));
   }
 
-  private dispatchByEvents(actionOrActions: any): Observable<any> {
+  private dispatchByEvents(actionOrActions: any | any[]): Observable<any> {
     if (Array.isArray(actionOrActions)) {
       if (actionOrActions.length === 0) return of(this._stateStream.getValue());
       return forkJoin(actionOrActions.map(action => this.dispatchSingle(action)));
--- a/packages/store/src/internal/dispatcher.ts
+++ b/packages/store/src/internal/dispatcher.ts
@@ -61,12 +61,10 @@ export class InternalDispatcher {
 
   private dispatchSingle(action: any): Observable<any> {
     const type: string | undefined = getActionTypeFromInstance(action);
-
     if (!type) {
       const error = new Error(
         `This action doesn't have a type property: ${action.constructor.name}`
       );
-
       return throwError(error);
     }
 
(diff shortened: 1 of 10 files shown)


splincode:
Update bundlesize please

packages/store/src/internal/dispatcher.ts
```diff
@@ -32,7 +33,7 @@ export class InternalDispatcher {
   /**
    * Dispatches event(s).
    */
-  dispatch(actionOrActions: any | any[]): Observable<any> {
```
markwhitfeld, line 35: What is the reason for making this change as part of this PR?

packages/store/tests/issues/issue-1603-no-type-property-on-the-action.spec.ts
```diff
@@ -0,0 +1,83 @@
+import { TestBed } from '@angular/core/testing';
+import { NgxsModule, State, Action, Store, Actions, ofActionDispatched } from '@ngxs/store';
+
+describe('https://github.com/ngxs/store/issues/1603', () => {
```
markwhitfeld, line 4: PS. I would prefer a description of the scenario here (we already have the issue reference number in the file path)

packages/store/src/internal/dispatcher.ts
arturovt: I actually was going through `dispatch` chain of calls and found that argument name is inconsistent everywhere. In some places it's called `event`, in some places it's called `action`, but it's actually `actionOrActions`, since those methods accept `actionInstance || Array<actionInstance>`.
We will never have another reason to change that, because no one will make a PR just to change the argument name :)
My hands just have reached there and from my perspective I thought it's better to keep naming consistent.

packages/store/tests/issues/issue-1603-no-type-property-on-the-action.spec.ts
arturovt: Okay, will upd.

packages/store/src/internal/dispatcher.ts
markwhitfeld: Ok. I'm happy with the parameter name change, but please keep the type to be `any | any[]`.
Even though it is really no different with type checking, it still documents the API well.

arturovt: Hmmm, might be. The reason why I changed that to `any` because any type unit with `any` always default to `any`. I can revert it back to `any | any[]` if you wish.

splincode:
@markwhitfeld review please

packages/store/src/internal/dispatcher.ts
```diff
@@ -59,6 +60,16 @@ export class InternalDispatcher {
...
   private dispatchSingle(action: any): Observable<any> {
+    const type: string | undefined = getActionTypeFromInstance(action);
+
+    if (!type) {
+      const error = new Error(
+        `This action doesn't have a type property: ${action.constructor.name}`
+      );
+
+      return throwError(error);
+    }
```
markwhitfeld, line 71:
```suggestion
    if (!type) {
      const error = new Error(
        `This action doesn't have a type property: ${action.constructor.name}`
      );
      return throwError(error);
    }
```

PR #1628, opened 2020-06-21
fix(devtools-plugin): actions with "action" property end up logged with <UNDEFINED>

joaqcid:
make action:null before sending to devtools

## PR Type

```
[ X ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

When dispatching an action which payload includes an `action` property, ReduxDevTools logs it as <UNDEFINED>

Issue Number: N/A

## What is the new behavior?

`action` property will be set to null before sending to devtools to log action correctly

```
[ ] Yes
[ X ] No
```
Screen Shot 2020-06-21 at 17 43 17

### DIFF
--- a/packages/devtools-plugin/src/devtools.plugin.ts
+++ b/packages/devtools-plugin/src/devtools.plugin.ts
@@ -65,7 +65,7 @@ export class NgxsReduxDevtoolsPlugin implements NgxsPlugin {
     if (isInitAction) {
       this.devtoolsExtension!.init(state);
     } else {
-      this.devtoolsExtension!.send({ ...action, type }, newState);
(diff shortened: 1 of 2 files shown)


splincode:
Can be add test?
https://github.com/ngxs/store/blob/master/packages/devtools-plugin/tests/devtools.spec.ts

joaqcid:
> Can be add test?
> https://github.com/ngxs/store/blob/master/packages/devtools-plugin/tests/devtools.spec.ts

added, although since we devtools is mocked in the test, i could only spy on the call made to the devtools

PR #1633, opened 2020-06-29
feat: upgrade to Angular 10

splincode:
## PR Checklist

Please check if your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://github.com/ngxs/store/blob/master/CONTRIBUTING.md#commit
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type

What kind of change does this PR introduce?

```
[ ] Bugfix
[x] Feature
[ ] Code style update (formatting, local variables)
[ ] Refactoring (no functional changes, no api changes)
[ ] Build related changes
[ ] CI related changes
[ ] Documentation content changes
[ ] Other... Please describe:
```

## What is the current behavior?

Issue Number: https://github.com/ngxs/store/issues/1630

## Does this PR introduce a breaking change?

```
[x] Yes
[ ] No
```

PR #1641, opened 2020-07-05
test(integrations): add ng 10 ivy

joaqcid:
## PR Type

```
[ ] Other... Please describe:
```
Integration test for Angular v10 (Ivy enabled)

```
[ ] No
```


joaqcid:
- I added same basic tests as ng9 (ivy). 
- I tried to include a few more tests, but i got an error.
- included script integration:ng10ivy to compile and run tests from root.

integrations/hello-world-ng10-ivy/.editorconfig
```diff
@@ -0,0 +1,16 @@
+# Editor configuration, see https://editorconfig.org
+root = true
+
+[*]
```
arturovt, line 4: Not needed.

integrations/hello-world-ng10-ivy/.gitignore
```diff
@@ -0,0 +1,47 @@
...
+# dependencies
+/node_modules
+
+# profiling files
+chrome-profiler-events*.json
+speed-measure-plugin*.json
+
+# IDEs and editors
+/.idea
+.project
```
arturovt, line 20: Not needed.

integrations/hello-world-ng10-ivy/README.md
```diff
@@ -0,0 +1 @@
+Integration test ng-10 with Ivy enabled
```
arturovt, line 1: Not needed.

arturovt:
> I tried to include a few more tests, but i got an error.

Umm? :thinking:

joaqcid:
> > I tried to include a few more tests, but i got an error.
> 
> Umm? 🤔

i tried running the default `app.component.spec.ts` that gets generated with ng new but i got the following error

image

and if i add `import 'jest-preset-angular';`

i get the following

image

splincode:
@joaqcid https://github.com/ngxs/store/pull/1633/files#diff-77f86b6e6a488ea46db06421d97eeb77L1

yes, it's not need now

joaqcid:
@splincode then the issue is the one above 

image

markwhitfeld:
@joaqcid Could you add those tests back in (the failing ones) so that we can see the state of this from the build server?

joaqcid:
added

markwhitfeld:
That's strange... it's not failing in CI.
Did you add that project's tests to CI for both CircleCI and Travis?

.circleci/config.yml
```diff
@@ -166,6 +166,14 @@ jobs:
...
           command: yarn integration:ivy
 
+  integration_ng10_ivy_tests:
+    <<: *job_defaults
+    steps:
+      - *attach_workspace
+      - run:
+          name: Run integration tests for ng10 (Ivy) application
+          command: yarn integration:ng10ivy                
+
```
joaqcid, line 176: @markwhitfeld i added this line, is that enough?

markwhitfeld: It needs to be added in some other places too.
Search through the file for `integration_ng` to find them ;-)

markwhitfeld:
@joaqcid @splincode @arturovt Does anyone have an idea on how to resolve this?

splincode:
we need remove `jest-preset-angular` :)

splincode:
@joaqcid please create issue in Angular repo about 

```
  ● AppComponent › should render title

    This constructor is not compatible with Angular Dependency Injection because its dependency at index 0 of the parameter list is invalid.
    This can happen if the dependency type is a primitive like a string or if an ancestor of this class is missing an Angular decorator.

    Please check that 1) the type for the parameter at index 0 is correct and 2) the correct Angular decorators are defined for this class and its ancestors.

```

PR #1662, opened 2020-09-06
feature(form-plugin): ngxsFormClearOnDestroy should allow the attribute with no value

sa-bangash:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

[stackoverflow](https://stackoverflow.com/questions/61628358/how-to-keep-angular-reactive-form/63767759#63767759)
ngxsFormClearOnDestroy="false" behavior as same is ngxsFormClearOnDestroy="true" so 'false' and 'true' treated as string.
just putting this "ngxsFormClearOnDestroy" will not clear the form
## What is the new behavior?
ngxsFormClearOnDestroy="false" will not clear the form on destroy also just putting ngxsFormClearOnDestroy will clear the form so no need to assign the true or false.

```
[ ] Yes
[x] No
```


packages/store/src/utils/utils.ts
```diff
@@ -64,3 +64,16 @@ export const setValue = (obj: any, prop: string, val: any) => {
...
+/**
+ * Boolean conversion:
+ *
+ *    like '', 'false' to boolean
+ *    ngxsFormClearOnDestroy, ngxsFormClearOnDestroy="true" should be treated as true
+ *    ngxsFormClearOnDestroy='false' should be treated as false
+ *
+ * @ignore
+ */
+export function coerceBoolean(val: any) {
```
splincode, line 77: Add unit test with `ngxsFormClearOnDestroy='false'` and `ngxsFormClearOnDestroy='true'`

packages/form-plugin/src/directive.ts
```diff
@@ -20,7 +20,15 @@ export class FormDirective implements OnInit, OnDestroy {
   debounce = 100;
 
   @Input('ngxsFormClearOnDestroy')
-  clearDestroy = false;
+  set clearDestroy(val: boolean) {
```
splincode, line 23: `val: boolean | string`

packages/store/src/utils/utils.ts
arturovt: This also can be moved out of `store` and stored directly in the form plugin's code.

markwhitfeld:
Technically, the correct way to use this property is with the input binding syntax: 
`[ngxsFormClearOnDestroy]="false"` or `[ngxsFormClearOnDestroy]="true"`

That being said, I can see it being useful to make reasonable sense of these "incorrect" binding values.

packages/store/src/utils/utils.ts
markwhitfeld: Agreed, this utility function should not be added to or exposed from the core lib, but rather added as a private function within the form plugin.

docs/concepts/actions.md
```diff
@@ -77,3 +77,55 @@ Examples:
...
+  export class FetchAll {
+    static readonly type = '[Todo] Fetch All'
+  }
+  
+  export class Delete {
+    static readonly type = '[Todo] Delete';
+    constructor(public id: number) { }
+  }
+}
+```
```
markwhitfeld, line 131: How do these changes relate to this PR?

markwhitfeld (changes requested):

packages/store/src/utils/utils.ts
sa-bangash: should I create utils file inside form plugin and also sperate file for utils.spec.ts

packages/form-plugin/src/directive.ts
sa-bangash: @splincode I wrote  the set getter so type issue for getter I think boolean is right here or any?

sa-bangash:
> Technically, the correct way to use this property is with the input binding syntax:
> `[ngxsFormClearOnDestroy]="false"` or `[ngxsFormClearOnDestroy]="true"`
> 
> That being said, I can see it being useful to make reasonable sense of these "incorrect" binding values.

@markwhitfeld yes you are right but I think a good approach will be to just put ngxsFormClearOnDestroy to do work. instead of passing false and true. most of the stander open-source libraries follow this approach.

markwhitfeld:
@sa-bangash So, are you saying that just the presence of this attribute should enable this behavior?
Makes sense in my opinion.
Does your change enable this?

packages/form-plugin/src/directive.ts
markwhitfeld: I think that it could stay as boolean, unless there is a type issue?

packages/store/src/utils/utils.ts
markwhitfeld: Please just put this code inline in the `set clearDestroy(val: boolean) { ...` function.
There is no need to extract a utility with just a single usage.

markwhitfeld:
I am about to cut a release, so please ping me on slack when you have made the requested changes.

packages/form-plugin/tests/form.plugin.spec.ts
```diff
@@ -236,7 +236,173 @@ describe('NgxsFormPlugin', () => {
       expect(input.value).toBe('Buy some coffee');
     });
 
-    it('should update the state if "ngxsFormClearOnDestroy" option is provided', () => {
+    it('should update the state if ngxsFormClearOnDestroy option is provided', () => {
```
arturovt, line 239: It might be better to add a test with listening to `ofActionDispatched(UpdateForm)` since this is the only action that is done in `ngOnDestroy` when `ngxsFormClearOnDestroy` is truthy. And also another test that ensures that this action is not dispatched. IMHO tests with assuring `store.snapshot` is not intuitive.

packages/form-plugin/src/utils.ts
```diff
@@ -0,0 +1,12 @@
...
+ * Boolean conversion:
+ *
+ *    like '', 'false' to boolean
+ *    ngxsFormClearOnDestroy, ngxsFormClearOnDestroy="true" should be treated as true
+ *    ngxsFormClearOnDestroy='false' should be treated as false
+ *
+ * @ignore
+ */
+export function coerceBoolean(val: boolean | string | undefined | null) {
+  return val != null && `${val}` !== 'false';
```
splincode, line 11: maybe try check undefined also? and use strict (`val !== null`)

sa-bangash:
> I am about to cut a release, so please ping me on slack when you have made the requested changes.

ok

packages/form-plugin/tests/form.plugin.spec.ts
markwhitfeld: I think that in this case using the `selectSnapshot` to check the state is the right way to go because the resultant state is the observable behavior of this plugin from a user's perspective.
Testing for an action is potentially testing the internals too much.

sa-bangash: hmmm it will cover more if we watch ofActionDispatched(UpdateForm) good point. let me change

sa-bangash: image
@arturovt is it ok?

docs/concepts/actions.md
sa-bangash: @markwhitfeld I did this one year ago and already merged,  but don't know why this appears.

packages/form-plugin/src/utils.ts
arturovt: `!= null` covers both null and undefined 😃

docs/concepts/actions.md
markwhitfeld: Oh, the process would be that in your repo, you should get the latest from master, then create a branch for the PR and then push changes to this branch. I think it is doing this because you used the same branch for the other PR.
Strange that it is giving conflicts though.

packages/form-plugin/src/directive.ts
```diff
@@ -20,7 +20,15 @@ export class FormDirective implements OnInit, OnDestroy {
...
+    this._clearDestroy = val !== null && `${val}` !== 'false';
```
markwhitfeld, line 24: Potentially could be:
```suggestion
    this._clearDestroy = val && `${val}` !== 'false';
```

sa-bangash: > Potentially could be:

this implementation is really focused on behavior change like just passing **ngxsFormClearOnDestroy** instead of doing this **ngxsFormClearOnDestroy="true"**. in your case if I pass ngxsFormClearOnDestroy it teated as false.

sa-bangash: null=>false (handle this condition)
undefined =>true
'false' => false
'true' => true
'' => true

markwhitfeld: No way! I thought that an attribute with the value omitted in the HTML was given the value of the name of that attribute.
Just tested it out and you are right. You learn something new every day!

markwhitfeld (approved):

packages/form-plugin/src/directive.ts
arturovt: Would we want to leave it as `value != null` except `!== null`? Since it will not work if I pass `undefined` there which is super easy when you have a property declaration in the component but it doesn't have assigned value:
```ts
@Component({
  template: '<form ngxsForm [clearOnDestroy]="clearOnDestroy">...'
})
export class CMP {
  clearOnDestroy?: unknown;
}
```
```ts
undefined !== null && `${undefined}` !== 'false' // => true
```

arturovt (approved):

packages/form-plugin/tests/form.plugin.spec.ts
```diff
@@ -236,7 +236,173 @@ describe('NgxsFormPlugin', () => {
...
+      fixture.destroy();
+      expect(store.selectSnapshot(({ todos }) => todos.todosForm)).toEqual({
+        model: {},
+        dirty: null,
+        status: null,
+        errors: {}
+      });
+    });
+
+    it('should not update the state if ngxsFormClearOnDestroy="false" option is provided', () => {
```
markwhitfeld, line 293: @sa-bangash I think that we should have an additional test just like this that checks that it does not clear the state if this option is left out. It would have picked up the issue regarding the `undefined` value that @arturovt pointed out.
After adding this test please just double-check the test by confirming that it fails if you make the `val !== null` instead of the `val != null` as currently exists.

markwhitfeld (changes requested):
Please add an additional test 👍

markwhitfeld (approved):
Actually I have decided to merge this and add the test myself so that I can get things ready for the release.
Thank you for your great work!!!

markwhitfeld:
PS. @sa-bangash This is the process we recommend for PRs to make things easier:
https://www.ngxs.io/community/contributing#submitting-a-pull-request-pr
It would have avoided the conflicts 👍

sa-bangash:
> https://www.ngxs.io/community/contributing#submitting-a-pull-request-pr

will try in future :)

PR #1727, opened 2021-03-07
perf(storage-plugin): tree-shake console.* calls and expand message

arturovt:


splincode:
@markwhitfeld any news?)

splincode (approved):

markwhitfeld:
@arturovt I am just wondering if this will remove these console calls from our published bundle?
We would not want them to be removed from the published bundle because a developer should still see these messages if they are using Angular in dev mode.

arturovt:
This will not remove console calls from our published bundles and developers will still see warnings in development mode. Console calls will be removed only if the target type is “application” (which means the storage plugin is used in some angular application) and the target mode is production (the app is built with `ng build —prod`).

markwhitfeld (approved):
This is really amazing!!! Nice find!
Can we do this anywhere else?

markwhitfeld:
@arturovt please add an update to the `CHANGELOG.md` in this PR and then go ahead and merge 👍

arturovt:
> This is really amazing!!! Nice find!
> Can we do this anywhere else?

Of course, I'm thinking about tree-shaking all dev only stuff, e.g. Ivy warnings (that Injectable decorator is not attached, etc).

markwhitfeld (approved):

arturovt:
Huh.. no worries I’ll check that

arturovt:
I will do a next tree shaking step and will update the config accordingly.

I love how master branch always fails comparing to PR branch 😄 😄

PR #1731, opened 2021-03-23
fix(form-plugin): UpdateFormDirty overriding form model with previous value

d3v0ps:
Sometimes when updateFormStateWithRawValue gets called UpdateFormDirty gets triggered after UpdateFormValue and for some reason I don't yet understand, it overwrites the model state with the previous one. This change seems to fix that problem.

## PR Checklist

## PR Type

```
[X] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

updateFormStateWithRawValue calls UpdateFormDirty, UpdateFormErrors and UpdateFormStatus in parallel with UpdateFormValue. For some weird reason this causes sometime the UpdateFormDirty action gets processed after UpdateFormValue but without the old state value in the model

Issue Number: N/A

## What is the new behavior?

updateFormStateWithRawValue calls UpdateFormDirty, UpdateFormErrors and UpdateFormStatus only when UpdateFormValue its processed

```
[ ] Yes
[ ] No
```

### DIFF
--- a/packages/form-plugin/src/directive.ts
+++ b/packages/form-plugin/src/directive.ts
@@ -133,10 +133,6 @@ export class FormDirective implements OnInit, OnDestroy {
     const value = this._formGroupDirective.control.getRawValue();
 
     const actions: any[] = [
-      new UpdateFormValue({
-        path: this.path,
-        value
-      }),
       new UpdateFormDirty({
         path: this.path,
         dirty: this._formGroupDirective.dirty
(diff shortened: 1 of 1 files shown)


arturovt:
Hey @d3v0ps 

Can I ask you to log a related issue and reference it from this PR.
It just makes it more searchable. Could you also add tests for that code change. Thanks!

markwhitfeld:
Hi @d3v0ps Thank you for this PR!
Would it be possible to log a related issue to provide more context, and to add some tests with respect to this fix?

d3v0ps:
Hi @arturovt @markwhitfeld sorry for my late response. I found this issue during the development of a big application so I can't provide you an easy way to reproduce it right now, but if I get some free time I will try to reproduce in a small sandboxed environment.

markwhitfeld:
@d3v0ps Thank you!
If you could reproduce it with a test that would be first prize!! ;-)
And then the test would pass in this PR 👍

arturovt:
I'll close this PR for now since it's been a bit old :) I wish we've had an issue that has a clear description of what we need to solve, this would be super helpful.

PR #1740, opened 2021-04-27
fix(store): do not re-use the global `Store` instance between different apps

arturovt:


markwhitfeld:
This is crazy!!!! Where do you come up with these mad scientist ideas?!!!
I am in awe of your awesomeness!

markwhitfeld:
Just a thought... 
You have come up with a way to access the injector from a decorator.
You could potentially make this more general-purpose by saving the `Injector` to the instance.
Which, in the case of a component, will be the `NodeInjector`.
`instance[StoreInstance] = ɵɵdirectiveInject(Store);`
could become:
`instance[InjectorInstance] = ɵɵdirectiveInject(Injector);`
and then the decorator is able to access anything it needs from the local injector.

Following on from this... potentially the API of this utility could be this simple:
in the decorator, a one-liner call `captureLocalInjector(target)` and then to retrieve something from the local injector call `localInject(this, Store)`.

What do you think?

markwhitfeld:
Not sure how you are going to test this, but I'm ready to be impressed ;-)
Ping me when you pull this out of draft!

splincode:
Maybe we also add integration test with module federations?

arturovt:
@markwhitfeld I got your idea, I'll try to find something similar to this.

asaadedd (commented):
Thank you for fixing this issue, also it will be great if we have the same fix to SelectSnapshot decorator from ngxs-labs as well.

In our project we created the same fix ( from an architecture perspective ) with some additional functionality. Sorry that I did not create a PR with it, but between the project and personal life I did not got the time for it.

In our project we found out that only overwriting the 'factory' method is not enough for all the cases, it seems that angular for some providers it created the instance from the 'factory' method inside the definition and other times it creates it from the 'ɵfac' inside the target contructor.

For use for example we have some services which are defined as provides in a lazyLoaded module ( so they are not provided in root ) and for those services the 'factory' method is not called which means that the store is not set to the service.

Sorry again for not creating a PR with our solution.

packages/store/src/decorators/select/select.ts
```diff
@@ -18,10 +31,84 @@ export function Select<T>(rawSelector?: T, ...paths: string[]): PropertyDecorato
...
+  // Let's try to get any definition.
+  const def = ɵprov || ɵɵpipe || ɵcmp || ɵdir;
+
+  // This means that `@Select()` decorator is used within some non-Angular class,
+  // e.g. some custom class which is not decorated with `@Injectable()` or any other Angular decorator.
+  if (!def) {
+    return;
+  }
+
+  def.factory = () => {
```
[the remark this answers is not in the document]
asaadedd: [ngxs-store.txt](https://github.com/ngxs/store/files/6391669/ngxs-store.txt)
You will find in the attached txt the compiled version of ngxs/store with our changes, if you will take a look you will better understand how we overwritten the 'ɵfac' method.

arturovt:
@markwhitfeld well, I'm able to get the node injector inside the getter but only for directives and components through:
```js
Object.defineProperty(target, key, {
  get() {
    const store = ɵgetLContext(this).lView[9].get(Store);
    return this[selectorId];
  }
});
```
That won't work in services and pipes since the `LView` exists only for directives and components.

I'm open to ideas tho.

packages/store/src/decorators/select/select.ts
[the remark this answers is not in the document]
markwhitfeld: Ok, so your modifications are between line 3880 and line 3979.
@arturovt have a look at this to see how it compares to your approach.

markwhitfeld:
@arturovt Yeah, I know that that approach will not work. I was referring to using your current approach, but I was saying that you could ask the injector for the `Injector` instead of the `Store`. Then you have access to the local injector and you can inject anything you need (in this case `Store`). I was also suggesting a refactoring into 2 simple methods that get called from the `Select` decorator to make it a bit cleaner.
PS. One thing to check with this is, if you keep a reference to the Injector, will it be disposed from Memory ok when the component is destroyed.

packages/store/src/decorators/select/select.ts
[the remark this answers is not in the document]
markwhitfeld: @arturovt for ease of reference, here is the extract of modified code:
```TS
function getFactory(type) {
  return type.ɵfac;
}

function getDef(type) {
  if (type[ɵNG_COMP_DEF] && type[ɵNG_COMP_DEF].type === type) {
    return type[ɵNG_COMP_DEF];
  } else if (type[ɵNG_DIR_DEF] && type[ɵNG_DIR_DEF].type === type) {
    return type[ɵNG_DIR_DEF];
  } else if (type[ɵNG_PIPE_DEF] && type[ɵNG_PIPE_DEF].type === type) {
    return type[ɵNG_PIPE_DEF];
  } else if (type[ɵNG_INJ_DEF] && type[ɵNG_INJ_DEF].type === type) {
    return type[ɵNG_INJ_DEF];
  } else if (ɵngcc0.ɵgetInjectableDef(type)) {
    return ɵngcc0.ɵgetInjectableDef(type);
  } else {
    return false;
  }
}

function overwriteFactory(target, storeKey) {
  const def = getDef(target.constructor);
  const factory = getFactory(target.constructor);
  const newFactory = (t) = {
    const store = ɵngcc0.ɵɵdirectiveInject(Store)  ɵngcc0.ɵɵinject(Store);

    Object.defineProperties(target, {
      [storeKey] {
        enumerable false,
        configurable true,
        get () = {
          return store;
        }
      },
    });
    return factory(t);
  };
  def.factory = newFactory;
  Object.defineProperty(target.constructor, 'ɵfac', {
    get () = newFactory,
    configurable true
  });
}


  @fileoverview added by tsickle
  @suppress {checkTypes,extraRequire,missingOverride,missingReturn,unusedPrivateMembers,uselessCode} checked by tsc
 

  Decorator for selecting a slice of state from the store.
  @template T
  @param {=} rawSelector
  @param {...} paths
  @return {}
 
function Select(rawSelector, ...paths) {
    return (
      @param {} target
      @param {} key
      @return {}
     
    function (target, key) {
      const name = key.toString();
      const selectorId = `__${name}__selector`;
      const storeKey = '__store__key';
      const selector = createSelectorFn(name, rawSelector);

      Object.defineProperties(target, {
        [selectorId] {
          writable true,
          enumerable false,
          configurable true
        },
        [name] {
          enumerable true,
          configurable true,
          get() {
            if (this[selectorId]) {
              return this[selectorId];
            }
            if (!this[storeKey]) {
              throw new Error(CONFIG_MESSAGES[VALIDATION_CODE.SELECT_FACTORY_NOT_CONNECTED]());
            }
            this[selectorId] = this[storeKey].select(selector);

            return this[selectorId];
          }
        }
      });

      if (!!getDef(target.constructor)) {
        overwriteFactory(target, storeKey);
      } else {
        Promise.resolve().then(() = overwriteFactory(target, storeKey));
      }
    });
}
```

arturovt:
@markwhitfeld could you elaborate more on the `captureLocalInjector(target)` idea? I'm a bit lost with this one :smile:

markwhitfeld:
Ok, for a lack of a better way to explain it, I'll just hack the code and paste it in this comment...
😆

`/select.ts` 
```TS
import { Store } from '../../store';
import {
  createSelectObservable,
  createSelectorFn,
  PropertyType,
} from './symbols';

import { ensureLocalInjectorCaptured, localInject } from './decorator-injector-util';

/**
 * Decorator for selecting a slice of state from the store.
 */
export function Select<T>(
  rawSelector?: T,
  ...paths: string[]
): PropertyDecorator {
  return function (target, key): void {
    const name: string = key.toString();
    const selectorId = `__${name}__selector`;
    const selector = createSelectorFn(name, rawSelector, paths);

    Object.defineProperties(target, {
      [selectorId]: {
        writable: true,
        enumerable: false,
        configurable: true,
      },
      [name]: {
        enumerable: true,
        configurable: true,
        get(this: PrivateInstance): PropertyType<T> {
          return (
            this[selectorId] ||
            (this[selectorId] = createSelectObservable(
                selector,
                localInject(this, Store)
              
            ))
          );
        },
      },
    });
    ensureLocalInjectorCaptured(target);
  };
}
```

`/decorator-injector-util.ts` 
```TS
import {
  ɵɵdirectiveInject,
  ɵivyEnabled,
  Type,
  AbstractType,
  InjectionToken,
} from '@angular/core';

/**
 * A `Symbol` which is used to save the `InjectorInstance` onto the class instance.
 */
const InjectorInstance: unique symbol = Symbol('InjectorInstance');

/**
 * A `Symbol` which is used to determine if factory has been decorated previously or not.
 */
const FactoryHasBeenDecorated: unique symbol = Symbol(
  'FactoryHasBeenDecorated'
);

export function localInject<T>(
  target: any,
  token: Type<T> | AbstractType<T> | InjectionToken<T>,
  notFoundValue?: T
) {
  const injector = target && target[InjectorInstance];
  return (injector && injector.get(token, notFoundValue)) || notFoundValue;
}

export function ensureLocalInjectorCaptured(target: any) {
  if (!ɵivyEnabled || FactoryHasBeenDecorated in target.constructor.prototype) {
    return;
  }

  target.constructor.prototype[FactoryHasBeenDecorated] = true;

  const constructor: ConstructorWithDefinitionAndFactory = target.constructor;
  // Means we're in AOT mode.
  if (typeof constructor.ɵfac === 'function') {
    decorateFactory(constructor);
  } else {
    // We're running in JIT mode and that means we're not able to get the compiled definition
    // on the class inside the property decorator during the current message loop tick. We have
    // to wait for the next message loop tick. Note that this is safe since this Promise will be
    // resolved even before the `APP_INITIALIZER` is resolved.
    Promise.resolve().then(() => {
      decorateFactory(constructor);
    });
  }
}

function decorateFactory({
  ɵprov,
  ɵɵpipe,
  ɵcmp,
  ɵdir,
  ɵfac,
}: ConstructorWithDefinitionAndFactory): void {
  // Let's try to get any definition.
  const def = ɵprov || ɵɵpipe || ɵcmp || ɵdir;

  // This means that `@Select()` decorator is used within some non-Angular class,
  // e.g. some custom class which is not decorated with `@Injectable()` or any other Angular decorator.
  if (!def) {
    return;
  }

  def.factory = () => {
    const instance = ɵfac!();
    // Note: `inject()` won't work here.
    // We can use the `directiveInject` only during the component
    // construction, since Angular captures the currently active injector.
    // We're not able to use this function inside the getter (when the `selectorId` property is
    // requested for the first time), since the currently active injector will be null.
    instance[InjectorInstance] = ɵɵdirectiveInject(Store);
    return instance;
  };
}

type Factory = () => PrivateInstance;

interface Definition {
  factory: Factory | null;
}

interface ConstructorWithDefinitionAndFactory extends Function {
  // Provider definition for the `@Injectable()` class.
  ɵprov?: Definition;
  // Pipe definition for the `@Pipe()` class.
  ɵɵpipe?: Definition;
  // Component definition for the `@Component()` class.
  ɵcmp?: Definition;
  // Directive definition for the `@Directive()` class.
  ɵdir?: Definition;
  ɵfac?: Factory;
}

interface PrivateInstance {
  [InjectorInstance]?: Store;
  [selectorId: string]: PropertyType<unknown>;
}
```

arturovt:
Ah got it, just move out to the function! 😄

markwhitfeld:
Yes, and attach the injector to the target as opposed to the store. Then you can get anything you want!
Even the `changeDetectorRef`... I'm thinking could be useful for the SelectSnapshot or similar

arturovt:
Now I get why you'd want to store the injector there :smile:

packages/store/src/decorators/select/select.ts
```diff
@@ -18,10 +28,127 @@ export function Select<T>(rawSelector?: T, ...paths: string[]): PropertyDecorato
...
+  ɵcmp?: Definition;
+  // Directive definition for the `@Directive()` class.
+  ɵdir?: Definition;
+  [NG_FACTORY_DEF]?: Factory;
+}
+
+interface PrivateInstance {
+  [InjectorInstance]?: Injector;
+  [selectorId: string]: PropertyType<unknown>;
+}
```
markwhitfeld, line 154: Extract these lines into a `decorator-injector-adapter.ts` file available under the `internals` subpackage and exported so that plugins could make use of this.

markwhitfeld (changes requested):
This is looking really good! Especially that it gets tree-shaken for view-engine.

markwhitfeld:
Note @arturovt We will need to check that everything is disposed correctly when the component is disposed and that the reference from the component instance to the injector does not create a memory leak. 🤔

arturovt:
> Note @arturovt We will need to check that everything is disposed correctly when the component is disposed and that the reference from the component instance to the injector does not create a memory leak.

image

markwhitfeld:
Ok, so if a component that uses this is created and destroyed 50 times (using a ngIf for example) then the memory stays at the same level as before (with the angular app still loaded)?

arturovt:
Umm, what do you mean "the memory stays at the same level as befor"?

markwhitfeld:
As in, there is no memory leak after disposing and recreating the component 50 times

integrations/hello-world-ng11-ivy/cypress/integration-ssr/issue-1646-wrong-ssr-state.spec.ts
```diff
@@ -0,0 +1,16 @@
+/// <reference types="cypress" />
+
+describe('Select decorator returning state from the wrong store during SSR (https://github.com/ngxs/store/issues/1646)', () => {
```
splincode, line 3: can this code be added as another PR?

splincode: or it's imposible?

markwhitfeld: This is one of the issues solved by this fix. I think that this is appropriate to be here.

markwhitfeld:
> As in, there is no memory leak after disposing and recreating the component 50 times

@arturovt have you checked for this?

arturovt:
> > As in, there is no memory leak after disposing and recreating the component 50 times
> 
> @arturovt have you checked for this?

Yeah, of course, there's no leak:

image

markwhitfeld (commented):
Awesome work! Just waiting to hear from you on these:

- If you have tested for a memory leak with a component being created and destroyed 50 times (toggled through ngIf).
- If you have reviewed @asaadedd's code to see if there are any edge cases you could have missed.

arturovt:
@markwhitfeld since you pasted the code from the `.txt` file you can see that changes are mostly identical (except function names, etc.).

markwhitfeld:
@arturovt Great! Could you add the changes to the `CHANGELOG.md`, then we are good to go!

markwhitfeld (approved):

markwhitfeld:
PS. @arturovt could you look at all plugins/decorators that use `SelectFactory` and implement this solution. Then we could potentially tree-shake `SelectFactory` when Ivy is enabled.

markwhitfeld:
@arturovt Have you seen my messages on slack about an issue that has arisen from this?

arturovt:
@markwhitfeld no, I haven't.

markwhitfeld:
> I am getting this failure when running tests on master. I think that it is to do with the @Select store injection fix.
image

markwhitfeld:
> Unfortunately jest is not too happy if you declare a component in the describe part of the test.
This scheduled microtask is the issue. This would most definitely break user project tests.
Do you have any ideas on how we could resolve this?
image

arturovt:
@markwhitfeld got it, thanks for mentioning that issue, haven't encountered it yet before. I'll have a look. I see that `zone.js` throws an error when the task is a microtask (promise) or a macrotask (timer), but doesn't throw if the task is an event task. I'll think if we can schedule the `decorateFactory` invocation using some event queue (not the microtask queue).

arturovt:
Could you help me to reproduce that issue by providing your code that you're running?

markwhitfeld:
I am able to reproduce it by doing the following steps:
- clone ngxs repo
- run `yarn`
- run `yarn build`
- run `npx ngcc`
- run `yarn test --project ngxs`
It breaks on the first test that uses `@Select`

It seems that it has got something to do with running in ivy mode.
I encountered it when I ran the integration tests, and then ran the ngxs tests.
The integration tests run the ngcc compiler against the root node_modules and it has the same affect as above.

arturovt:
Got it, will try it out.

PR #1741, opened 2021-05-02
perf(store): tree-shake `ConfigValidator`, `HostEnvironment` and `isAngularInTestMode`

arturovt:


packages/store/src/internal/config-validator.ts
```diff
@@ -1,33 +1,43 @@
...
+      // The `getIncorrectDevelopmentMessage()` will be tree-shaken away.
+      if (ngDevMode && !this._config.developmentMode) {
+        console.warn(getIncorrectDevelopmentMessage());
+      } else if (!ngDevMode && this._config.developmentMode) {
+        console.warn(getIncorrectProductionMessage());
+      }
+    } else {
+      const ngDevMode = isDevMode();
+      // If `ngDevMode` is undefined then it means we're using the View Engine compiler.
+      if (ngDevMode && !this._config.developmentMode) {
```
splincode, line 36: do you think this parameter is now meaningless?

because now 

`config.developmentMode === ngDevMode`

splincode: I mean, can now not force users to specify this parameter?  but just check for ngDevMode in the code and then you can pass this parameter deprecated

arturovt: I was also thinking if we still need this parameter or not. Basically it's used only to checker whether we have to freeze the state through `deepFreeze` or not. I'm wondering if we can remove the `ConfigValidator` completely and move all checks to `state-operations.ts`.

splincode: @arturovt I agree with you on everything this

arturovt:
@markwhitfeld @splincode so I propose the following:

1. Let's remove `ConfigValidator` and `HostValidator`. The `HostValidator` basically injects tokens (and nothing more). The `ConfigValidator` doesn't validate the config itself but rather verified development mode.
2. Let's move all checks to `state-operations`
3. Deprecate the `developmentMode` property as proposed by @splincode 

If the user is running Ivy:
* then `deepFreeze` the state only when `ngDevMode` is truthy
* we'll be able to tree-shake the `deepFreeze` in production mode

If the user is running View Engine:
* use the `isDevMode()` function from `@angular/core`

arturovt:
@markwhitfeld friendly ping.

markwhitfeld (approved):

markwhitfeld:
> @markwhitfeld @splincode so I propose the following:
> 
> 1. Let's remove `ConfigValidator` and `HostValidator`. The `HostValidator` basically injects tokens (and nothing more). The `ConfigValidator` doesn't validate the config itself but rather verifies development mode.
> 2. Let's move all checks to `state-operations`
> 3. Deprecate the `developmentMode` property as proposed by @splincode
> 
> If the user is running Ivy:
> 
> * then `deepFreeze` the state only when `ngDevMode` is truthy
> * we'll be able to tree-shake the `deepFreeze` in production mode
> 
> If the user is running View Engine:
> 
> * use the `isDevMode()` function from `@angular/core`

I'm happy with proposals 1 & 2.
Regarding proposal 3, we could add this to the list for v4. But I think we need to investigate if there are any valid use cases where the developmentMode and the ngDevMode would not align.

markwhitfeld:
@arturovt could you rebase this PR?

PR #1744, opened 2021-05-03
perf(store): tree-shake `SelectFactory`

arturovt:


arturovt:
@markwhitfeld friendly ping.

markwhitfeld:
This is on hold until we sort out the issues from PR #1740 
See https://github.com/ngxs/store/pull/1740#issuecomment-845728824

arturovt:
@markwhitfeld merged with master, would be happy to hear the feedback.

packages/store/src/decorators/select/select.ts
```diff
@@ -23,9 +24,15 @@ export function Select<T>(rawSelector?: T, ...paths: string[]): PropertyDecorato
         enumerable: true,
         configurable: true,
         get(): PropertyType<T> {
+          // The `localInject` will be tree-shaken away in apps that
+          // still use the View Engine.
+          const store = ɵivyEnabled ? localInject(this, Store) : null;
```
splincode, line 29: what do you think about put to `store/internals/utils.ts`

```ts
export function injectNgxsStore(): Store | null {
  return ɵivyEnabled ? localInject(this, Store) : null;
}

export function injectNgxsConfig(): NgxsConfig | null {
  return ɵivyEnabled ? localInject(this, NgxsConfig) : null;
}
```

and as 

```ts
// ...
get(): PropertyType<T> { 
 
   if (isNil(this[selectorId])) {
      selector = selector || createSelectorFn(injectNgxsConfig(), name, rawSelector, paths);
      this[selectorId] = createSelectObservable(selector, injectNgxsStore());
   }

  return this[selectorId];
}
```

packages/store/src/decorators/select/symbols.ts
```diff
@@ -46,8 +50,36 @@ export function removeDollarAtTheEnd(name: string): string {
...
   ? Observable<ExtractTokenType<T>>
   : T extends (...args: any[]) => any
   ? Observable<ReturnType<T>>
   : any;
+
+function createSelectObservableIvy<T = any>(
+  selector: any,
+  store: Store | null
+): Observable<T> {
+  if (ngDevMode && !store) throwSelectFactoryNotConnectedError();
```
splincode, line 69: more readable


```ts
 if (ngDevMode && !store) { 
    throwSelectFactoryNotConnectedError();
 }
```

packages/store/src/decorators/select/symbols.ts
```diff
@@ -46,8 +50,36 @@ export function removeDollarAtTheEnd(name: string): string {
...
+  return store!.select(selector);
+}
+
+function createSelectObservableViewEngine<T = any>(selector: any): Observable<T> {
+  if (!SelectFactory.store) throwSelectFactoryNotConnectedError();
```
splincode, line 74: more readable


```ts
 if (!SelectFactory.store) { 
   throwSelectFactoryNotConnectedError();
 }
```

packages/store/src/decorators/select/symbols.ts
```diff
@@ -46,8 +50,36 @@ export function removeDollarAtTheEnd(name: string): string {
...
+  return SelectFactory.store!.select(selector);
+}
+
+function createSelectorFnIvy(propsArray: string[], config: NgxsConfig | null) {
+  if (ngDevMode && !config) throwSelectFactoryNotConnectedError();
```
splincode, line 79: more readable

```ts
if (ngDevMode && !config) {
   throwSelectFactoryNotConnectedError();
}
```

splincode:
LGTM, but some questions)

packages/store/src/decorators/select/select.ts
arturovt: This seems to be an over-engineering.

splincode: up to you

splincode:
@markwhitfeld please review))

packages/store/src/decorators/select/symbols.ts
markwhitfeld: I agree that this is more readable, but I don't mind either way.
I think that the only time that you could potentially leave out the block is when it is a guard clause or assertion (like this). But, maybe it is just better to always do the block for consistency.
I think prettier actually has a rule to force this if we choose to make a general rule.

packages/store/src/decorators/select/select.ts
```diff
@@ -11,7 +12,7 @@ export function Select<T>(rawSelector?: T, ...paths: string[]): PropertyDecorato
   return function(target, key): void {
     const name: string = key.toString();
     const selectorId = `__${name}__selector`;
-    const selector = createSelectorFn(name, rawSelector, paths);
+    let selector: SelectorFn | null = null;
```
markwhitfeld, line 15: Could there be any unintended consequence of this only being don in the `get`?
I can't think of any, but it is worth a thought.

markwhitfeld (commented):
Some questions

integrations/hello-world-ng11-ivy/bundlesize.config.json
```diff
@@ -3,7 +3,7 @@
     {
       "path": "./dist-integration/main.*.js",
       "target": "es2015",
-      "maxSize": "250.12 kB",
+      "maxSize": "250.58 kB",
```
arturovt, line 6: Caretaker note: it grew up because we merged fixes related to router, websocket and devtools plugins.

markwhitfeld (approved):

splincode (approved):
Great job

markwhitfeld:
Yes, really great job @arturovt!

PR #1753, opened 2021-05-22
fix(store): do not run `Promise.then` within synchronous tests when decorating factory

arturovt:


setupJest.ts
```diff
@@ -48,6 +49,6 @@ if (CI) {
   });
 }
 
-// Most of our tests are not aware of the `ngDevMode` Ivy and still rely on the old
-// behavior (as it was in the View Engine).
-((global as unknown) as { ngDevMode: undefined }).ngDevMode = undefined;
+if (typeof ɵglobal.ngDevMode === 'undefined') {
```
splincode, line 52: why do we have to force this value?

arturovt: Our tests are using VE still in the root level, VE doesn’t provide this value.

markwhitfeld (approved):
Awesome work!

splincode (approved):

PR #1795, opened 2021-11-03
Feat upsert item

joaqcid:
## PR Checklist

## PR Type

```
[ ] Bugfix
[X] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: N/A

## What is the new behavior?

this PR includes a new operator`upsertItem`. The operator allows to update or insert a new item in an array depending whether item exists or not in the array

```
[ ] Yes
[X] No
```


splincode (approved):

markwhitfeld (commented):
This is a good start.
We will need more tests, but more specifically we need to ensure that we clarify this API (this code was written by me late at night, in the browser, with no discussions, and I don't think the API is clear from the signature without reading the code)

We really need to think about scenarios which are best handled by this operator and which would be better handled by another operator instead of overloading this one too much.

Some scenarios in this problem space that we can think about how we will handle:
- Item exists
  - Replace item if it matches the provided predicate/index
  - Patch item if it matches the provided predicate/index
  - Apply a state operator to item if it matches the provided predicate/index
- Item doesn't exist
  - Prepend item if no items match the provided predicate
  - Append item if no items match the provided predicate
  - Insert item at a specified index if no items match the provided predicate
  - Insert item at a position determined by a function if no items match the provided predicate (think adding item to a sorted list - IMO this should not be part of this operator, but worth mentioning to be thorough)
  - Insert item at provided index which is one past the last index of the array
  - Insert item at provided index which is a more than one past the last index of the array

markwhitfeld:
I think that this merits a discussion.
I started one here: https://github.com/ngxs/store/discussions/1796

arturovt:
I think this has been stalled, will close for now.

PR #1804, opened 2021-11-22
fix(store): allow `@Select()` decorated properties to be accessed within constructors

arturovt:
Hey, this is the fastest solution I came with. I don't have other ideas. I'm opening a PR so we can discuss the implementation and I'm open to other ideas!

### DIFF
--- a/packages/store/src/decorators/select/select.ts
+++ b/packages/store/src/decorators/select/select.ts
@@ -1,9 +1,15 @@
 import { ɵivyEnabled } from '@angular/core';
-import { ensureLocalInjectorCaptured, localInject } from '@ngxs/store/internals';
+import { ReplaySubject } from 'rxjs';
+import { mergeMap } from 'rxjs/operators';
+import {
+  localInject,
+  ensureLocalInjectorCaptured,
+  ensureInjectorNotifierIsCaptured
+} from '@ngxs/store/internals';
 
 import { Store } from '../../store';
 import { NgxsConfig } from '../../symbols';
-import { createSelectObservable, createSelectorFn, PropertyType, SelectorFn } from './symbols';
+import { createSelectObservable, createSelectorFn, SelectorFn } from './symbols';
 
 /**
  * Decorator for selecting a slice of state from the store.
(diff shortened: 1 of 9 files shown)


splincode (approved):

packages/store/src/decorators/select/select.ts
```diff
@@ -23,17 +34,25 @@ export function Select<T>(rawSelector?: T, ...paths: string[]): PropertyDecorato
...
+              mergeMap(() => {
+                const store = localInject(this, Store);
+                const config = localInject(this, NgxsConfig);
+                selector = selector || createSelectorFn(config, name, rawSelector, paths);
+                return createSelectObservable(selector, store);
+              })
+            ));
+          } else {
+            selector = selector || createSelectorFn(null, name, rawSelector, paths);
+            return (this[selectorId] = createSelectObservable(selector, null));
```
splincode, line 54:
```ts
this[selectorId] = createSelectObservable(selector, null);
return this[selectorId];
```

packages/store/src/decorators/select/select.ts
```diff
@@ -23,17 +34,25 @@ export function Select<T>(rawSelector?: T, ...paths: string[]): PropertyDecorato
...
+          }
```
splincode, line 55: what do you think?)

```ts
if (!this[selectorId]) {

  if (ɵivyEnabled) {
    this[selectorId] = injectorNotifier$!.pipe(
      mergeMap(() => {
        const store = localInject(this, Store);
        const config = localInject(this, NgxsConfig);
        selector = selector || createSelectorFn(config, name, rawSelector, paths);
        return createSelectObservable(selector, store);
      })
    ));
  } else {
     selector = selector || createSelectorFn(null, name, rawSelector, paths);
     this[selectorId] = createSelectObservable(selector, null)
  }

}

return this[selectorId];
```

arturovt: Yeah, I don't mind changing that. That looks better.

I'll wait for Mark's response and I'll then update the code with your suggestion.

splincode:
@markwhitfeld friendly ping))

markwhitfeld (approved):

splincode:
@markwhitfeld can we release v3.8?

markwhitfeld:
Wait!!! Why did you merge this with failing tests!

markwhitfeld:
Ok, I'll have to fix it on master. There are a number of things to come together for v3.8

splincode:
@markwhitfeld oh, sorry for that, i thought these tests should be ignored

PR #1806, opened 2021-11-26
fix(store): do not infer T from argument to patch

david-shortman:
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: N/A

```
[ ] Yes
[ ] No
```

### DIFF
--- a/packages/store/operators/src/compose.ts
+++ b/packages/store/operators/src/compose.ts
@@ -1,6 +1,7 @@
 import { StateOperator } from '@ngxs/store';
+import { NoInfer } from './utils';
 
-export function compose<T>(...operators: StateOperator<T>[]): StateOperator<T> {
+export function compose<T>(...operators: NoInfer<StateOperator<T>[]>): StateOperator<T> {
   return function composeOperator(existing: Readonly<T>): T {
     return operators.reduce((accumulator, operator) => operator(accumulator), existing);
   };
--- a/packages/store/operators/src/patch.ts
+++ b/packages/store/operators/src/patch.ts
@@ -8,7 +8,7 @@ type PatchValues<T> = {
 };
 
 export function patch<T>(patchObject: NoInfer<PatchSpec<T>>): StateOperator<NonNullable<T>> {
-  return function patchStateOperator(existing: Readonly<PatchValues<T>>): NonNullable<T> {
+  return (function patchStateOperator(existing: Readonly<PatchValues<T>>): NonNullable<T> {
     let clone = null;
     for (const k in patchObject) {
       const newValue = patchObject[k];
@@ -24,5 +24,5 @@ export function patch<T>(patchObject: NoInfer<PatchSpec<T>>): StateOperator<NonN
       }
     }
     return clone || existing;
-  };
+  } as unknown) as StateOperator<NonNullable<T>>;
 }
--- a/packages/store/operators/src/patch.ts
+++ b/packages/store/operators/src/patch.ts
@@ -1,5 +1,5 @@
 import { StateOperator } from '@ngxs/store';
-import { isStateOperator } from './utils';
+import { isStateOperator, NoInfer } from './utils';
 
 export type PatchSpec<T> = { [P in keyof T]?: T[P] | StateOperator<NonNullable<T[P]>> };
 
@@ -7,14 +7,12 @@ type PatchValues<T> = {
   readonly [P in keyof T]?: T[P] extends (...args: any[]) => infer R ? R : T[P];
 };
 
-type PatchOperator<T> = <U extends PatchValues<T>>(existing: Readonly<U>) => U;
-
-export function patch<T>(patchObject: PatchSpec<T>): PatchOperator<T> {
-  return function patchStateOperator<U extends PatchValues<T>>(existing: Readonly<U>): U {
+export function patch<T>(patchObject: NoInfer<PatchSpec<T>>): StateOperator<NonNullable<T>> {
+  return function patchStateOperator(existing: Readonly<PatchValues<T>>): NonNullable<T> {
     let clone = null;
     for (const k in patchObject) {
       const newValue = patchObject[k];
-      const existingPropValue = existing[k];
+      const existingPropValue = existing[k as Extract<keyof T, string>];
       const newPropValue = isStateOperator(newValue)
         ? newValue(<any>existingPropValue)
         : newValue;
--- a/packages/store/operators/src/update-item.ts
+++ b/packages/store/operators/src/update-item.ts
@@ -1,6 +1,13 @@
 import { StateOperator } from '@ngxs/store';
 
-import { isStateOperator, isPredicate, isNumber, invalidIndex, RepairType } from './utils';
+import {
+  isStateOperator,
+  isPredicate,
+  isNumber,
+  invalidIndex,
+  RepairType,
+  NoInfer
+} from './utils';
 import { Predicate } from './internals';
 
 /**
@@ -10,14 +17,14 @@ import { Predicate } from './internals';
  * function that can be applied to an existing value
  */
 export function updateItem<T>(
-  selector: number | Predicate<T>,
-  operatorOrValue: T | StateOperator<T>
+  selector: number | NoInfer<Predicate<T>>,
+  operatorOrValue: NoInfer<T> | NoInfer<StateOperator<T>>
 ): StateOperator<RepairType<T>[]> {
   return function updateItemOperator(existing: Readonly<RepairType<T>[]>): RepairType<T>[] {
     let index = -1;
 
     if (isPredicate(selector)) {
-      index = existing.findIndex(selector);
+      index = existing.findIndex(selector as Predicate<T>);
     } else if (isNumber(selector)) {
       index = selector;
     }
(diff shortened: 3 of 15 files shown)


david-shortman:
@markwhitfeld Some questions in order to author a full PR-

1. What is the scope of what should be fixed? All operators in the `operators` directory?
2. Should a newer type testing library like ts-snippet be used to assert the functionality of the updates?
3. Should `NoInfer` be exposed for consumers? Or should this PR publish another mechanism (like a function, if possible) for devs to create their own operators?

packages/store/operators/src/patch.ts
```diff
@@ -9,11 +10,12 @@ type PatchValues<T> = {
...
 
-export function patch<T>(patchObject: PatchSpec<T>): PatchOperator<T> {
+export function patch<T>(patchObject: PatchSpec<NoInfer<T>>): PatchOperator<T> {
   return function patchStateOperator<U extends PatchValues<T>>(existing: Readonly<U>): U {
     let clone = null;
-    for (const k in patchObject) {
-      const newValue = patchObject[k];
+    const thePatchObject = <PatchSpec<T>>patchObject;
+    for (const k in thePatchObject) {
+      const newValue = thePatchObject[k];
```
david-shortman, line 18: although `T` should not be inferred from `patchObject`, we still expect that the actual shape of `patchObject` will be `PatchSpec<T>`

markwhitfeld:
@david-shortman Thank you for this PR! Looking good. Please ping me when you take it out of draft.
I was playing with this trick in a TypeScript playground and it almost brought a tear to my eye. The experience of using state operators is incredible again! Even better than when I first created them. I am so ecstatic!!!

Answers to your questions:

1. I think it would be great to fix the operators that may suffer from the parameter interfering with the inference... which I suspect will be most, if not all, of them. You can do them all in this PR, or you can split them up... up to you!
2. I think we stick to DTSLint in this PR, and then introduce a different approach in another PR, so that we can demonstrate the tradeoffs and potentially the scenarios that are better covered with another tool. PS. I have no problem with using both tools for different reasons if there is merit in that.
3. I think that it makes sense for `NoInfer` to be exposed publically for consumers implementing their own operators. Potentially we could include some guidance docs in a later PR for creating your own custom operators.

PS. I see that that `NoInfer` can be applied in different ways with the same effect:
`NoInfer<PatchSpec<T>>` or `PatchSpec<NoInfer<T>>`
I have no real opinion on which is better or more obvious in terms of what it does. What do you think?

david-shortman:
> @david-shortman Thank you for this PR! Looking good. Please ping me when you take it out of draft. I was playing with this trick in a TypeScript playground and it almost brought a tear to my eye. The experience of using state operators is incredible again! Even better than when I first created them. I am so ecstatic!!!
> 
> Answers to your questions:
> 
> 1. I think it would be great to fix the operators that may suffer from the parameter interfering with the inference... which I suspect will be most, if not all, of them. You can do them all in this PR, or you can split them up... up to you!
> 2. I think we stick to DTSLint in this PR, and then introduce a different approach in another PR, so that we can demonstrate the tradeoffs and potentially the scenarios that are better covered with another tool. PS. I have no problem with using both tools for different reasons if there is merit in that.
> 3. I think that it makes sense for `NoInfer` to be exposed publically for consumers implementing their own operators. Potentially we could include some guidance docs in a later PR for creating your own custom operators.
> 
> PS. I see that that `NoInfer` can be applied in different ways with the same effect: `NoInfer<PatchSpec<T>>` or `PatchSpec<NoInfer<T>>` I have no real opinion on which is better or more obvious in terms of what it does. What do you think?

1. I'll try to update all operators as needed.
2. 👍
3. Will export `NoInfer` without any docs for now
PS. Great point. Not only does `PatchSpec<NoInfer<T>>` read more logically (we want to tell TS to not infer `T`, not some complex type built using `T`), but TS infers the type of the argument correctly that way such that I don't have to cast it as I claimed I "needed" to in [this comment](https://github.com/ngxs/store/pull/1806/files#r757595492).

markwhitfeld:
PS, regarding 2, can TS snippet check the type inferred (and therefore the intellisense auto-completion) for an argument at a call site. I would be amazed if any tool could test this crazy requirement!

markwhitfeld:
PPS. Amazing article by the way!
Leaving here for reference:
https://dev.to/davidshortman/weird-ts-types-using-contextual-typing-and-deferred-inference-to-plan-an-alien-conquest-bm8
PPSPS. Could you DM me on Twitter so that I can message you. Your DMs are currently closed.

packages/store/operators/src/compose.ts
```diff
@@ -1,6 +1,7 @@
 import { StateOperator } from '@ngxs/store';
+import { NoInfer } from './utils';
 
-export function compose<T>(...operators: StateOperator<T>[]): StateOperator<T> {
+export function compose<T>(...operators: NoInfer<StateOperator<T>[]>): StateOperator<T> {
```
david-shortman, line 4: sometimes, Typescript was happier about the inferred internals of the function when wrapping `NoInfer` around the entire parameter type. other times, it was happier only wrapping `T`. so I have implemented a mixture with the principle being whatever results in the smallest code change for correct inference.

packages/store/operators/src/update-item.ts
```diff
@@ -29,10 +36,11 @@ export function updateItem<T>(
     let value: T = null!;
     // Need to check if the new item value will change the existing item value
     // then, only if it will change it then clone the array and set the item
-    if (isStateOperator(operatorOrValue)) {
-      value = operatorOrValue(existing[index] as Readonly<T>);
+    const theOperatorOrValue = operatorOrValue as T | StateOperator<T>;
+    if (isStateOperator(theOperatorOrValue)) {
+      value = theOperatorOrValue(existing[index] as Readonly<T>);
```
david-shortman, line 41: Typescript will not narrow the type correctly, so I had to assert the type of `operatorOrValue` first.

david-shortman:
@markwhitfeld A build error is outputting an error that is not present in my IDE. When I put the relevant code into a TS playground using the same TS version as the repo, 3.9.7, there is also no error: [playground](https://www.typescriptlang.org/play?ts=3.9.7#code/C4TwDgpgBAysCGwIHlICdEHs0B4AqAfFALxQAUEAHgJYDOw1AdgOYBcUAShPACaaMAbEPgIBKEkTwBuAFAyAZgFdGAYwb8odAKqMeEeUwg8yAN3gDFEdvEYhR7Mxeh0oyvQcZGoAbxlQoaBDAimiMUKCQmPJQjpYkxKQA5G76hjyJsgC+ckqq6mF0AHLUAiKm5pbseFAAPlCMigICta66qZ489jEVzrT1jc11KR5evv6BwaHdTvGkDU21ddptI8axEKJZchHQXGDw1Gh44BAiJFDVVEi6fcBocQD8UABGmJgC3GFVUFcQN1Dycy0aBPV7vT5QKqyGQ7KCFTAASUY8gguEI50ulGuPD6TBRaFgUCeMEh9QgJlR0IA9AAqGl+GlQAAC+wwAFtpnEALRQABqPXCmE0jGBaGADOZrPgHIA2s99NgIAAFTC0aj5AC6UB5sEgKmoBi8TD0lEFwtFwE50Hl8kVABooJgwPlzAyqQplGpqBomBaEUg2WU-Fb2PCkfiRHbgzbFSq1fkHuwGmz5WgZF04IgUOgsLg9gcjicRDKNUQxgEgiEwrkvT6RajgP6IGzUKjcxQaPQmGxONw+IJhPnDsdIMXS10h4XR4QSz5g-4qVSoAAJeAUs0qAAWEBUAGsfpQwAJqPrgEIAdgoAADeYCK9QGw8a-DNJX+dQRcvHfwRTA6-re8VBsL9rwABivB1gE3X9rwAQgAqAAHcSmaCYq2vO5LDffx-ANcgihKconHEAAyEiDzoBgWHEcscIrSYwioSjuwfPpJxHU4Zw1WQ6Oyd9PyVNAnQbEAoDZeBRPlKAeBARhpRPB9dDE6hmE3S1GEwS0mPoLwY0Cd88LIWCAEE0AwEAADo6FM8yO2Y6iaPfcZKymGV1lY3t9mHItCG498+LolR+HoKAVAEfhoFIbSqOYCzaGPFQIDITY5Doj5LWNKhzlA6E6M-eEyS8YAhSAhYryKRQU1RbCcM-eA+kQ6B4ECULtz3FiryIcD6k068CFIcD+KXclURAKCWKoRLnX6KqCWQhY0KmQEBD-bd9Lowy9OVVV1W9RhYKgLrHLo3C2lNUgtrjXb+FgnicICnCwoiuKjxPJLMsoB1QIddz6s8gsOJEFK6MWsIns8O7Mi2IA).

Build error:

> ../../packages/store/operators/src/insert-item.ts(21,15): error TS2352: Conversion of type 'NoInfer<T>' to type 'RepairType<T>' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.
../../packages/store/operators/src/insert-item.ts(35,28): error TS2352: Conversion of type 'NoInfer<T>' to type 'RepairType<T>' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.

markwhitfeld:
@david-shortman Thank you for your thorough updates to this PR.
I will see if I can reproduce those build errors my side.

markwhitfeld:
@david-shortman I have fixed the build issues, but this lead to discovering another crazy typings challenge with the `iif` operator.
I have solved it (after a month of thrashing on this), but that would need to be resolved in a separate PR.
I have pushed to your branch (surprised that I was able to do so) and, hopefully, I can get this merged.

markwhitfeld (approved):

PR #1820, opened 2022-01-30
fix(storage-plugin): add storage fallback

marcelpotirnac:
Support storage-plugin when localstorage is not available

## PR Checklist

## PR Type

Support `storage-plugin` in browsers that deny access to `localStorage` or `sessionStorage`.
It adds a fallback Storage implementation.

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Applications using storage-plugin will not load when localStorage is not available with:
``` console
DOMException: Failed to read the 'localStorage' property from 'Window': Access is denied for this document.
```

Issue Number: N/A

## What is the new behavior?
Applications will be able to load and work without persisting the state after page reload.

```
[ ] Yes
[x] No
```


packages/storage-plugin/src/internals.ts
```diff
@@ -70,10 +70,52 @@ export function engineFactory(
...
+    } catch (e) {
+      console.warn('SessionStorage is not available, using a fallback implementation!', e);
+      return fallbackStorage();
+    }
   }
 
   return null;
 }
+
+export function fallbackStorage(): Storage {
```
splincode, line 97: Could you add unit test, please?

marcelpotirnac: Will try today

marcelpotirnac: does that look good?

splincode (approved):

docs/plugins/storage.md
```diff
@@ -248,3 +248,10 @@ In the migration strategy, we define:
 - `key`: The key for the item to migrate. If not specified, it takes the entire storage state.
 
 Note: Its important to specify the strategies in the order of which they should progress.
+
+### Fallback storage
+
+When default storage implementation is not available, we will default to a fallback implementation that will persist the state in memory until reload.
+This should avoid breaking apps upon loadingr or the need to create a custom storage engine to support this use case.
```
arturovt, line 255:
```suggestion
This should avoid breaking apps upon loading or creating a custom storage engine to support this use case.
```

docs/plugins/storage.md
```diff
@@ -248,3 +248,10 @@ In the migration strategy, we define:
...
+
+`Block third-party cookies and site data` from chromium browsers and related settings from modern browsers will disable localstorage implementation. Some browsers do this in their incognito mode by default.
```
arturovt, line 257:
```suggestion
`Block third-party cookies and site data` from chromium browsers, and related settings from modern browsers will disable local storage implementation. Some browsers do this in their incognito mode by default.
```

splincode:
@marcelpotirnac update CHANGELOG.md please

packages/storage-plugin/src/internals.ts
```diff
@@ -70,10 +70,52 @@ export function engineFactory(
...
-    return sessionStorage;
+    try {
+      const verifyKey = `${DEFAULT_STATE_KEY}-VERIFY`;
+      sessionStorage.setItem(verifyKey, '1');
+      sessionStorage.removeItem(verifyKey);
+      return sessionStorage;
+    } catch (e) {
+      console.warn('SessionStorage is not available, using a fallback implementation!', e);
+      return fallbackStorage();
+    }
```
arturovt, line 91: Seems like you have duplicated try-catch statements, you can simplify it:
```ts
const storage =
  options.storage === StorageOption.LocalStorage
    ? localStorage
    : options.storage === StorageOptions.SessionStorage
    ? sessionStorage
    : null;

if (storage === null) {
  return null;
}

try {
  const verifyKey = `${DEFAULT_STATE_KEY}-VERIFY`;
  storage.setItem(verifyKey, '1');
  storage.removeItem(verifyKey);
  return storage;
} catch (error) {
  // Caretaker note: we have still left the `typeof` condition in order to avoid
  // creating a breaking change for projects that still use the View Engine.
  if (typeof ngDevMode === 'undefined' || ngDevMode) {
    console.warn(
      `${
        options.storage === StorageOption.LocalStorage
          ? 'localStorage'
          : 'sessionStorage'
      } is not available, using a fallback implementation!`,
      error
    );
  }
  return fallbackStorage();
}
```

arturovt (changes requested):
Can you also updated the changelog?

packages/storage-plugin/src/internals.ts
marcelpotirnac: thanks @arturovt. 
I don't think this suggestion makes the code easier for me, but I added your suggestion. (probably only a method with try catch with a param of the store engine would have been something I should have added from the start).  Anyhow, small details 😄 

Can I ask, why are we hiding the warning there sometimes? I think we should always print a message.

packages/storage-plugin/tests/storage.plugin.fallback.spec.ts
```diff
@@ -0,0 +1,136 @@
...
+    const state: CounterStateModel = store.selectSnapshot(CounterState);
+
+    // Assert
+    expect(spyWarn).toHaveBeenCalledTimes(1);
+    expect(state.count).toBe(2);
+  });
+
+  it('storage engine should work with fallback implementation, when sessionStorage is not available', () => {
+    // Arrange
+    disableSessionStorage();
```
markwhitfeld, line 62: Not sure if jest automatically restores the `window` object after each test, but I think it would be safer if we wrapped this test with a `try...finally` and restore the original property on `window` afterwards.

packages/storage-plugin/tests/storage.plugin.fallback.spec.ts
```diff
@@ -0,0 +1,136 @@
...
+  }
+
+  beforeEach(() => {
+    disableLocalStorage();
+    disableSessionStorage();
+  });
+
+  it('storage engine should work with fallback implementation, when localStorage is not available', () => {
+    // Arrange
+    disableLocalStorage();
```
markwhitfeld, line 38: Same as other comment with restoring the `window` object.

markwhitfeld (commented):
@marcelpotirnac Thank you so much for your PR. I think that this is a great addition!
I apologise for taking so long to get to reviewing it. I have had a lot going on personally in the last few months.

I just have a question regarding the `fallbackStorage`...
What would the point of this be?
The whole reason for the storage plugin is for certain store values to persist past the page reload, but with this being in-memory it really doesn't add any benefit over returning a `null` engine (and therefore becoming a noop for the plugin).
Are there any specific use cases that I am not seeing where the in-memory fallback would be useful?

marcelpotirnac:
@markwhitfeld Thanks for checking this out. 
Your question is excellent and makes me realise that I should look at this problem differently while still focusing on the goal.
_The goal is to avoid crashing applications using the plugin._ 

I guess when I proposed this solution my mind was thinking to avoid any crash where the storageEngine would be used (the plugin is trying to use the engine after each update not just on the reload -> probably this one made me choose the initial path )
I also used similar technique in other applications where localStorage was not accessible, but here I failed to recognise that the in memory engine is the Store itself. 
Currently in applications I am solving this problem with a custom storage implementation. I decided to fix the problem upstream and rushed with the same solution (but we can do better/easier here).

I will try to clean this up next week.

markwhitfeld:
Thanks @marcelpotirnac! I look forward to the updated PR 🎉 .

packages/storage-plugin/tests/storage.plugin.fallback.spec.ts
marcelpotirnac: it does automatically restore based on my tests. I also tried to use a finally block

packages/storage-plugin/src/storage.plugin.ts
```diff
@@ -29,12 +28,11 @@ declare const ngDevMode: boolean;
...
   constructor(
     @Inject(NGXS_STORAGE_PLUGIN_OPTIONS) private _options: NgxsStoragePluginOptions,
-    @Inject(STORAGE_ENGINE) private _engine: StorageEngine,
-    @Inject(PLATFORM_ID) private _platformId: string
+    @Inject(STORAGE_ENGINE) private _engine: StorageEngine
   ) {}
 
   handle(state: any, event: any, next: NgxsNextPluginFn) {
-    if (isPlatformServer(this._platformId) && this._engine === null) {
+    if (this._engine === null) {
```
marcelpotirnac, line 35: I am not super confident on this one.
I checked the intial PR and its been said if is SSR and custom store. But on SSR the engine will default to null, and if is custom its not null anymore.

packages/storage-plugin/src/storage.plugin.ts
```diff
@@ -57,41 +55,42 @@ export class NgxsStoragePlugin implements NgxsPlugin {
...
+        } catch (e) {
+          // Caretaker note: we have still left the `typeof` condition in order to avoid
+          // creating a breaking change for projects that still use the View Engine.
+          if (typeof ngDevMode === 'undefined' || ngDevMode) {
+            console.error(
+              `Error ocurred while deserializing the ${key} store value, falling back to default store value, the value obtained from the store: `,
+              val
+            );
           }
+          continue;
```
marcelpotirnac, line 75: We don't want store an empty object when this doesn't work, it will break the default state values, so we just ignore this part.

marcelpotirnac:
@markwhitfeld Hopefully this one makes more sense, it's a simpler solution.
If I would have checked a previous work here to fix a similar issue, https://github.com/ngxs/store/pull/1119, There someone already tried to introduce an inmemory 😄

marcelpotirnac:
@arturovt @markwhitfeld Do you guys think that this PR makes sense now for you?

packages/storage-plugin/src/storage.plugin.ts
```diff
@@ -57,41 +55,42 @@ export class NgxsStoragePlugin implements NgxsPlugin {
...
-                `Error ocurred while deserializing the ${key} store value, falling back to empty object, the value obtained from the store: `,
-                val
-              );
-            }
-            val = {};
+        let val: any;
+        try {
+          val = this._engine.getItem(key!);
+          if (val === 'undefined' || val == null) {
+            continue;
```
markwhitfeld, line 62: This `continue;` here will cause an `undefined` or `null` value to skip running through the migrations, which is a change from previous behaviour, and potentially a breaking change for some production use cases.

markwhitfeld (changes requested):
Apologies for missing your update to this PR.
I think that the one change below may introduce a bug. Do you agree?

packages/storage-plugin/src/storage.plugin.ts
marcelpotirnac: I starred at this for a while and started looking at alternatives.
Once I reverted the code I realised this is the previous behaviour. I just reversed the condition like a guard, easier to read.
`if (val !== 'undefined' && val != null) {` <--
@markwhitfeld Are you thinking about something else?

markwhitfeld: Hi Marcel. Apologies for missing your response again.
When you inverted the condition and added the `continue;` code, the code execution will skip all of the rest of the code in the `for (const key of keys) {` loop and move to the next iteration. This will therefore skip the `if (this._options.migrations) {` section of code, which was not the case before.
I haven't verified what this subsequent code would do with a value of `null` or `'undefined'` (which could potentially be an edge case that has not been tested and exhibits some reasonable behavior, or even a bug). But this change is nevertheless an unintended change to program flow.

Do you think that there is a different way to make your changes and preserve the program flow to avoid this side effect?

markwhitfeld (changes requested):

marcelpotirnac:
@markwhitfeld going to close this PR.

The latest changes to the plugin solved the initial problem.
Using Browsers with all storage options blocked works now.

PR #1824, opened 2022-02-10
feat(selector-utils): Impl. selector utils

Carniatto:
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: N/A

```
[ ] Yes
[ ] No
```


packages/store/tests/property-selectors.spec.ts
```diff
@@ -0,0 +1,98 @@
...
+  });
+
+  it('Passing undefined', () => {
+    const slices = createPropertySelectors(undefined);
+
+    expect(slices).toBe(null);
+  });
+
+  it('Passing empty object', () => {
+    const slices = createPropertySelectors({});
```
markwhitfeld, line 37: Passing through anything that could potentially be a selector (a function or an object) should result in a selector being returned.

I guess the other 'empty' test case here is if the selector that you are slicing returns an empty object, or an object that does not contain the property that you are slicing out. Then the slice selector for the missing property should return undefined.

markwhitfeld (approved):

packages/store/tests/property-selectors.spec.ts
```diff
@@ -0,0 +1,98 @@
...
+    });
+
+    const store: Store = TestBed.inject(Store);
+    const myState = store.selectSnapshot<MyStateModel>(MyState);
+    const slices1 = createPropertySelectors<MyStateModel>(MyState);
+    const slices2 = createPropertySelectors<MyStateModel>(MyState);
+
+    expect(slices1.property1(myState) === slices2.property1(myState)).toBeTruthy();
+    expect(slices1.property2(myState) === slices2.property2(myState)).toBeTruthy();
+  });
```
markwhitfeld, line 97: So, I guess the memoisation tests should be these cases:
- Memoisation of the selectors: If I use createPropertySelectors to create a 'slices' object then if I repeatedly ask for the selector for property 1, it should return the same (`===`) selector every time. If createPropertySelectors is called again to create another 'slices' object, then I don't think that the slices object nor the selectors for the same property from each slices object should be equal.
- Memoisation of the values (not fetched again): This is a contrived scenario, but focuses on the memoisation of the property values by the input object. Let's say that you have a prop `foo` that returns a new value every time it is called (we could implement this with a property getter returning a new object every time... or a randomly generated number). Then if the `foo` property selector (which was created by the slicer) is called with the same object twice then it should return the same result (ie. the getter would only ever have been called once).

arturovt (approved):
We've been already using this implementation in production for the last ~2 years. I don't have anything to review here since it's obvious and clear.

Carniatto:
> We've been already using this implementation in production for the last ~2 years. I don't have anything to review here since it's obvious and clear.

Same for me 💗

markwhitfeld (approved):

docs/advanced/selector-utils.md
```diff
@@ -142,8 +150,10 @@ export class AnimalsSelectors {
   selector: 'my-zoo'
   template: `
     <h1> Zebras and Pandas </h1>
-    <ol>
-      <li *ngFor="zebra in zebrasAndPandas$ | async"> {{ zebra }} </li>
+    <h1> Pandas and Zoos </h1>
```
Carniatto, line 153: Do we want two titles here?

packages/store/tests/selectors/create-model-selector.spec.ts
```diff
@@ -0,0 +1,262 @@
...
+        error = err as Error;
+      }
+      // Assert
+      expect(error).not.toBeNull();
+      expect(error?.message).toMatchInlineSnapshot(
+        `"[createModelSelector]: The value provided as the selector for the 'test' property is not a valid selector."`
+      );
+    });
+
+    it('should fail if a class that is not a selector is provided in the selector map', () => {
```
Carniatto, line 170: This test seems to be duplicated

packages/store/tests/selectors/create-model-selector.spec.ts
```diff
@@ -0,0 +1,262 @@
...
+        hello: prop3.hello,
+      });
+      // Assert
+      expect(modelSelector).toBeDefined();
+      const snapshot1 = store.selectSnapshot(modelSelector);
+      expect(snapshot1).toStrictEqual({ bar: 'Tada', hello: 'there' });
+      patchState({ property2: [2, 4] });
+      const snapshot2 = store.selectSnapshot(modelSelector);
+      expect(snapshot2).toBe(snapshot1);
+      expect(snapshot1).toStrictEqual({ bar: 'Tada', hello: 'there' });
```
Carniatto, line 259: For memoisation should we test that `this.store.select` observable does not emit?

markwhitfeld: This is implied because the snapshot returns the exact same object (same reference).
`select` has a `distinctUntilChanged` so it would not emit.
This is just the simpler way to test this.

docs/advanced/selector-utils.md
markwhitfeld: Good pickup. Fixed.

Issue #1764, opened 2021-06-18
[BUILD TOOLS]: Replace Bundlesize


markwhitfeld:
### Description

Bundlesize has issues comparing PR bundle sizes to master.
Updating the bundlesize manually in PRs is a real pain, and it doesn't give a comparison to master (because of issue https://github.com/siddharthkp/bundlesize/issues/308). There must be something better!
We are probably also abusing the maxSize threshold prop by keeping it so close to the actual size and updating it all the time.

We need to investigate alternatives!

Some thoughts:
- It would be nice to commit the actual sizes to keep a history
- We really need to be able to compare a PR to master to see the change in size
- It should not interfere with PRs so much, the check and comparison is important, but fiddling values all the time is tedious
- It would be great if there was some easier way to automate the bundle size commits in PRs (potentially a commit from CI - although we would not want this to be recursive!)

We really need 2 different things...
- Report the size difference to master from a PR
- Commit a list of the current bundle sizes so that we can have a history
Maybe these could come from different tools

markwhitfeld:
This might be an easy drop-in replacement:
https://bundlewatch.io/#/getting-started/using-bundlewatch
I think it is a fork of the other project

Bundlewatch has an action too
https://github.com/jackyef/bundlewatch-gh-action

We could use something like this to push updates back to the PR... as long as we have a way to save or retrieve the bundle sizes as a json file. I think that we could use one of the internal bundlesize apis to fetch the file details
https://github.com/marketplace/actions/git-auto-commit

markwhitfeld:
Could be useful:
https://brennerm.github.io/posts/trigger-github-actions-on-pr-close.html

markwhitfeld:
This looks like a great alternative that was released recently:
https://github.com/LironEr/bundlemon

markwhitfeld:
@splincode This looks like a great alternative that was released recently:
https://github.com/LironEr/bundlemon

splincode:
@markwhitfeld it's really interesting! what about add github action?
https://github.com/LironEr/bundlemon-github-actions

markwhitfeld:
@splincode Yeah, it looks like great integration!

arturovt:
@rfprod since you’re a pro at GitHub actions, would you want to take this one also?

PR #1828, opened 2022-03-02
perf(store): run change detection once for all selectors when asynchronous action has been completed

arturovt:


packages/store/tests/zone.spec.ts
```diff
@@ -25,53 +25,10 @@ describe('zone', () => {
     }
   }
 
-  describe('[store.select]', () => {
```
arturovt, line 28: Removed it since it duplicates the below test.

splincode (approved):

packages/store/src/internal/state-stream.ts
```diff
@@ -3,13 +3,19 @@ import { BehaviorSubject } from 'rxjs';
...
 @Injectable()
 export class StateStream extends BehaviorSubject<PlainObject> {
-  constructor() {
+  constructor(private _ngxsExecutionStrategy: InternalNgxsExecutionStrategy) {
     super({});
   }
+
+  next(value: PlainObject): void {
+    this._ngxsExecutionStrategy.leave(() => super.next(value));
+  }
```
markwhitfeld, line 20: I think that this could have unintended side effects for other aspects of the state management (like plugins etc.).
I think it would be a better approach if `Store` had a derived state stream that calls the `leave` to change the execution context, and then is shared using the `shareReplay` operator.
All selects would then use this, and it would call `leave` only once for any state change across all active selectors. What do you think @arturovt?
This would be used by the `select` and `subscribe` methods of the `Store`.

markwhitfeld (changes requested):

packages/store/src/internal/state-stream.ts
arturovt: @markwhitfeld I've made some changes, could you have a look. This is the only approach that doesn't break unit tests, using `shareReplay` breaks a lot of tests, I didn't dig why 😄

packages/store/src/store.ts
```diff
@@ -87,7 +98,9 @@ export class Store {
    * Allow the user to subscribe to the root of the state
    */
   subscribe(fn?: (value: any) => void): Subscription {
-    return this._stateStream.pipe(leaveNgxs(this._internalExecutionStrategy)).subscribe(fn);
+    return this._selectableStateStream
+      .pipe(leaveNgxs(this._internalExecutionStrategy))
```
markwhitfeld, line 102: Is this still necessary since we are using the new shared stream?

packages/store/tests/issues/issue-933-selectors-causing-ticks.spec.ts
```diff
@@ -53,7 +53,7 @@ describe('Selectors within templates causing ticks (https://github.com/ngxs/stor
   class TestModule {}
 
   it(
-    'should run change detection whenever the selector emits after asynchronous action has been completed',
+    'should not run change detection whenever the selector emits after asynchronous action has been completed',
```
markwhitfeld, line 56: Potentially better to say "should run change detection once for all selectors when acting action has been completed"

packages/store/src/store.ts
arturovt: This is necessary in case the Store class is subscribed outside of the Angular zone.

markwhitfeld: I thought that the new `_selectableStateStream` would handle ensuring that the subscribers are notified in the correct zone?
Or is this for the case where we are reading the current value that already exists in the stream?
If this is the case, then would we not also have to do this if the `store.select` is subscribed to outside of the Angular Zone?

packages/store/src/store.ts
```diff
@@ -25,6 +35,7 @@ export class Store {
     @Inject(INITIAL_STATE_TOKEN)
     initialStateValue: any
   ) {
+    (this._selectableStateStream as ConnectableObservable<PlainObject>).connect();
```
markwhitfeld, line 38: I think that there may be an issue with the explicit `connect()` where, if the subscriber count reaches 0, then the stream will disconnect and I think it won't reconnect. Potentially we would want to add a `refCount()` after the `publishReplay(1)` pipe in the observable declaration and get rid of the manual `connect()`.
We definitely need to have a test that checks a subscription to the select and then subscription (taking the subscriptions to zero) and then after a state change, subscribing to a select again and checking it gets the latest state as well as further updates.
This article goes into great depth to explain the differences:
https://itnext.io/the-magic-of-rxjs-sharing-operators-and-their-differences-3a03d699d255

packages/store/src/store.ts
markwhitfeld: I thought that the new `_selectableStateStream` would handle ensuring that the subscribers are notified in the correct zone? But maybe there would be no guarantee for new subscriptions where we first receive the current value that already exists in the stream?
If this is the case, then would we not also have to do this if the `store.select` is subscribed to outside of the Angular Zone?

markwhitfeld (changes requested):

packages/store/src/store.ts
arturovt: The `_selectableStateStream` cannot ensure that completely. Given this example:
```ts
constructor(store: Store, ngZone: NgZone) {
  ngZone.runOutsideAngular(() => {
    setTimeout(() => {
      store.subscribe(() => console.log(Zone.current)); // will log angular
    }, 3000);
  });
}
```
Otherwise, it'll log `<root>`.

markwhitfeld:
@arturovt is this still in draft? 
Is there anything you need from me to get this ready?

arturovt:
Glad you're back! No, I've made is as a draft initially, but this is ready now to be merged (if all requirements are passed).

markwhitfeld (approved):
I'm happy with this. Just need to fix the failing bundlesizes ;-)
Might be worth merging master into this branch before doing this too.

arturovt:
Updated, hope is passes, this bundlesize is such annoying guy... 😸

markwhitfeld:
I couldn't agree more!!!
https://github.com/ngxs/store/issues/1764

PR #1830, opened 2022-03-10
fix(store): complete the state stream once the root view is removed

arturovt:


splincode (approved):
You are amazing, cool PR!

markwhitfeld (approved):
100% agree. You are amazing @arturovt !!!!

PR #1831, opened 2022-03-15
fix(store): be more explicit when checking if Angular is in test mode

arturovt:
@markwhitfeld I've noticed that this function doesn't work in ESP app because there's no `MockNgModuleResolver` anymore (the `@angular/compiler/testing`) just doesn't export anything now. Thee `MockNgModuleResolver` was available before the VE support has been removed.

We couldn't see that in NGXS repository since our packages are version 9.

### DIFF
--- a/packages/store/internals/src/angular.ts
+++ b/packages/store/internals/src/angular.ts
@@ -1,30 +1,10 @@
-import { getPlatform, COMPILER_OPTIONS, CompilerOptions, PlatformRef } from '@angular/core';
-import { memoize } from './memoize';
+import { ɵglobal } from '@angular/core';
 
-/**
- * @description Will be provided through Terser global definitions by Angular CLI
- * during the production build. This is how Angular does tree-shaking internally.
- */
-declare const ngDevMode: boolean;
-
-function _isAngularInTestMode(): boolean {
-  const platformRef: PlatformRef | null = getPlatform();
-  if (!platformRef) return false;
-  const compilerOptions = platformRef.injector.get(COMPILER_OPTIONS, null);
-  if (!compilerOptions) return false;
-  const isInTestMode = compilerOptions.some((item: CompilerOptions) => {
-    const providers = (item && item.providers) || [];
-    return providers.some((provider: any) => {
-      return (
-        (provider && provider.provide && provider.provide.name === 'MockNgModuleResolver') ||
-        false
-      );
-    });
-  });
-  return isInTestMode;
+export function isAngularInTestMode(): boolean {
+  return (
+    typeof ɵglobal.__karma__ !== 'undefined' ||
+    typeof ɵglobal.jasmine !== 'undefined' ||
+    typeof ɵglobal.jest !== 'undefined' ||
+    typeof ɵglobal.Mocha !== 'undefined'
+  );
 }
-
-export const isAngularInTestMode =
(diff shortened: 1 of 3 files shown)


packages/store/internals/src/angular.ts
```diff
@@ -8,20 +9,12 @@ import { memoize } from './memoize';
...
-      );
-    });
-  });
-  return isInTestMode;
+  return (
+    typeof ɵglobal.__karma__ !== 'undefined' ||
+    typeof ɵglobal.jasmine !== 'undefined' ||
+    typeof ɵglobal.jest !== 'undefined' ||
+    typeof ɵglobal.Mocha !== 'undefined'
+  );
```
splincode, line 17: What is the minimum version of Angular that it supports?

splincode (approved):

packages/store/internals/src/angular.ts
arturovt: It’s cross-version, because Angular has been using Karma as a test runner from the very beginning. Jest is available when using `@angular-builders/jest` and Nrwl Nx Jest runner. Mocha is also available since zone.js has a Mocha plugin. Otherwise no other frameworks are available for running tests.

packages/store/tests/utils/angular.spec.ts
```diff
@@ -1,82 +1,73 @@
...
 describe('[utils/angular]', () => {
   describe('isAngularInTestMode', () => {
-    function resetEnv() {
+    // Just an empty object so `typeof !== undefined` will be truthy.
+    const nonUndefinedValue = {};
+
+    function resetMemoization() {
       const fn = <any>isAngularInTestMode;
       // Reset the memoization
       fn && fn.reset && fn.reset();
```
markwhitfeld, line 13: I don't think it is necessary to memoise this anymore because the checks are so simple!
We could remove the memoisation from the lib export and remove this testing code.

packages/store/tests/utils/angular.spec.ts
```diff
@@ -1,82 +1,73 @@
...
-    it(`should return true if the Angular Test Module has been bootstrapped`, () => {
-      // Arrange
-      platformBrowserDynamicTesting();
-      // Act
-      const result = isAngularInTestMode();
-      // Assert
-      expect(result).toEqual(true);
+    it('should return true if `__karma__` is available', () => {
+      // Arrange & act & assert
+      const __karma__ = ɵglobal.__karma__;
```
markwhitfeld, line 26: We could have a test helper (like the `freshPlatform` one you created) that wraps the test execution with a `try..finally` and stores and restores the values on `ɵglobal` (or deletes them again if they weren't there to begin with).

markwhitfeld (commented):
2 small improvements

packages/store/tests/utils/angular.spec.ts
arturovt: Removed memoization.

packages/store/tests/utils/angular.spec.ts
arturovt: I'd like to keep these tests simple for now. We could've thought about the `isolateGlobalTestProps` implementation in the future.

arturovt:
Mark, I’ve merged it, just wanna test it in our app. I can make further changes in another PR if needed.

PR #1832, opened 2022-03-16
fix(store): be more explicit when checking if Angular is in test mode

arturovt:


packages/store/tests/utils/angular.spec.ts
```diff
@@ -2,7 +2,7 @@ import { ɵglobal } from '@angular/core';
 
 import { isAngularInTestMode } from '@ngxs/store/internals';
 
-describe('[utils/angular]', () => {
+xdescribe('[utils/angular]', () => {
```
markwhitfeld, line 5: Yes, I'm not sure how we would write tests for this either ;-)

markwhitfeld (approved):

PR #1841, opened 2022-05-14
feat(storage-plugin): allow overriding the default state key

arturovt:
We have a hardcoded default state key, which stores the entire serialized state. We don't allow developers to override the `@@STATE` key. This means that it's not possible to use the plugin in a micro-frontend application where developers might want to use storage plugins in different apps since all plugins will use the same state key.


splincode (approved):

markwhitfeld (changes requested):
I think that having an option called `defaultStateKey` wouldn't make sense to the user. The fact that there is a default state key is actually an internal implementation detail.

We are trying to provide the user some control over the keys for a micro-frontend application and prevent conflicts. The tricky thing here is that there are two cases that need to be covered:
- if there is no state key provided then the entire state is stored and we woud want to control what key to use in storage.
- if there are one or more state keys provided then we would want to ensure that these keys (that come from the state names) do not conflict with other MFE applictions.

(I'll add my proposal in another comment, but if it warrants further discussion we should open an issue for this)

markwhitfeld:
I propose that we could have a `namespace` option which is used in the following way:
- When we store the entire state, then we should store the entire state under this key in the storage.
- When we store multiple explicit keys then we could prefix the key names with this to create the unique storage keys for the states. For example, with a `namespace` of `@@APP_ONE` and a state with the name `contacts` being stored, then the state would be stored under a key `@@APP_ONE:contacts` (we could use a dot, colon, or whatever makes this namespace make sense).

What do you think?

packages/storage-plugin/src/storage.module.ts
```diff
@@ -31,12 +32,16 @@ export class NgxsStoragePluginModule {
...
+          deps: [USER_OPTIONS, DEFAULT_STATE_KEY]
         },
         {
           provide: STORAGE_ENGINE,
           useFactory: engineFactory,
           deps: [NGXS_STORAGE_PLUGIN_OPTIONS, PLATFORM_ID]
+        },
+        {
+          provide: DEFAULT_STATE_KEY,
+          useValue: (options && options.defaultStateKey) || '@@STATE'
```
markwhitfeld, line 44: PS. I see that you are using an InjectionToken and a provider to communicate this value internally.
Is there any reason to do this over just having the additional property in the options?
Does this give any additional locus of control for MFE's?

arturovt: There's no particular reason for this. I was doing a rapid programming 🤣

arturovt:
@markwhitfeld yeah that seems reasonable, I haven't thought from the start that the default state key is not applicable to the user, the namespace is a great idea. I've made updates but I feel like I've made some mess 🤣

markwhitfeld (changes requested):
I think we need to take the setting of the default state names it of this, and just have the `namespace` option. It satisfies the initial Module Federation use case and the key conflict use case too.

markwhitfeld:
@arturovt what do you think? ☝️

arturovt:
> I think we need to take the setting of the default state names it of this, and just have the `namespace` option. It satisfies the initial Module Federation use case and the key conflict use case too.

Can you give a quick example please? I'm a bit confused with that.

packages/storage-plugin/src/internals.ts
```diff
@@ -77,3 +77,9 @@ export function engineFactory(
 
   return null;
 }
+
+export function prefixNamespace(key: string, namespace: string | undefined): string {
+  // Prepends the `namespace` option to any key if it's been provided by a user.
+  // So `@@STATE` becomes `my-app:@@STATE`.
+  return namespace ? `${namespace}:${key}` : key;
+}
```
markwhitfeld, line 85: I think we should rename this to `getStorageKey`, and then pass the options object instead of the namespace:
```suggestion
export function getStorageKey(key: string, options: NgxsStoragePluginOptions | undefined): string {
  // Prepends the `namespace` option to any key if it's been provided by a user.
  // So `@@STATE` becomes `my-app:@@STATE`.
  return options && options.namespace ? `${options.namespace}:${key}` : key;
}
```

packages/storage-plugin/src/storage.plugin.ts
```diff
@@ -64,7 +64,8 @@ export class NgxsStoragePlugin implements NgxsPlugin {
           continue;
         }
 
-        let storedValue: any = this._engine.getItem(key!);
+        const namespacedKey = prefixNamespace(key, this._options.namespace);
```
markwhitfeld, line 67:
```suggestion
        const storageKey = getStorageKey(key, this._options);
```

markwhitfeld (changes requested):

packages/storage-plugin/src/internals.ts
arturovt: Love it.

packages/storage-plugin/src/symbols.ts
```diff
@@ -11,7 +11,12 @@ export interface NgxsStoragePluginOptions {
   /**
    * Key for the state slice to store in the storage engine.
    */
-  key?: undefined | StorageKey;
+  key?: StorageKey;
+
+  /**
+   *
+   */
+  namespace?: string;
```
markwhitfeld, line 19: Gonna need dem doczez!

markwhitfeld (changes requested):
Just a description in the options, and maybe in the docs too ;-)
PS. A description for this PR would also make the release announcement easier too.

markwhitfeld (approved):
Approving, but marking this as a 'draft' PR to block merging until we have the next patch release out.

PR #1850, opened 2022-06-04
feat: add support for Angular 14

arturovt:


markwhitfeld:
Can't quite figure out the reason for the failure here. Rerunning to see if it is a circle ci issue.

arturovt:
I've removed `test-config.helper.ts` because it's not used and it's never been used, I don't even know why it was created, I'm guessing it was copied from some repo by mistake. And it keeps being copied over integrations. The same with readme files, they're just copied from old integration projects. Just want to keep it clean w/o copying unneeded files.

I'll keep debugging the failure.

arturovt:
@markwhitfeld I just re-run workflow and it seems to be green now.

markwhitfeld:
Thanks! Makes sense regarding those files.
Circle CI seems to be giving so many issues lately!

markwhitfeld (approved):

markwhitfeld:
Argh. We forgot the changelog update. I have made a commit on master.
We really need a bot to add this to the checks or something (...or conventional commits!).

PR #1862, opened 2022-06-15
ci: remove packages from bundlesize and leave only production bundles

arturovt:
The bundlesize honestly makes me sad while doing any code updates. We don't really benefit from checking unminified files and has never had benefit before. E.g., the Angular teams cares only about production bundles. When I do tree-shaking PR I need to increase the size of unminified files, but then I need to decrease the size of production bundles, this leads to a paradox.
Objectively, only production size matters, e.g. if we add the code that does some development only checks, then the production bundle size will not be affected.


markwhitfeld (commented):
I do feel like there is value in checking these bundle sizes because there are many tools that measure these bundle sizes as part of measuring the "impact" of using a library.
Whether their approach is valid or not is debatable.
What do you think?

Potentially #1764 is something we should try soon.

arturovt:
Closing per #1865

PR #1863, opened 2022-06-16
fix(storage-plugin): provide more meaningful error message when the storage quota exceeds

arturovt:


markwhitfeld:
@arturovt I have classified this as a `fix` instead of a `feature` because it doesn't offer any new capabilities of the lib.

PR #1864, opened 2022-06-17
ci(publish): check the repo owner before executing publish steps

rfprod:
- [x] publish packages only if the repo owner is 'ngxs';

## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[x] CI related changes
[ ] Other... Please describe:
```

Currently the publish jobs are triggered always, in forked repos as well.

Issue Number: N/A

## What is the new behavior?

The publish jobs are triggered only if the repo owner is `ngxs`.
The publish jobs are not triggered from forks.

```
[ ] Yes
[x] No
```

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -145,7 +145,7 @@ jobs:
         uses: actions/checkout@v3
 
       - name: Setup
-        if: ${{ github.repository_owner == 'ngxs' }}
+        if: github.repository_owner == 'ngxs'
         uses: ./.github/actions/setup
         with:
           node-version: 16.x
@@ -153,11 +153,11 @@ jobs:
           github-sha: ${{ github.sha }}
 
       - name: Check @ngxs build artifacts
-        if: ${{ github.repository_owner == 'ngxs' }}
+        if: github.repository_owner == 'ngxs'
         uses: ./.github/actions/check-build-artifacts
 
       - name: Production release - publish tagged builds to all @ngxs packages
-        if: ${{ github.repository_owner == 'ngxs' }}
+        if: github.repository_owner == 'ngxs'
         run: |
           echo "//registry.npmjs.org/:_authToken=${NPM_TOKEN}" > ~/.npmrc
           yarn publish:tagged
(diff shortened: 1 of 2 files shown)


rfprod:
@markwhitfeld please check this PR.
Test run before finalization https://github.com/ngxs/store/runs/6934045960?check_suite_focus=true
The `check-repo-owner` action output can be used to skip unnecessary steps for forks.

.github/workflows/release.yml
```diff
@@ -144,17 +144,24 @@ jobs:
       - name: Checkout sources
         uses: actions/checkout@v3
 
+      - name: Check repository owner
+        id: check-repo-owner
+        uses: ./.github/actions/check-repo-owner
+
       - name: Setup
+        if: ${{ steps.check-repo-owner.outputs.ngxs == 'true' }}
```
markwhitfeld, line 152: Just wondering if it is worth adding the composite action for this, or if we could just have the following instead:
```suggestion
        if: ${{ github.repository_owner == 'ngxs' }}
```

markwhitfeld: PS. We could also add this `if` at the job level instead.
This example from the docs actually does almost the same as what you are doing here:
https://docs.github.com/en/actions/using-jobs/using-conditions-to-control-job-execution#example-only-run-job-for-specific-repository

.github/workflows/release.yml
```diff
@@ -145,16 +145,19 @@ jobs:
         uses: actions/checkout@v3
 
       - name: Setup
+        if: ${{ github.repository_owner == 'ngxs' }}
```
arturovt, line 148: Should we remove braces?

> When you use expressions in an if conditional, you may omit the expression syntax (${{ }}) because GitHub automatically evaluates the if conditional as an expression.

markwhitfeld (changes requested):
Great! Can we move the `if` to the job level as opposed to the step level?

rfprod:
> Great! Can we move the `if` to the job level as opposed to the step level?

@markwhitfeld updated

markwhitfeld (approved):

Issue #1866, opened 2022-06-20
🐞[BUG]: 'RouterState' class should be decorated with @Injectable() right after the @State() decorator


emzet:
### Affected Package
The issue is caused by package @ngxs/router-plugin

### Description
When running `ngcc`, there is an compilation error (see below) also when using `Jest` in Angular via `jest-preset-angular` there is warning in console (due to compilation error).

## 🔬 Minimal Reproduction
Console log is enough to determine the the problem (but in case if really needed I can prepare stackblitz demo).

## 🔥 Exception or Error
Error when running ngcc:

<pre><code>
Error: Failed to compile entry-point @ngxs/router-plugin (`esm5` as esm5) due to compilation errors:
node_modules/@ngxs/router-plugin/esm5/src/router.module.js:9:5 - error NG3004: Unable to import class NgxsFeatureModule.
  The symbol is not exported from {PROJECT_PATH}/node_modules/@ngxs/store/index.d.ts (module '@ngxs/store').

9 var NgxsRouterPluginModule = /** @class */ (function () {
      ~~~~~~~~~~~~~~~~~~~~~~

  node_modules/@ngxs/store/src/modules/ngxs-feature.module.d.ts:11:1
     11 export declare class NgxsFeatureModule {
        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     12     constructor(_store: Store, internalStateOperations: InternalStateOperations, factory: StateFactory, states: StateClassInternal<any, any>[][] | undefined, lifecycleStateManager: LifecycleStateManager);
        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    ...
     16     static ɵinj: ɵngcc0.ɵɵInjectorDeclaration<NgxsFeatureModule>;
        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     17 }
        ~
    The class is declared here.
</code></pre>

[...]

PR #1867, opened 2022-06-21
fix(store): do not check if the state class is injectable within the decorator since the `ɵprov` will not exist in JIT mode

arturovt:
Issue: #1866


markwhitfeld (commented):
Looking good, but is there a test that we could add that exposes this issue and demonstrates that it is fixed?

arturovt:
No. This issue is reproducible at ASI with Jest tests right now, all tests that use the router plugin warn that it's not decorated with injectable. But with this change there's no warning.

PR #1870, opened 2022-06-22
feat(store): warn on unhandled actions

arturovt:


markwhitfeld:
@arturovt Could you add a description for this PR and the motivation for the feature?

packages/store/src/public_api.ts
```diff
@@ -46,3 +46,7 @@ export { NgxsExecutionStrategy } from './execution/symbols';
 export { ActionType, ActionOptions } from './actions/symbols';
 export { NoopNgxsExecutionStrategy } from './execution/noop-ngxs-execution-strategy';
 export { StateToken } from './state-token/state-token';
+export {
+  UnhandledActionsLogger,
```
markwhitfeld, line 50: Is it necessary to export this type?
How would a user of the library use this, or when would they need to reference it?

packages/store/src/public_api.ts
```diff
@@ -46,3 +46,7 @@ export { NgxsExecutionStrategy } from './execution/symbols';
...
+  ignoreActions
```
markwhitfeld, line 51: This `ignoreActions` API feels like it doesn't match the rest of the public API of the library.

packages/store/tests/unhandled-actions.spec.ts
```diff
@@ -0,0 +1,85 @@
...
+  class FireAndForget {
+    static readonly type = 'Fire & forget';
+  }
+
+  const plainObjectAction = { type: 'Fire & Forget' };
+
+  const testSetup = (warnOnUnhandledActions = true) => {
+    TestBed.configureTestingModule({
+      imports: [NgxsModule.forRoot([], { warnOnUnhandledActions })]
+    });
```
markwhitfeld, line 15: I propose a bit of a different API for this.
No config option, or `ignoreActions` utility, but rather a module that is provided:
```suggestion
    TestBed.configureTestingModule({
      imports: [
        NgxsModule.forRoot([]),
        ngDevMode && NgxsDevelopmentModule.forRoot({ 
          warnOnUnhandledActions: true
          /* or supply an object:
            warnOnUnhandledActions: { 
              ignore: [ 'SomeActionType', FooActionClass ] 
            }
          */
        })
      ]
    });
```
This keeps the API consolidated and clear, and follows the existing conventions.

markwhitfeld (changes requested):
I think that the API for this needs to be revised.
See the last comment.

packages/store/tests/unhandled-actions.spec.ts
arturovt: Love it.

markwhitfeld (approved):

Issue #1875, opened 2022-07-10
🚀[FEATURE]: Give back control about NgxsConfig.developmentMode 


maku:
### Relevant Package

This feature request is for @ngxs/store

### Description
Currently (since version 3.7.4) it seems that the developmentMode parameter in NgxsConfig has no effect as long the Ivy compiler is used.  This effects how object freezing is handled. Objects are freezed when dev mode is active without the possibility to overrule this.

e.g. The point is that my app works with angular version 14.0.5 and @ngxs version 3.7.3 but not with version 3.7.4

### Describe the problem you are trying to solve
This breaks existing code without the possibility to set the behaviour via a config flag (because objects are freezed by ngxs based on environment)

### Describe the solution you'd like
Give back the behaviour to set the developmentMode from the outside to handle object freezing.

evkw:
I have the same issue. Seems to also be causing issues with internal NGXS things as well 
![image](https://user-images.githubusercontent.com/15670238/178409019-87332cf6-1533-45e7-8da1-35a72aac698c.png)

arturovt:
@evkw that's not related.

@maku we want to tree-shake `deepFreeze` in production mode since it's recommended to be used only in development to ensure you're not doing state mutations. The `deepFreeze` is used when the development mode is enabled and it seems weird if you keep development mode enabled in production.

evkw:
@arturovt the stack trace goes into deepFreeze and is resolved when reverting to an earlier version of ngxs.

We have an app that's been in production for a few years now with the flag set to true. I get it's best practice to deepFreeze in dev and if errors surfaced at compile time it would be an easy pill to swallow, but the app I work on has a few hundred lazy loaded feature states that now might be a time bomb because I can't guarantee every line of code in a feature state is hit with e2e. This change basically locks us at 3.7.3. 

Terser probably works fine but it's another layer of complexity where a configuration was fine before.

I also understand it puts us at risk of future Dev in NGXS causing issues with mutated state.. but being stuck at 3.7.3 is probably more of a risk.

maku:
> @evkw that's not related.
> 
> @maku we want to tree-shake `deepFreeze` in production mode since it's recommended to be used only in development to ensure you're not doing state mutations. The `deepFreeze` is used when the development mode is enabled and it seems weird if you keep development mode enabled in production.
> 
> What I can propose is that we provide a variable called `ngxsForceFreeze` that you can explicitly provide through Terser and it will always freeze objects based on your Terser configuration.

[...]

PR #1878, opened 2022-07-17
fix(store): give back control to `developmentMode` config property

arturovt:
Issue: #1875 

The previous change didn't allow to disable `deepFreeze` during the development, since it was always enabled in Ivy, because I relied on the `ngDevMode`. We should respect `developmentMode` during the development only and then tree-shake it in production.


markwhitfeld (commented):
Looking good. Could you add a description with a motivation for this change, and a link to the issue that it resolves (if one has been logged).

arturovt:
> Looking good. Could you add a description with a motivation for this change, and a link to the issue that it resolves (if one has been logged).

UPD.

markwhitfeld (approved):

Issue #1880, opened 2022-07-18
🐞[BUG]: Old value emitted in new select after a reset if a select occurred prior to the reset


SLKnutson:
### Affected Package
@ngxs/store

### Is this a regression?
Yes, worked correctly prior to version 3.7.4

### Description
Old value emitted in new select after a reset if a select has already occurred

## 🔬 Minimal Reproduction

https://stackblitz.com/edit/ngxs-repro-mth3x9?file=src/app/app.component.ts

Look at the console. The first select value should match the snapshot (100), instead it initially shows the old value (3), and then goes to the new value (100)

## Environment

Libs:
- @ngxs/store version: 3.7.4

node & browser
 
 
 ## Thanks!

markwhitfeld:
Great news! [v3.7.5](https://github.com/ngxs/store/releases/tag/v3.7.5) has been released and it includes a fix for this issue.
We are closing this issue, but please re-open it if the issue is not resolved.
Please leave a comment in the [v3.7.5 release discussion](https://github.com/ngxs/store/discussions/1909) if you come across any regressions with respect to your issue.

PR #1883, opened 2022-07-24
fix(store): do not use `refCount()` since it makes the stream cold

arturovt:
Issue: #1880 

I don't know if there're other ways to fix that issue since using `refCount()` just makes the stream cold when there're no observers, but `publishReplay` buffers the last received value. See the issue and unit test.

The `shareReplay` increases the number of ticks by 1, but fixes the issue.

### DIFF
--- a/packages/store/src/store.ts
+++ b/packages/store/src/store.ts
@@ -1,14 +1,7 @@
 // tslint:disable:unified-signatures
 import { Inject, Injectable, Optional, Type } from '@angular/core';
 import { Observable, of, Subscription, throwError } from 'rxjs';
-import {
-  catchError,
-  distinctUntilChanged,
-  map,
-  publishReplay,
-  refCount,
-  take
-} from 'rxjs/operators';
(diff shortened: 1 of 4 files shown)


markwhitfeld (approved):
I'm happy with this solution! Nicely done.
Just a question, does the extra tick happen once for the app, or once for every new subscription, or and extra time for every emitted change?

arturovt:
> Just a question, does the extra tick happen once for the app, or once for every new subscription, or an extra time for every emitted change?

I'm not really sure how this is works in unit tests since ticks seem to be called illegibly there. I've just tested it in real app with patching `tick` through:
```js
const tick = ApplicationRef.prototype.tick;
let ticksCalled = 0;
ApplicationRef.prototype.tick = function () {
  console.log(`Ticks: ${ticksCalled++}`);
  return tick.call(this);
};
```
Ticks are called in a normal way per each event (like clicks, timers, etc), but I don't see any difference in the amount of clicks for this change in the real app.

markwhitfeld:
I'm happy with that. Thank you!

Issue #1857, opened 2022-06-10
🐞[BUG]: State plugin not having state after ofActionComplete(UpdateState)


lurock:
### Affected Package

 The issue is caused by package @ngxs/storage-plugin

### Is this a regression?

 Yes, the previous version in which this bug was not present was: 3.7.3

### Description

  Selecting state using this.store.selectSnapshot method after ofActionCompleted(UpdateState) on a browser refresh when using the Storage plugin worked fine in version 3.7.3 but in 3.7.4 it stopped working.  The state loaded from the storage plugin is no longer available at that time.

## Environment

<pre><code>
Libs:
- @angular/core version: 14.0.0
- @ngxs/store version: 4.7.4

arturovt:
Hey @lurock . Thanks for the quick response on the release. Is there a possibility to provide a reproducible example?

arturovt:
@lurock could you try installing the `3.7.4-dev.master-310a613` version and give it a try? Just update all packages to the `3.7.4-dev.master-310a613` version as follows:
```json
"@ngxs/store": "3.7.4-dev.master-310a613",
"@ngxs/storage-plugin": "3.7.4-dev.master-310a613"
```

lurock:
@arturovt I tried it with the 3.7.4-dev.master-310a613 version and it still does not have the state restored when the ofActionComplete(UpdateState) subscription fires.   

I have setup a repository with an example of the issue with NGXS on version 3.7.4: https://github.com/lurock/ngxs-state-storage-restore

If you look at the code you will see I console.log out the state after UpdateState is completed on this line https://github.com/lurock/ngxs-state-storage-restore/blob/7c5b527fda7a8bd823c5219ce10d1830dbdee8e5/src/app/actions-listener.service.ts#L19

For contrast I have setup a working version in a separate branch.  The only difference is the version of NGXS is now on 3.73.
https://github.com/lurock/ngxs-state-storage-restore/tree/ngxs-working

Steps to reproduce:
git clone https://github.com/lurock/ngxs-state-storage-restore.git
yarn install
yarn start

Launch the app at http://localhost:4200
1. click the save button to change the state
2. look at the console output the state is not restored from the storage plugin
3. Ctrl-c to stop the app

To compare with working version:
git checkout ngxs-working
yarn install
yarn start

Launch the app at http://localhost:4200
1. click the save button to change the state
2. look at the console output the state is now restored from the storage plugin

markwhitfeld:
[...]

Issue #1146, opened 2019-07-05
🐞[BUG]: Cannot modify state in ngxsOnInit when using NgxsStoragePlugin and NgxsRouterPlugin


AlexCleggett:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[X] Bug report  
[ ] Performance issue
[ ] Feature request
[ ] Documentation issue or request
[ ] Support request => https://github.com/ngxs/store/blob/master/CONTRIBUTING.md
[ ] Other... Please describe:
</code></pre>

## Current behavior

Any changes to the state made in the ngxsOnInit function are being overridden when using the router plugin and the storage plugin. Changes are not being persisted from the ngxsOnOnit function and all the persisted states are being rehydrated when the UpdateState action is handled by the storage plugin.

## Expected behavior

The router plugin or any other lazy loaded module should not cause the other states to be rehydrated from the storage engine. The router state itself is not part of the persisted states, so there should be no rehydrating at all when the UpdateState action is handled. Changes to state using patchState or setState in the ngxsOnInit function should persist.

## Minimal reproduction of the problem with instructions

https://stackblitz.com/edit/ngxs-simple-7pydiq

The ngxsOnInit function is resetting the count to -1.
1. Hit the 'click me' button several times to increase the count
1. Refresh the page.
1. Notice that the count is the same as before the page was refreshed and is not -1 like expected.
1. Try this again, but this time comment out the NgxsRouterModule import in the app module.

Notice that the count will now be -1 as expected, however the value saved in local storage does not reflect this.

## What is the motivation / use case for changing the behavior?

Cannot modify state in ngxsOnInit when using the router plugin and the storage plugin.

## Environment

<pre><code>
Libs:
- @angular/core version: 3.4.3
- @ngxs/store version: 3.4.3
- @ngxs/router-plugin version: 3.4.3
- @ngxs/storage-plugin version: 3.4.3

Browser:
- [X] Chrome (desktop) version 75
- [ ] Chrome (Android) version XX
- [ ] Chrome (iOS) version XX
- [ ] Firefox version XX
- [ ] Safari (desktop) version XX
- [ ] Safari (iOS) version XX
- [ ] IE version XX
- [ ] Edge version XX
 
For Tooling issues:
- Node version: XX  
- Platform:  

Others:

</code></pre>

## Notes

[...]

PR #1887, opened 2022-07-30
fix(storage-plugin): ensure the deserialization is not skipped for master key

arturovt:
Issues: #1857 and #1146

**1857**: This is a follow-up fix for https://github.com/ngxs/store/commit/2f9542d9f44d6b18d463ebcda5decf3c5b97ae18. We added a check if we need to re-hydrate the state again or not (considering `action.addedStates`). But this should never happen for the default state key (`@@STATE`).

**1146**: This is an old issue that also related to unnecessary state re-hydration when the default state key is used. We re-hydrate the whole state when any feature state is added by doing `state = { ...state, ...val }`. My fix adds some kind of "smart merge" that checks what feature state has been added and picks its name from `val`.


markwhitfeld (approved):
Great work!!!

Issue #1822, opened 2022-02-04
🐞[BUG]: Infinite Action Dispatching Cycle


adityaingle2:
### Affected Package

@ngxs/form-plugin: 3.7.3

### Description

When using the ngxs/form-plugin, some or multiple conditions cause the application to dispatch the "Update Form Value" , "Update Form Dirty", and "Update Form Errors" in an infinite loop, ultimately causing the application to crash. 

## 🔬 Minimal Reproduction

The simplest way to repro this issue is to have two forms on the same component bound to the same "ngxsForm" directive. (Essentially, we can follow the example for the ngxs forms plugin on the official website, and simply create a second copy of the form in the component). The snippets below shows a simpler example of this:

the code in app.component.html:
![appcomponenthtml](https://user-images.githubusercontent.com/13954083/152598483-275e35a8-c8f4-4ea2-aa71-c81bea85288f.png)

the code in app.component.ts:
![appcomponentts](https://user-images.githubusercontent.com/13954083/152598668-974255aa-9764-4661-8cc4-96cad9504fe7.png)

## 🔥 Exception or Error

After simply launching the application, the following can be observed when looking at the ReduxDevTools (the number of actions dispatched does not stop at the number in the red box - it continues endlessly until the application crashes): 

![ngxsFormLoop](https://user-images.githubusercontent.com/13954083/152596713-b6898083-985c-44dd-b1a2-a23be4848b7f.png)

## Environment

We ran into the same behavior (i.e. the infinite action dispatching loop) in our production app, which is more complex; however, the circumstance in which it occurred there is not identical to this (we did not have two distinct forms bound to the same ngxsForm directive); however, the example above was the simplest way to reproduce that behavior. 

With that, it's difficult to pinpoint if there are multiple conditions that can cause this infinite action dispatching loop, or if all instances of this behavior can be attributed to a single common condition. 

Libs:
- @angular/core version: 13.2.0
- @ngxs/store version: 3.7.3

Browser:
- [ ] Chrome (desktop) version 97

Others:
@ngxs/devtools-plugin: 3.7.3

markwhitfeld:
@adityaingle2 Thanks for the report. It was never the intention of this plugin that the same state would be bound twice, but, given your example, there could be an improvement to guard against the infinite loop.
The issue there is that Reactive Forms always returns a new object on any update, even if the property values haven't changed. This causes another update from the state because the objects aren't `===` (even though they have equal props).

[...]

PR #1890, opened 2022-07-31
fix(form-plugin): prevent actions infinite loop with multiple `ngxsForm` directives

arturovt:
Issue: #1822 

I'm unsure if there's another way to fix this issue because `valueChanges` continuously emits a new object (with a new reference), which means just adding `distinctUntilChanged()` isn't enough.

I've been thinking of potential breaking changes with this change. Since binding 2 `ngxsForm` is impossible for now, we can imagine that people are binding `ngxsForm` only once. So if the form value is updated outside with the same value and we're not dispatching any actions - this probably might not break anything.


markwhitfeld (approved):
Fantastic!!!

PR #1892, opened 2022-08-02
docs(store): add "Typing Operators" section to `operators.md`

### LINKED ISSUE #1799: 📚[DOCS]: State Operator docs only covers basic usage
### Description

State Operators are quite useful, but the examples provided in the docs only cover the most basic of usages: updating a primitive value at the root of the state.  Most examples also lack all or most template type specifiers.  It would be quite valuable for the State Operator page to include examples on how to update an inner property in a nested structure.  I also find myself constantly tempted to add `//@ts-ignore` comments when I use state operators, because they want a template type specified, and it's not always clear to an inexperienced user (like myself) what types I should specify.  Including those template type specifiers in the examples would at least provide... well, examples.

arturovt:
Closes #1799

### DIFF
--- a/docs/advanced/operators.md
+++ b/docs/advanced/operators.md
@@ -182,7 +182,7 @@ You will see that in each case above the state operators are wrapped within a ca
 
 ## Typing Operators
 
-Always specifying types for the `patch` operator when doing nested updates is necessary. You can face cases when the `patch` operator cannot infer the nested structure type. Let's look at the following state:
+Specifying types for the `patch` operator is always necessary when doing nested updates. You can face cases when the `patch` operator cannot infer the nested type structure. Let's look at the following state:
 
 ```ts
 export class UpdateLine1 {
@@ -231,7 +231,7 @@ export class AddressState {
 }
 ```
 
-If we don't specify the `patch` type explicitly, all objects are inferred as `unknown`, meaning the TypeScript cannot tell us that we're doing something wrong or using the wrong type. The correct way of specifying nested types looks as follows:
+If we don't specify the type explicitly for `patch`, all objects are inferred as `unknown`, meaning that TypeScript cannot tell us that we're doing something wrong or using the wrong type. The correct way of specifying nested types is shown below:
 
 ```ts
 export class UserState {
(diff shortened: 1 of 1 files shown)


docs/advanced/operators.md
```diff
@@ -147,42 +147,120 @@ export class ChangePandaName {
...
     );
   }
+}
 ```
 
 You will see that in each case above the state operators are wrapped within a call to the `patch` operator. This is only done because of the convenience that the `patch` state operator provides for targeting a nested property of the state.
 
+## Typing Operators
+
+Always specifying types for the `patch` operator when doing nested updates is necessary. You can face cases when the `patch` operator cannot infer the nested structure type. Let's look at the following state:
```
Carniatto, line 185:
```suggestion
Specifying types for the `patch` operator is always necessary when doing nested updates. You can face cases when the `patch` operator cannot infer the nested structure type. Let's look at the following state:
```

docs/advanced/operators.md
```diff
@@ -147,42 +147,120 @@ export class ChangePandaName {
...
+            })
+          })
+        })
+      })
+    );
+  }
+}
+```
+
+If we don't specify the `patch` type explicitly, all objects are inferred as `unknown`, meaning the TypeScript cannot tell us that we're doing something wrong or using the wrong type. The correct way of specifying nested types looks as follows:
```
Carniatto, line 234:
```suggestion
If we don't specify the type explicitly for `patch`, all objects are inferred as `unknown`, meaning that TypeScript cannot tell us that we're doing something wrong or using the wrong type. The correct way of specifying nested types is shown below:
```

docs/advanced/operators.md
```diff
@@ -147,42 +147,120 @@ export class ChangePandaName {
...
+            })
+          })
+        })
+      })
+    );
+  }
+}
+```
+
+If we change `country` to `Qcountry` (intentional mistake), the compiler will tell us `Object literal may only specify known properties, but 'Qcountry' does not exist`. The same technique may be used with other operators if they cannot infer the type, specify the state model and chain properties to get the desired type.
```
Carniatto, line 255:
```suggestion
If we change `country` to `Qcountry` (intentional mistake), the compiler will tell us `Object literal may only specify known properties, but 'Qcountry' does not exist`. The same technique may be used with other operators if they cannot infer the type. 

💡 Tip: we can specify the state model type and chain properties to get the desired type. Like the example above.
```

Carniatto (approved):
LGTM

markwhitfeld (approved):
Great addition!
PS. In the improvements to state operator typing that I am working on for v3.8 the explicit types will not be needed anymore! I have found a way to infer the parameter types from the contextually required return type 🎉

Issue #1898, opened 2022-08-04
🐞[BUG]: Router Plugin overrides inner navigation with redirect url


anschm:
### Affected Package
This issue is caused by package ngxs/router-plugin.

### Description

In my case we are using auth0 as OIDC management system. If a blocked user would like to login auth0 redirects to the callback url with the following query parameters ?error=unauthorized&error_description=user is blocked and so on. This redirect is handled from the auth0 angular library and can be observed. We are catching this error in our application and would like to navigate to a specific page which shows the user that he is blocked. But this does not work, because the router plugin canceled our inner navigation to triggers a new navigation to the redirect url which was performed from the auth0 server. This is done by the following code:

const subscription = this._router.events
      .pipe(first((event): event is RoutesRecognized => event instanceof RoutesRecognized))
      .subscribe(({ url }) => {
        // `location.pathname` always equals manually entered URL in the address bar
        // e.g. `location.pathname === '/foo'`, but the `router` state has been initialized
        // with another URL (e.g. used in combination with `NgxsStoragePlugin`), thus the
        // `RouterNavigation` action will be dispatched and the user will be redirected to the
        // previously saved URL. We want to prevent such behavior, so we perform this check
        // `url` is a recognized URL by the Angular's router, while `currentUrl` is an actual URL
        // entered in the browser's address bar
        // `PathLocationStrategy.prototype.path()` returns a concatenation of
        // `PlatformLocation.pathname` and normalized `PlatformLocation.search`.
        // `Location.prototype.normalize` strips base href from the URL,
        // if `baseHref` (declared in angular.json) for example is `/en`
        // and the URL is `/test#anchor` - then `_locationStrategy.path(true)` will return `/en/test#anchor`,
        // but `/en/test#anchor` is not known to the Angular's router, so we have to strip `/en`
        // from the URL
        const currentUrl = this._location.normalize(this._locationStrategy.path(true));
        const currentUrlTree = this._urlSerializer.parse(currentUrl);
        // We need to serialize the URL because in that example `/test/?redirect=https://google.com/`
        // Angular will recognize it as `/test?redirect=https:%2F%[2Fwww.google.com](http://2fwww.google.com/)%2F`
        // so we have to run the `currentUrl` via the `UrlSerializer` that will encode characters
        const currentSerializedUrl = this._urlSerializer.serialize(currentUrlTree);
        // If URLs differ from each other - we've got to perform a redirect to the manually entered URL
        // in the address bar, as it must have a priority
[...]

Issue #1204, opened 2019-08-02
🐞[BUG]: Router Plugin + Storage Plugin results in initial route being redirected incorrectly


dallastjames:
## I'm submitting a...
<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[x] Bug report  
[ ] Performance issue
[ ] Feature request
[ ] Documentation issue or request
[ ] Support request => https://github.com/ngxs/store/blob/master/CONTRIBUTING.md
[ ] Other... Please describe:
</code></pre>

## Current behavior

When using both the ngxs/router-plugin and the ngxs/storage-plugin together in a project, the initial routing to a child component is canceled and a redirect to the blank state is triggered. See the following stackblitz for a reproduction: https://angular-tdus7x.stackblitz.io/

Navigate to either the page one or page two route, and refresh the page. On refresh, the app redirects back to the initial empty route. 

## Expected behavior

When refreshing a page while both the ngxs/router-plugin and ngxs/storage-plugin are installed, the application remains on the current page and does not redirect to the empty route.

## Minimal reproduction of the problem with instructions
https://angular-tdus7x.stackblitz.io/

https://github.com/dallastjames/ngxs-routerstoragebug

Run the example provided. When navigated to either of the child pages a page refresh will redirect back to the empty route. Removing either the storage-plugin or the router-plugin fixes the issue, but as soon as both are added back, the problem returns.

As far as I've been able to tell the issue has to do with the initial NavigationStart event not being triggered when the app is not on the blank route. Logging the events that come from the router, There is a NavigationStart event triggered when the application is on the blank route. This always occurs, regardless of what plugins are installed. When navigating to child pages and refreshing, if both plugins are installed, this event is not seen and the app is redirected back to the blank route. If either of the plugins is removed, the NavigationStart event can then be seen when refreshing the page on a child route. 

I added some logging in the router.state.ts file around the `checkInitialNavigationOnce()` method. The url received from the router events is always blank when the router plugin is installed with the storage plugin. Without the storage plugin, the value received here is correct and the method works correctly.

<details>
<summary>Router Events when only the router-plugin is installed:</summary>

![image](https://user-images.githubusercontent.com/16655763/62398318-99382b00-b535-11e9-83fb-bbae8b24a765.png)

</details>

<details>
<summary>Router Events when both the router-plugin and storage-plugin are installed:</summary>

[...]

PR #1907, opened 2022-08-07
fix(router-plugin): prevent router overriding valid navigation

arturovt:
Issue: #1898

We initially added this check when trying to fix router + storage plugins integration.

We had an open issue related to this (https://github.com/ngxs/store/issues/1204), but I've closed it since I verified the issue is not reproducible anymore with the latest NGXS version. The issue was reproducible with the `3.5.0` version (which was used in the reproducible example). I have upgraded NGXS and removed the `checkInitialNavigationOnce` method, and the behavior was the same as expected (the issue still wasn't reproducible).


markwhitfeld (approved):
Great! I love the smell of deleted code in the morning!
Bundlemon is your biggest fan.

markwhitfeld:
@arturovt Could you update the change log?

Issue #1728, opened 2021-03-12
🐞[BUG]: State override in case of concurrent dispatch


krszktc:
### Affected Package
 @ngxs

### Description

There are three components. Two of them show data aggregated in specific, different way. Third showing potential issue with data. As for now I have two issues with data:

1. Both ComponentA and ComponentB dispatching data but ValidationComponent achieve only one message from store. I'm aware that log could be slower than action execution on store so this is 'optional issue'. Some inconvenience just to mention in margin :)

![image](https://user-images.githubusercontent.com/24602491/111193139-22d92700-85ba-11eb-845c-43e009dfc129.png)

2. This is something what makes me more problem with my data consistency. Looking at the logs above, before any interaction, status for StatusState is **{id: "id-2", status: "valid"}**. And now when you click "BAD" button which simulate data with validation issue, app status looks like on screen below. 

Doesn't matter if I'm using async/await, pipe(tap()) or just subscribe to service in DataState, all the time I'm getting one extra emission in ValidationComponent which looks to be some kind of snapshot of data, before dispatch from componentsA/B. Log can be broken - ok, but at the end data in components are inconsistent.

I found some temporary workaround where if I dispatch one extra action from DataState it's working as I expect ... but being hones I have not idea why.

![image](https://user-images.githubusercontent.com/24602491/111196229-55385380-85bd-11eb-987c-3acbadad4266.png)

## 🔬 Minimal Reproduction

https://stackblitz.com/edit/angular-ivy-ph8qwv
   

## Environment
<pre><code>
Libs:
    "@ngxs/devtools-plugin": "^3.6.2",
    "@ngxs/logger-plugin": "^3.6.2",
    "@ngxs/store": "^3.6.2",

Browser:
- [x] Chrome (desktop) version  88.0.4324.190
- [ ] Chrome (Android) version XX
- [ ] Chrome (iOS) version XX
- [ ] Firefox version XX
- [ ] Safari (desktop) version XX
- [ ] Safari (iOS) version XX
- [ ] IE version XX
- [ ] Edge version XX
 
For Tooling issues:
- Node version: 14.7.0
- Platform:  Linux, Windows

</code></pre>

VitaliyFilatov:
trick with setTimeout with 0 delay is help to solve the problem

VitaliyFilatov:
it reproduce with one component https://stackblitz.com/edit/angular-ivy-otxvqt?file=src/app/one.component.ts

[...]

PR #1908, opened 2022-08-07
fix(store): avoid incorrectly ordered state observable events

arturovt:
Issue: #1728 

<s>Note: `do not emit old values from the state` is a wrong commit name, I just haven't found anything meaningful, this should be renamed.</s>

I've updated `tsconfig.spec.json` to stop warning on unused locals, because it's a pain writing unit tests with these options enabled.


markwhitfeld (approved):
Great fix!!!!!

Issue #1567, opened 2020-04-16
🐞[BUG]: ngxsOnChanges is not being called when ngxs form state changes


timonmasberg:
### Affected Package

 The issue is related to the packages @ngxs/store and @ngxs/form-plugin

### Description

The ngxsOnChanges implemented with this [PR ](https://github.com/ngxs/store/pull/1389) is not being called when the Angular Reactive Form bound to the form state changes.
As the documentation for ngxsOnChanges suggests, subscribing to the store can be replaced with the ngxsOnChanges method. Sadly this is not the case for ngxsForms.

## 🔬 Minimal Reproduction
Please check the form.state.ts 
In the console you are able to see only the subscription on the state works as expected, the ngxsOnChanges will only be called once.
https://stackblitz.com/edit/angular-nczdzz

## Environment

<pre><code>
Libs:
- @angular/core version: 9.1.0
- @ngxs/store version: 3.6.2
- @ngxs/form-plugin version: 3.6.2

Browser:
- [ x] Chrome (desktop) version XX
- [ ] Chrome (Android) version XX
- [ ] Chrome (iOS) version XX
- [ ] Firefox version XX
- [ ] Safari (desktop) version XX
- [ ] Safari (iOS) version XX
- [ ] IE version XX
- [ ] Edge version XX

</code></pre>

splincode:
Good catch

LeParadoxHD:
Any change on this issue?

irowbin:
Any news of this issue?

splincode:
@LeParadoxHD @irowbin unfortunately, we do not have enough time to solve some problems, could you try to investigate this and send PR?

arturovt:
This has been fixed as part of the [v3.7.6](https://github.com/ngxs/store/releases/tag/v3.7.6) release.
Please test this out and let us know if this is not resolved.

PR #1926, opened 2022-10-19
fix(store): call `ngxsOnChanges` whenever state changes (even through plugins)

arturovt:
Issue: #1567

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,8 +12,9 @@
 $ npm install @ngxs/store@dev
 ```
 
-- Fix: Storage Plugin - Do not skip deserialization for keys with dot notation [#1924](https://github.com/ngxs/store/pull/1924)
 - Fix: Use `isObservable` to test whether actions return an observable [#1925](https://github.com/ngxs/store/pull/1925)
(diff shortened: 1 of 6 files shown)


markwhitfeld (commented):
I like this approach. It is much cleaner, and keeps this logic quite isolated to one place.
A subtle side effect of this change would be that previously the `ngxsOnChanges` would have been called before the state is actually set on the root `BehaviorSubject`, and now it is called in response to the change in the root behavior subject.
I think that this is ok, and perhaps more logical if the user was to ever get a snapshot of the state from within this function.

I also see that due to the fact that this is literally the very first subscription that connects to a state as it is initialised it will in the **majority** of cases still honour the contract of being the first thing that is called. After this change, the only situation where another selector for that state would be called before this function is if the state is lazy loaded as part of a feature and a selector is set up before the state is loaded and initialised. While this is a subtle side effect of the change I don't think that it has enough impact to merit concern. You should really not be using the `ngxsOnChanges` hook for major logic and especially not logic that would change the same feature state.

markwhitfeld (approved):
Excellent work!!!

Issue #1691, opened 2020-11-06
🚀[FEATURE]: Make choose to forward or not the errors to the global handler


XavierDupessey:
I'm submitting a...

<pre><code>[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[ ] Performance issue
[x] Feature request
[ ] Documentation issue or request
[ ] Support request => https://github.com/ngxs/store/blob/master/CONTRIBUTING.md
[ ] Other... Please describe:
</code></pre>

## Current behavior

If we don't want to catch the errors directly in the actions to handle them with _ofActionErrored_, they get actually handled twice: once by _ofActionErrored_ and once by the global error handler.

## Expected behavior

I think we should be able to decide if NGXS should forward the errors to the global error handler or not. Another possibility could be to wrap them in a custom error object (e.g. _UnhandledNgxsActionError_) to be able to filter them in the handler.

## Minimal reproduction of the problem with instructions

* Go to https://stackblitz.com/edit/ngxs-error-handling-c3brj2?file=app%2Fapp.component.ts
* Click on "Load data": 2 alerts are shown: one from _ofActionErrored_ in _app.component.ts_ and another one from _handleError_ in _global-error-handler.ts_

## Environment

<pre><code>Libs:
- @angular/core version: 9.1.12
- @ngxs/store version: 3.7.0
</code></pre>

markwhitfeld:
@XavierDupessey Thank you for submitting this issue.
Apart from the proposal of having the _UnhandledNgxsActionError_ error type to assist with filtering, do you have any thoughts of how you would like to specify that a particular action should not be handled by the global error handler?

XavierDupessey:
For a particular action, could it be part of the _ActionOptions_ object?
```
@Action(Submit, { throwErrors: true })
submit(...)
{
  // Won't be forwarded to GlobalErrorHandler because this action was expected to throw errors
  return throwError(new Error('API is down'));
}
```

But I would prefer an option to set it globally (with _NgxsConfig_ maybe?). I think it would be annoying to put `{ throwErrors: true }` everywhere if we are used to handle errors with `ofActionErrored`, what do you think?

markwhitfeld:
One option we could have a look at is to detect when an action is being monitored by an `ofActionErrored` pipe and then deem that action as handled and not forward to the global error handler in that case.
But this may be too much magic, and not make for an obvious usage pattern.

kctang:
[...]

PR #1927, opened 2022-11-01
fix(store): do not delegate errors to `ErrorHandler` if users catch them manually

arturovt:
Issue: #1691

### DIFF
--- a/packages/store/src/internal/error-handler.ts
+++ b/packages/store/src/internal/error-handler.ts
@@ -1,0 +1,69 @@
+import { ErrorHandler, Injectable, Injector } from '@angular/core';
+import { Observable } from 'rxjs';
+
+import { leaveNgxs } from '../operators/leave-ngxs';
+import { NgxsExecutionStrategy } from '../execution/symbols';
+
+/**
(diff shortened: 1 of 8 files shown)


packages/store/src/internal/error-handler.ts
```diff
@@ -0,0 +1,69 @@
...
+            }
+          })
+        );
+      }
+    });
+
+    return new Observable(subscriber => {
+      subscribed = true;
+      return source.pipe(leaveNgxs(ngxsExecutionStrategy)).subscribe(subscriber);
+    });
```
markwhitfeld, line 46: This is a bit different to what we came up with when we paired on this.
Won't this result in the scenario where, if you `subscribe` to the dispatch but don't handle the error, then the error would not be handled globally (because the error handler on line 27 would prevent global propagation and the guard on line 33 would prevent the error handler from being hit)?
We would need to unsubscribe to line 26 when a subscriber arrives so that we can allow for the normal rxjs strategy to be used when errors are not handled in a subscription.

markwhitfeld (approved):

PR #1947, opened 2022-12-03
feat(operators): improve type checking and intellisense for State Operators

markwhitfeld:
## PR Checklist

## PR Type

```
[x] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

There have been some long-running issues with the way that the typing of state operators work.
This stemmed from some changes in Typescript 3.x in the way that it handled co-variance and contra-variance.
As a result of this change in typescript, state operators would infer their types from the parameters and the return type would satisfy the expected type only due to type structure similarities.
The problem is that this resulted in the fact that additional properties (potentially misspelt properties) would be accepted and would not show an error. Also, if the user tried to use IntelliSense within a patch operator, it would not show any useful suggestions.

## What is the new behavior?

The type of a state operator is not inferred from the contextual requirement of the State Operator on it's return type.
The typescript default of inferring the type of T from the arguments of the operator has been blocked so that it allows for this "reverse" inference. As a result of this change, both type checking and IntelliSense work as originally intended.

```
[ ] Yes
[x] No
```

After upgrading, the developer may get some compile errors, but this is indicating areas in the code that are incorrectly typed or have an unsafe typing context. Some users may see this as a "breaking change", but it is an enhancement to the typing of the lib instead.


markwhitfeld:
Marking this as a draft PR because I still need to decide how some new types are exposed from the lib.

packages/store/src/public_api.ts
```diff
@@ -33,6 +33,9 @@ export {
 export {
   StateContext,
   StateOperator,
+  ExistingState,
```
arturovt, line 36: I think this should be exported with a barred O indicating it's a private stuff.

arturovt: This still can be used by library consumers who work on custom operators in their apps. But still should be exported as private symbols to indicate that the type might be ch aged between versions and use it on their risk.

markwhitfeld: I do consider this type as part of the public api, because people should be able to create their own operators. 
I'm thinking that there needs to be a operator types specific subpackage though

arturovt: Well, I consider this to be a sub package so it’s not tied to the core. Also that would remove a reference from operators of the core package.

arturovt: Umm, well… lemme get to PC and check again

markwhitfeld: It could just be a types import into the core package from the operators subpackage

arturovt: I'm ok with leaving it in the core

arturovt (approved):

packages/store/src/symbols.ts
```diff
@@ -79,7 +79,24 @@ export class NgxsConfig {
   }
 }
 
-export type StateOperator<T> = (existing: Readonly<T>) => T;
+type _NoInfer<T> = T extends infer S ? S : never;
+export type NoInfer<T> = T extends (infer O)[] ? _NoInfer<O>[] : _NoInfer<T>;
```
arturovt, line 83: Since we're exposing it publicly could you leave meaningful comments?

markwhitfeld: @arturovt I have added the doc comments now.

Issue #1965, opened 2023-01-19
🐞[BUG]: Angular ErrorHandler not called anymore (regression)


jbjhjm:
### Affected Package

 @ngxs/store

### Is this a regression?

Yes, used to work in 3.7.4
(likely introduced in 3.7.6)
97% sure its caused by this change: https://github.com/ngxs/store/pull/1927/files

### Description

Errors thrown in Action handlers are not propagated to Angular ErrorHandler anymore.
This is breaking several of our unit tests and also hides errors that else would be propagated to ng ErrorHandler, followed by our Error Management system connecting to that.

With https://github.com/ngxs/store/pull/1927 I found what is likely the source of that change.
As markwhitfield annotated here (without any answer to that unfortunately):
https://github.com/ngxs/store/pull/1927/files#r1018904814
this does change behavior:
As soon as subscribing in any way, ngxs disables error logging.
Even if the only subscriber listens for e.g. completion.

The original feature request asked for a solution to stop doubled error logging. https://github.com/ngxs/store/issues/1691
This solution unfortunately makes it 0 error logging for us and likely others :/

Also it is a breaking change which should be documented and not released in a minor version.
I get why the new approach is better for some users...
but many will go blind, not knowing about this change, implementing stuff, not being notified about thrown errors anymore.
I only learned about this because we had some unit tests based on that exact behavior.
Which made me start searching what has changed and then I found out about this PR.

As a compromise, I ask to make this configurable by a root config flag like it was done for `suppressErrors`.
By adding a flag `disableErrorPropagationWhenSubscribed` or similar, this can be an opt-in feature,
not breaking any behavior for existing implementations.
And it allows any devs plagued by doubled error logging to switch this feature on at any time.

## 🔬 Minimal Reproduction

here's a simple reproduction with a jest test.

```ts
import { ErrorHandler, Injectable } from '@angular/core';
import { TestBed } from '@angular/core/testing';
import { Action, NgxsModule, State, Store } from '@ngxs/store';

// eslint-disable-next-line @typescript-eslint/no-empty-interface
interface TestModel {}

class ThrowingAction {
	static readonly type = '[ErrorHandlingIssueState] Throw an Error';
}

@Injectable()
@State<TestModel>({
	name:'errorhandlingreproduction',
	defaults:{}
})
export class ErrorHandlingIssueState {
	@Action(ThrowingAction)
	errorThrowingHandler() {
		throw new Error('Cannot handle something!')
	}
}


[...]

PR #1971, opened 2023-03-01
fix(store): add setting that allows disabling dual error handling

arturovt:
Issue: #1965 

NGXS has had an explicit error-handling mechanism from actions for the last years. However, this caused issues with `ErrorHandler` being called twice with the same error. The explicit error handling mechanism is required because of the NGXS action handling strategy. The default action handling process leaves the Angular zone, so NGXS actions are invoked in the `<root>` zone context. All caught actions must be returned to the Angular error handler.

Angular's default error-handling mechanism requires zone.js to be bundled. The forked Angular zone has an `onHandleError` hook which calls `zone.onError.emit(error)` when any error is being caught in Angular zone. Angular subscribes to `onError` when calls `bootstrapModuleFactory`:
```js
const exceptionHandler = moduleRef.injector.get(ErrorHandler, null);

...

const subscription = ngZone.onError.subscribe({
    next: (error) => {
        exceptionHandler.handleError(error);
    }
});
```

This PR adds "switchers" between explicit and implicit error-handling mechanisms. We either catch them manually and delegate them to `handleError` or allow Angular to catch errors in its zone and delegate them to `handleError`.

We would use the old behavior where we catch errors explicitly to be backward-compatible and allow users to switch to the implicit behavior where zone.js is responsible for catching errors and delegating them to Angular.

### DIFF
--- a/docs/advanced/errors.md
+++ b/docs/advanced/errors.md
@@ -9,9 +9,6 @@ import { NgModule, ErrorHandler } from '@angular/core';
 export class MyErrorHandler implements ErrorHandler {
   handleError(error: any) {
     console.log('ERROR! ', error);
-
-    // Make sure to rethrow the error so Angular can pick it up
-    throw error;
   }
 }
 
@@ -27,12 +24,14 @@ export class MyErrorHandler implements ErrorHandler {
 export class AppModule {}
 ```
 
-## Handling errors within an `@Select`
+## Handling Errors Within a `select`
 
 ```ts
 @Component({ ... })
 class AppComponent {
-  @Select(state => state.count.number.value) count$: Observable<number>;
+  count$: Observable<number> = this.store.select(state => state.count.number.value);
+
+  constructor(private store: Store) {}
 }
 ```
 
@@ -76,7 +75,7 @@ This option allows to track errors and handle them.
 ```ts
 @Component({ ... })
 class AppComponent {
-  @Select(state => {
+  count$: Observable<number> = this.store.select(state => {
     try {
       return state.count.number.value;
     } catch (error) {
@@ -84,8 +83,9 @@ class AppComponent {
       // throw error;
       // Automatic unsubscription will occur if you use the `throw` statement here. Skip it if you don't want the stream to be completed on error.
     }
-  })
-  count$: Observable<number>;
+  });
+
+  constructor(private store: Store) {}
 }
 ```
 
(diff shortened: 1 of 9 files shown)


docs/advanced/errors.md
```diff
@@ -76,39 +75,54 @@ This option allows to track errors and handle them.
...
       // error is thrown
     } catch (err) {
-      console.log('error catched inside @Action wont propagate to ErrorHandler or dispatch subscription')
+      console.log('error catched inside @Action will not propagate to ErrorHandler or dispatch subscription')
     }
   }
 ```
 
-## Handling errors after dispatching an action
+You can return an observable that completes immediately after error has been handled when dealing with streams:
```
Carniatto, line 111:
```suggestion
You can return an observable that completes immediately after the error has been handled when dealing with streams:
```

markwhitfeld (commented):
I think that this is a great addition, but I'm unsure about the name.
What other names have you thought about?

arturovt:
```ts
NgxsModule.forRoot([], { useNgxsErrorHandlingStrategy: false })

NgxsModule.forRoot([], { useAngularErrorHandlingStrategy: true })

NgxsModule.forRoot([], { delegateAllErrorsToErrorHandler: true })
```

markwhitfeld:
@arturovt could you describe the two different modes?
The PR description gives a great overview and it describes the two modes as explicit and implicit error handling but the exact difference is unclear. 

Could you summarize each mode with details of behavior under each mode?

This will go a long way to help with naming.
I'll give my own attempt at this summary in the next comment.

markwhitfeld:
Upon reviewing this, is suspect that there may actually be a fix that allows this to work as expected in all cases.
I think that there is a bug as I suspected in this comment: https://github.com/ngxs/store/pull/1927/files#r1018904814
The other bug as mentioned here (where the error is reported async):
https://github.com/ngxs/store/issues/1965#issuecomment-1396913098

could be to do with the `Promise.resolve().then(...` found here:
https://github.com/ngxs/store/pull/1927/files#diff-d211bce5ab62868a141a2954aeb1c7a4f3d3cf03dd08a8ef4f1c8496e1548f0dR32

Do you remember the motivation for this new microtask?

arturovt:
The microtask is required because the dispatch observable may be synchronous which means the `error` callback is invoked before the `subscribed` is set to `true`.

If `useNgxsErrorHandlingStrategy` is `false`:
```
dispatch action
             |
action handler throws error
             |
     does user handle error through subscribe({ error: ... }) ?
     /                                          \
    yes                                          no
    |                                            |
propagate error to error callback    call to ErrorHandler.handleError
```

If `useNgxsErrorHandlingStrategy` is `true`:
```
dispatch action
             |
action handler throws error
             |
     does user handle error through subscribe({ error: ... }) ?
     /                                          \
    yes                                          no
    |                                            |
propagate error to error callback    call to ErrorHandler.handleError
+ call to ErrorHandler.handleError
```

markwhitfeld:
I had a discussion with @arturovt and we closed this PR in favour of another approach that will be taken in an upcoming PR.

PR #1972, opened 2023-03-02
feat(form-plugin): allow `ngxsFormDebounce` to be string

arturovt:
The `ngxsFormDebounce` is a static binding for now which means we're not watching its changes. We read it only once and provide to `debounceTime`. The change allows providing `ngxsFormDebounce` as a string, and not as a number so we extend possible values to be provided:
```html
<form [ngxsFormDebounce]="300">...</form>
<form ngxsFormDebounce="300">...</form>
```


markwhitfeld:
@arturovt is there a related issue for this one?

markwhitfeld (approved):

markwhitfeld (changes requested):
Failing build

markwhitfeld (commented):
LGTM

PR #1983, opened 2023-03-28
build: upgrade to Angular 15

arturovt:


jest.config.js
```diff
@@ -105,7 +123,12 @@ module.exports = {
    * project's root directory to prevent it from accidentally ignoring all of
    * your files in different environments that may have different root directories.
    */
-  testPathIgnorePatterns: ['/node_modules/', '/types/', '/helpers/'],
+  testPathIgnorePatterns: [
+    '/node_modules/',
+    '/types/',
+    '/helpers/',
+    path.join(__dirname, 'packages/hmr-plugin')
```
markwhitfeld, line 130: Why is this now ignored?

packages/devtools-plugin/tests/utils/redux-connector.ts
```diff
@@ -11,8 +11,11 @@ export class ReduxDevtoolsMockConnector {
   private dispatcher: Subject<NgxsDevtoolsAction> = new Subject<NgxsDevtoolsAction>();
   private countId = 0;
 
-  public subscribe(fn: Function): Subscription {
-    return this.dispatcher.subscribe(e => fn(e));
+  public subscribe(fn: Function): VoidFunction {
+    const subscription = this.dispatcher.subscribe(e => fn(e));
+    return () => {
+      return subscription.unsubscribe();
+    };
```
markwhitfeld, line 18: What is the reason for this extra code and slight signature change?

packages/hmr-plugin/src/hmr-bootstrap.ts
```diff
@@ -29,7 +29,7 @@ export async function hmr<T>(
   const dataTransfer: HmrDataTransfer = webpackModule.hot.data || {};
 
   const storage = new HmrStorage<any>(dataTransfer.snapshot || {});
-  const manager = new HmrManager<T>(options, storage);
+  const manager = new HmrManager<any>(options, storage);
```
markwhitfeld, line 32: Why have we changed to `any` here?

tsconfig.build.json
```diff
@@ -2,8 +2,8 @@
   "extends": "./tsconfig.json",
   "compilerOptions": {
     "outDir": "out-tsc/lib",
-    "target": "es2015",
-    "module": "es2015",
+    "target": "es2022",
+    "module": "es2022",
```
markwhitfeld, line 6: Warning, this is the build target that introduces the typescript bug.
This could cause issues with selectors, where the actual selector (not the value) comes through as `undefined`.
Just so that you are aware if anything strange comes up.

jest.config.js
arturovt: Tests are failing for this plugin and I have no idea and time to fix the plugin and tests :(

packages/devtools-plugin/tests/utils/redux-connector.ts
arturovt: This is the correct signature. The actual redux connector doesn't return a subscription but rather a `VoidFunction`.

packages/hmr-plugin/src/hmr-bootstrap.ts
arturovt: Failing.

jest.config.js
markwhitfeld: In that case we should maybe remove it from the published libs for the next version.
(we can keep the code in the repo and assess later what approach is necessary)

Issue #850, opened 2019-02-18
🚀[FEATURE]: @ngxs/storage-plugin forFeature


Carniatto:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[ ] Performance issue
[x] Feature request
[ ] Documentation issue or request
[ ] Support request => https://github.com/ngxs/store/blob/master/CONTRIBUTING.md
[ ] Other... Please describe:
</code></pre>

## Current behavior

only supports forRoot option

## Expected behavior

allow only part of the state to be persisted in localStorage

## What is the motivation / use case for changing the behavior?

In large applications or even in Nx Workspaces one's might find interesting to have the state for some lazy modules persisted in localStorage but not all the state

## Environment

<pre><code>
Libs:
- @angular/core version: 7.1.2
- @ngxs/store version: 3.3.4

Browser:
- [ ] Chrome (desktop) version XX
- [ ] Chrome (Android) version XX
- [ ] Chrome (iOS) version XX
- [ ] Firefox version XX
- [ ] Safari (desktop) version XX
- [ ] Safari (iOS) version XX
- [ ] IE version XX
- [ ] Edge version XX
 
For Tooling issues:
- Node version: v11.4.0
- Platform:  OSx Mojave

splincode:
Can in more detail

eranshmil:
You can use the `key` option.
https://ngxs.gitbook.io/ngxs/plugins/storage#options

Carniatto:
@eranshmil  I don't see this working, can you provide me an example?

@splincode My use case is when you have a root state and couple of  lazy loaded states but you only wanna save in storage one of the lazy loaded states.

drakenfly:
I thought about the same thing. What @Carniatto wanted to say with "allow only part of the state to be persisted in localStorage" is (at least it was like following in my opinion): Allow to add some keys with a forFeature dynamically, based on the lazy loaded modules.

For example, if you split your application like this:
- AppModule
|-- AdminModule (lazy module, canLoad and canActivate to prevent unauthorized access)
|-- CustomerModule (lazy module)

Then I'd personally don't like to configure the NgxsStorage-plugin for AdminModule when loading AppModule initially.

kxc0re:
it's also a performance gain since the filtering for keys happens after a storage-update has been initiated

https://stackblitz.com/edit/ngxs-repro-yffmsq
[...]

PR #1994, opened 2023-04-11
feat(storage-plugin): allow providing feature states

arturovt:
Issue: #850

This commit introduces the ability to specify states for persistence at the feature level.
Previously, you could only specify a list of states for persistence when providing the storage
plugin at the root level. This feature is only available with the standalone API, as Angular has
an `ENVIRONMENT_INITIALIZER` feature that allows us to add these states at the feature level.

The API is simple and straightforward. It introduces the `withStorageFeature` function, which
can be called along with `provideStates`. It requires a list of states to be provided. If all
states are already being persisted because the developer specified `keys: *` at the root level,
this won't do anything.


markwhitfeld:
Very interesting work going on here! 🎉 
Could you add a description to the PR (as well as a linked issue or discussion if applicable) so that the problem, and proposed solution are clear?

arturovt:
> Very interesting work going on here! tada Could you add a description to the PR (as well as a linked issue or discussion if applicable) so that the problem, and proposed solution are clear?

```js
atob('bm90IHJlYWR5IHlldA==')
```

:joy:

markwhitfeld (changes requested):
This is a really awesome feature!
We need a bit more information of what this feature looks like, so could you add a section in the docs about it, or describe it in the PR description, so that what the feature looks like is clear for review.
Is there a linked issue or discussion? This may assist in making the the problem, and proposed solution clearer.

packages/storage-plugin/tests/storage-for-feature.spec.ts
```diff
@@ -0,0 +1,94 @@
...
+  @Injectable()
+  class BlogState {}
+
+  @Component({ template: 'This is blog' })
+  class BlogComponent {}
+
+  @NgModule({
+    imports: [RouterModule.forChild([{ path: '', component: BlogComponent }])],
+    declarations: [BlogComponent],
+    providers: [provideStates([BlogState], withStorageFeature([BlogState]))]
```
markwhitfeld, line 40: I think that, in addition to this standalone API, we should have the `NgxsStoragePluginModule.forFeature([...])` API.

arturovt: We can't, because there's no `imports` feature on the `Route` level, similar to `providers`.

packages/storage-plugin/internals/src/storage-key.ts
```diff
@@ -12,7 +12,7 @@ export interface KeyWithExplicitEngine {
 
 /** Determines whether the provided key has the following structure. */
 export function ɵisKeyWithExplicitEngine(key: any): key is KeyWithExplicitEngine {
-  return key != null && !!key.engine;
+  return !!key?.engine;
```
markwhitfeld, line 15: Although very succinct, it is difficult to understand this. Adding brackets would help with understanding.
```suggestion
  return !!(key?.engine);
```
Caveat... if the prettier formatter wants to remove the brackets, then we would have to rather revert to the previous longer form of this.

packages/storage-plugin/tests/storage-for-feature.spec.ts
```diff
@@ -0,0 +1,94 @@
...
+          }
+        ],
+        {
+          initialNavigation: 'disabled'
+        }
+      ),
+      NgxsModule.forRoot([CounterState]),
+      NgxsStoragePluginModule.forRoot({
+        key: [CounterState]
+      })
```
markwhitfeld, line 67: We also need to test the case where the root plugin has been initialised with no keys present (therefore every state is stored), and then the `blog` route is loaded, which will set a key and engine. Will the existing states be stored? I'm not sure if the current logic handles this scenario. 

I think the expectation would be that all states would be stored in the default key, and that explicitly specified feature states would be stored separately. 
Maybe we need to discuss what would happen for any other feature states... What is the current behaviour for lazy feature states?

packages/storage-plugin/src/keys-manager.ts
```diff
@@ -0,0 +1,69 @@
...
+} from '@ngxs/storage-plugin/internals';
+
+interface KeyWithEngine {
+  key: string;
+  engine: StorageEngine;
+}
+
+@Injectable({ providedIn: 'root' })
+export class ɵNgxsStoragePluginKeysManager {
+  readonly keysWithEngines: KeyWithEngine[] = [];
```
markwhitfeld, line 26: Exposing this breaks encapsulation. We need to discuss alternatives to exposing this property.

markwhitfeld (changes requested):

packages/storage-plugin/internals/src/storage-key.ts
rhutchison:
```suggestion
  return Boolean(key?.engine);
```

rhutchison (approved):

isochronous:
Please, add some documentation around this `withStorageFeature` function

isochronous:
And does this method support the `beforeSerialize` and `afterDeserialize` lifecycle methods that are provided via the `.forRoot` version?

PR #2002, opened 2023-05-03
fix: use optional chaining in the ngOnDestroy method

valentinpalkovic:
Refactor state-factory.ts to use optional chaining in the ngOnDestroy method, because it might be called twice if "forFeature" method is used.

## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Currently, when `forFeature` method is used and the application tries to unmount (calling `destroy` method on the application reference), an error is thrown, because the [ngOnDestroy()](https://github.com/ngxs/store/blob/master/packages/store/src/internal/state-factory.ts#L142) method is called twice and `this._actionsSubscription` is `null` in the second call.

Issue Number: N/A

## What is the new behavior?

Removed the non-null assertion and used optional chaining instead.

```
[ ] Yes
[x] No
```

## Other information

Reproduction: https://stackblitz.com/edit/angular-standalone-glfhol?file=src%2Fmain.ts


arturovt:
This is, unfortunately, an invalid fix for now since I've checked your example, and runtime differs in different cases when using `bootstrapApplication` and `bootstrapModule`. The `StateFactory` is not created twice when bootstrapping module but is created each time the `forFeature` is called in a standalone mode. I'll discuss it internally and get back to you shortly.

valentinpalkovic:
@arturovt Thank you for the fast response! :) Let me know whatever works for you as a solution.

PR #2010, opened 2023-05-09
fix(store): ensure `StateFactory` does not connect actions multiple times

arturovt:
This fixes the case where `StateFactory` is created multiple times on root and feature levels and might connect actions multiple times. We have to subscribe to `Actions` stream only once on the root level.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -92,9 +92,6 @@ jobs:
     steps:
       - *attach_workspace
       - run:
-          name: Store Unit Tests
-          command: yarn nx test store --configuration=ci --verbose
-      - run:
           name: Unit Tests
           command: yarn nx run-many --target=test --all --configuration=ci --maxWorkers=4
       - store_test_results:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -92,8 +92,11 @@ jobs:
     steps:
       - *attach_workspace
       - run:
+          name: Store Unit Tests
+          command: yarn nx test store --configuration=ci --verbose
+      - run:
           name: Unit Tests
-          command: yarn nx run-many --target=test --all --configuration=ci --parallel=1
+          command: yarn nx run-many --target=test --all --configuration=ci --verbose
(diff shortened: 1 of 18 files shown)


dmitry-stepanenko:
@markwhitfeld @arturovt  I created a PR https://github.com/ngxs/store/pull/2014, tested changes of this branch and it seems to not hang anymore. See details there

.circleci/config.yml
```diff
@@ -91,6 +91,9 @@ jobs:
       JUNIT_REPORT_PATH: coverage/junit/unit_tests
     steps:
       - *attach_workspace
+      - run:
+          name: Store Unit Tests
+          command: yarn nx test store --configuration=ci --verbose
```
markwhitfeld, line 96: Let's remove this again
```suggestion
```

packages/store/project.json
```diff
@@ -23,7 +23,8 @@
       "configurations": {
         "ci": {
           "ci": true,
-          "codeCoverage": true
+          "codeCoverage": true,
+          "verbose": true
```
markwhitfeld, line 27: It may be worth leaving the `verbose` in there. Let's see how it changes the ci output...

markwhitfeld (approved):

.circleci/config.yml
dmitry-stepanenko, line 96: we need to remove this

markwhitfeld: Haha. Agreed. See my review above 😉

dmitry-stepanenko: this one will still fail, because I haven't changed configurations, only an actual test command for circle ci. I'm not really positive to define amount of workers within jest.config as Jest is usually smart enough to identify amount of CPUs properly (at least it works locally and on GHA)

packages/store/project.json
markwhitfeld: Ok, I removed this, to force a cache invalidation.

markwhitfeld (approved):

PR #2032, opened 2023-07-10
feat: add schematics

dmitry-stepanenko:
This PR is based on progress made in the following branches:
https://github.com/ngxs/store/tree/feature/schematics-2023
https://github.com/ngxs/store/tree/feature/schematics-migration

Currently, available targets:
* `actions`: adds an action as file
* `ng-add`: setups NGXS store with option to:
  * choose plugins to use
  * add imports to app.module.ts
* `starter-kit`: creates actions and state to an existing project
* `state`: create a state file with a selector
* `store`: similar to above but with action handler and an action file


packages/store/schematics/src/ng-add/ng-add.factory.ts
```diff
@@ -0,0 +1,113 @@
...
+
+function addNgxsPackageToPackageJson(options: NgxsPackageSchema): Rule {
+  return (host: Tree, context: SchematicContext) => {
+    const ngxsStoreVersion: string = versions['@ngxs/store'];
+    if (!ngxsStoreVersion) {
+      throw new SchematicsException('Could not resolve the version of "@ngxs/store"');
+    }
+
+    Object.values(LIBRARIES)
+      .filter(lib => options.packages?.includes(lib))
```
profanis, line 43: If the `packages` option is reflected in the CLI, I was wondering if renaming it to `plugins` will be more consistent with this https://www.ngxs.io/plugins

markwhitfeld (approved):

PR #2033, opened 2023-07-11
refactor(websocket-plugin): get rid off `rxjs/webSocket` and use `WebSocket` directly

arturovt:
This commit updated the WebSocket handler implementation and switched to
using the native `WebSocket` directly, instead of relying on `rxjs/webSocket`.
The RxJS WebSocket implementation can be cumbersome when it comes to handling
socket events and lacks simplicity. By handling socket events manually, we avoid
the need to use the RxJS implementation.

### DIFF
--- a/packages/websocket-plugin/src/websocket-handler.ts
+++ b/packages/websocket-plugin/src/websocket-handler.ts
@@ -1,8 +1,7 @@
-import { Injectable, Inject, OnDestroy } from '@angular/core';
+import { Injectable, Inject, OnDestroy, NgZone } from '@angular/core';
 import { Actions, Store, getValue, ofActionDispatched } from '@ngxs/store';
-import { Subscription } from 'rxjs';
-
-import { WebSocketSubject, WebSocketSubjectConfig } from 'rxjs/webSocket';
+import { ReplaySubject, Subject, fromEvent } from 'rxjs';
+import { takeUntil } from 'rxjs/operators';
 
 import {
   ConnectWebSocket,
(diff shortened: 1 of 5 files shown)


packages/websocket-plugin/src/websocket-handler.ts
```diff
@@ -17,157 +16,156 @@ import {
...
-  };
+  private _socket: WebSocket | null = null;
 
-  private typeKey = this.options.typeKey!;
+  private readonly _socketClosed$ = new Subject<void>();
 
-  private subscription = new Subscription();
+  private readonly _typeKey = this._options.typeKey!;
+
+  private readonly _destroy$ = new Subject<void>();
```
markwhitfeld, line 27:
```suggestion
  private readonly _destroy$ = new ReplaySubject<void>();
```
It is always better to use a replay subject for the _destroy$ observable in case a subscription happens after the service is destroyed.

markwhitfeld (changes requested):
We need to make sure that we test and handle correctly the scenario where the connect method is called after the service is destroyed.

packages/websocket-plugin/src/websocket-handler.ts
arturovt: The subscription happens only in constructor.

markwhitfeld (commented):
I would still prefer to have the `_destroy$` changed to a `ReplaySubject` just so that we are consistent with a recommended practice. But I'll leave it up to you on what you do.
Also, can you resolve the conflicts with `master`?
I tried, but saw that it was reintroducing the initialiser code that you removed. Best that you do it 👍

PR #2034, opened 2023-07-12
fix(storage-plugin): access local and session storages globals only in browser

arturovt:


Carniatto (approved):
LGTM. Would be nice to add a small PR description with 1 or 2 lines of motivation. What are the usages that we don't have a browser?

PR #2036, opened 2023-07-16
fix(storage-plugin): require only `getItem` and `setItem` on engines

arturovt:
This commit updates the signature of the `StorageEngine` interface
and removes unnecessary properties that are not being used by the storage
plugin. The properties `length`, `removeItem`, and `clear` have been removed.
This change makes it easier to maintain storage engines that implement the
interface because only the necessary properties will be required. It is important
to note that this change is not considered a breaking change since properties are
being removed rather than added.


Carniatto (approved):
LGTM

PR #2038, opened 2023-07-21
fix(devtools-plugin): do not re-enter Angular zone when resetting state

arturovt:


markwhitfeld (approved):
Is there a related issue that this resolves?
I can't find any that are obviously related.
If there isn't then could you add a description to this PR with the motivation for this change.

I am approving so that you can go ahead an merge, but please do add the necessary details.

PR #2042, opened 2023-08-12
feat(store): improve selector decorator types (#1929)

ConnorUllmann:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

The `@Selector` decorator does not enforce that the arguments of the decorated function match the return types of the selectors that it depends on.

Issue Number: #1929 

## What is the new behavior?

The `@Selector` decorator enforces that the arguments of the functions that it decorates match the return values of the selectors that it was passed. 

Limitations:
- The models of state classes are not inferred. Any type is allowed for these arguments (however `StateToken` works correctly).
- The arguments list is checked against both the possibility of an injected state and the possibility of no injected state. This is regardless of the actual setting, and regardless of whether the decorator is used in a State class or in an independent class.

```
[ ] Yes
[x] No
```

Although this is not a breaking change in the sense that a codebase with correct typings on their selector functions will require no changes, the purpose of this change is to expose failures to do that typing correctly and prevent them in the future. As a result, I would expect any reasonably-sized codebase to require fixes to their selector implementations as a result of this change.


markwhitfeld (commented):
@ConnorUllmann Thank you so much for this contribution!
My apologies for the long delay in reviewing this.
We would like to include your PR in NGXS v17, which is landing very soon.

Would it be possible to add some comments / descriptions to the types that you created which explain how they work / what they do / etc. so that it is easier for us to be able to understand and maintain this code into the future.

One other quick comment is that after this PR is merged, we would like to build on top of your changes to add a deprecation notice to any occurrence of the `@Selector(...)` decorator where the selector function is expecting the container state as the first parameter (except for the parameterless version). I am just trying to figure out where exactly that doc comment would be 😄  .

ConnorUllmann:
@markwhitfeld Glad to hear it! I'm writing up the descriptions of those types now.

I had to make some modifications in order to allow the parameterless call to not be deprecated when injecting container state while deprecating the version of the call with paramters. Do the examples in the image below accurately cover what you're expecting to see deprecated?

I can include the deprecation myself in my update, if you like, otherwise I'll note where it will need to be put.

image

markwhitfeld:
Yes, I agree with what you have there.
If you would like to add the deprecation notice, it would be the same as the one found here:
https://github.com/ngxs/store/blob/3bbb9e1609145637ac7e3adce4f990844848c412/packages/store/internals/src/symbols.ts#L86
The deprecation notice should point to the page that will go live when the release goes out:
`https://ngxs.io/deprecations/inject-container-state-deprecation.md`
You can see the development version of this page [here](https://www.ngxs.io/v/master/deprecations/inject-container-state-deprecation).

markwhitfeld (changes requested):
Fantastic work.
Can I ask you to name the `T` and `U` type placeholders according to what they are?
This should be the last puzzle piece to make this more understandable and therefore maintainable.

arturovt (approved):

markwhitfeld (approved):
Thank you so much. This is really super work!

PR #2052, opened 2023-09-15
fix(storage-plugin): expose storage plugin internals privately through subpackage

arturovt:
This commit exposes some of the storage plugin internals through a subpackage, following a similar approach to how router plugin internals are exposed. These internals are necessary to expose if someone wishes to extend the storage engine's behavior. For example, to check for the default state key `@@STATE` or to use helper functions like extracting string keys.

### DIFF
--- a/packages/storage-plugin/internals/src/storage-key.ts
+++ b/packages/storage-plugin/internals/src/storage-key.ts
@@ -23,7 +23,7 @@ export type StorageKey = string | StateClass | StateToken<any> | KeyWithExplicit
 
 /** This symbol is used to store the metadata on state classes. */
 const META_OPTIONS_KEY = 'NGXS_OPTIONS_META';
-export function ɵexctractStringKey(storageKey: StorageKey): string {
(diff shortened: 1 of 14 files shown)


packages/storage-plugin/internals/src/storage-key.ts
```diff
@@ -23,9 +23,9 @@ export type StorageKey = string | StateClass | StateToken<any> | KeyWithExplicit
 
 /** This symbol is used to store the metadata on state classes. */
 const META_OPTIONS_KEY = 'NGXS_OPTIONS_META';
-export function exctractStringKey(storageKey: StorageKey): string {
+export function ɵexctractStringKey(storageKey: StorageKey): string {
```
markwhitfeld, line 26: Let's fix the spelling of this API before we expose it.
```suggestion
export function ɵextractStringKey(storageKey: StorageKey): string {
```

markwhitfeld (changes requested):
Small API spelling tweak

PR #2060, opened 2023-09-23
feat(store): add support for standalone APIs in schematics

dmitry-stepanenko:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Adding support for standalone APIs in schematics

With this PR `ng-add` will be able to add NGXS plugins both to root module as modules and to app.config as provider functions.

image
image

Please note that there's a new folder `packages/store/schematics/src/utils/ng-utils`. It contains files that were copied from the `@schematics/angular@16.2` package. It contains very convenient utils that allow adding declarations both in standalone and regular Angular apps. There's no point to reinvent the bicycle and try to implement that functionality ourselves as we will be able to use them directly from `@schematics/angular` as we'll start to support the v17 as the minimum required version.

Issue Number: N/A

```
[ ] Yes
[x] No
```

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -9,11 +9,12 @@ checks:
   method-lines:
     config:
       threshold: 50
-exclude_paths:
+exclude_patterns:
   - '.github/'
   - 'assets/'
   - 'docs/'
   - 'packages/**/tests/'
+  - 'packages/store/schematics/src/utils/ng-utils/'
   - 'integrations/'
   - 'integration/'
   - 'tools/'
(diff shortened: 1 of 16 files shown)


markwhitfeld (changes requested):
@dmitry-stepanenko There has been a large drop on test coverage in this PR. I think that this is a result of the code that was copied across.
Please could you either add some tests (maybe if there are some tests to copy too), or find a way to exclude these folders from the test coverage metrics.

dmitry-stepanenko:
@markwhitfeld I've excluded the utils folder in `.codeclimate.yml` and it now shows a single .ts file within `ratings` section as expected. However, it still tries to calculate the coverage with those excluded files. Any ideas? Maybe I can try to push the change to the config in a separate PR and then recreate this one?

markwhitfeld (approved):

PR #2062, opened 2023-09-25
refactor: move metadata into internals

arturovt:
This commit includes several updates:

1) Moves metadata interfaces and functions to the
`@ngxs/store/internals` subpackage.
2) Transfers interfaces and objects from `plugin_api.ts` to the
`@ngxs/store/plugins` subpackage.
3) Adds the `@deprecated` annotation to functions and types that were
exported from `public_to_deprecate.ts`.


packages/store/internals/src/memoize.ts
```diff
@@ -28,7 +28,7 @@ function areArgumentsShallowlyEqual(
  *
  * @ignore
  */
-export function memoize<T extends (...args: any[]) => any>(
+export function ɵmemoize<T extends (...args: any[]) => any>(
```
markwhitfeld, line 31: Shouldn't this be aliased with the barred-o in the `internals` package exports as opposed to internally within the package?

packages/store/internals/src/index.ts
```diff
@@ -1,12 +1,9 @@
...
-export { INITIAL_STATE_TOKEN, InitialState } from './initial-state';
-export {
-  PlainObjectOf,
-  PlainObject,
-  ɵStateClass,
-  ɵMETA_KEY,
-  ɵMETA_OPTIONS_KEY,
-  ɵSELECTOR_META_KEY
-} from './symbols';
+export * from './symbols';
```
markwhitfeld, line 1: We are exposing a lot more here than originally intended.
Do we have an option to not expose so much?

packages/store/internals/src/memoize.ts
arturovt: I think we could make all of the namings without barred-o and re-export them with `ɵ`, but this package is 'exports-only', which means it only exports the stuff which is used outside. As thus I'm guessing this would've been an extra code.

packages/store/internals/src/index.ts
arturovt: Everything which is exported from there is used outside.

PR #2066, opened 2023-09-26
fix(store): show error when state initialization order is invalid

arturovt:
This commit adds a `console.error` when the `UpdateState` is dispatched before the `InitState`. This typically indicates that the state initialization order is invalid and must be updated to ensure that states are initialized in the correct order. The incorrect order may prevent `ngxsOnInit` from being called on feature states or lead to other unpredictable errors.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ $ npm install @ngxs/store@dev
 
 ### To become next patch version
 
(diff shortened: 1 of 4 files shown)


markwhitfeld (approved):

markwhitfeld (changes requested):
Please add to the change log

Carniatto (approved):
LGTM

PR #2068, opened 2023-10-02
chore: `tsconfig.spec.json` files should extend from root `tsconfig.spec.json`

arturovt:


markwhitfeld (changes requested):
What is the motivation for this change?

PR #2083, opened 2023-11-28
fix(store): ensure features are initialized after root state

arturovt:
We have a check that ensures states are initialized in the correct order and that the `UpdateState` action shouldn't be dispatched before the `InitState` action. This code updates the order of environment providers to place features at the end. As a result, the root environment initializer runs earlier, dispatching the `InitState` action.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,8 @@ $ npm install @ngxs/store@dev
 
 ### To become next patch version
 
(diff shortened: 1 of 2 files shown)


markwhitfeld (approved):

markwhitfeld:
@arturovt do you have any idea on why the unit tests might be timing out?

arturovt:
I don't know, it's typical for Circle to time out, but there's no reason. Seems like it's executed more than it's allowed to.

PR #2089, opened 2023-12-14
feat: state, store and starter-kit schematics support a project option

profanis:
When we use the schematics to create
- state
- store
- starter-kit

we can rely on the project type and have the spec files with either `provideStore()` or `NgxsModule.forRoot()` if it's a standalone or not.

Furthermore, if we have a mono-repo workspace we can provide the `project` where we want to have the schematic run.

### DIFF
--- a/docs/concepts/state.md
+++ b/docs/concepts/state.md
@@ -16,12 +16,15 @@ You have the option to enter the options yourself
 ng generate @ngxs/store:state --name NAME_OF_YOUR_STATE
 ```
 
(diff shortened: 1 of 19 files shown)


arturovt (approved):
LGTM

dmitry-stepanenko (approved):

markwhitfeld (approved):
@profanis Great work!!!
The only thing missing is an update to the `changelog.md` file describing these changes.
But, I'll merge without it so that you don't have to dismiss the existing reviews.

PR #2093, opened 2024-02-04
fix(store): change `instanceof Promise` to `isPromise` to allow any promisable object

arturovt:
We need to use `isPromise` instead of checking whether `result instanceof Promise`.
In zone.js patched environments, `global.Promise` is the `ZoneAwarePromise`. Some APIs,
which are likely not patched by zone.js for certain reasons, might not work with `instanceof`.
For instance, the dynamic import returns a native promise (not a `ZoneAwarePromise`), causing
this check to be falsy.


arturovt:
Ugh the Circle always randomly failing with a timeout...

markwhitfeld (approved):

markwhitfeld (approved):

PR #2097, opened 2024-02-05
feat(store): add `selectSignal`

arturovt:


packages/store/src/store.ts
```diff
@@ -93,6 +101,23 @@ export class Store {
...
+   */
+  selectSignal<T>(
+    selector: (state: any, ...states: any[]) => T,
+    options?: CreateComputedOptions<T>
+  ): Signal<T>;
+  selectSignal<T = any>(
+    selector: string | Type<any>,
+    options?: CreateComputedOptions<T>
+  ): Signal<T>;
+  selectSignal<T>(selector: StateToken<T>, options?: CreateComputedOptions<T>): Signal<T>;
```
markwhitfeld, line 115: Because immutability is a fundamental premise for data to exist in our state, a user should never have a reason to specify the equality comparison function (which is the only thing that the `CreateComputedOptions` interface provides). 
We should therefore not expose the `options` as an argument to this function.

packages/store/src/store.ts
```diff
@@ -93,6 +101,23 @@ export class Store {
...
+   * Select a signal from the state.
+   */
+  selectSignal<T>(
+    selector: (state: any, ...states: any[]) => T,
+    options?: CreateComputedOptions<T>
+  ): Signal<T>;
+  selectSignal<T = any>(
+    selector: string | Type<any>,
+    options?: CreateComputedOptions<T>
+  ): Signal<T>;
```
markwhitfeld, line 114: We can remove some of the bad "non type-safe" NGXS legacy for this new API. We should therefore not support specifying a string or a function as arguments for this new API. This also means that specifying a state class should not be supported.

We should only support `TypedSelector<TModel>` as the sole argument. Essentially this constrains our users to use the `selectSignal` function with a typed selector or a StateToken, both of which carry type information.

integration/app/app.component.ts
```diff
@@ -39,10 +38,10 @@ export class AppComponent implements OnInit {
...
 
   greeting: string;
-  todos$: Observable<Todo[]> = this._store.select(TodoState);
-  pandas$: Observable<Todo[]> = this._store.select(TodoState.getPandas);
-  pizza$: Observable<Pizza> = this._store.select(TodosState.getPizza);
-  injected$: Observable<string> = this._store.select(TodosState.getInjected);
+  todos: Signal<Todo[]> = this._store.selectSignal(TodoState);
+  pandas: Signal<Todo[]> = this._store.selectSignal(TodoState.getPandas);
+  pizza: Signal<Pizza> = this._store.selectSignal(TodosState.getPizza);
+  injected: Signal<string> = this._store.selectSignal(TodosState.getInjected);
```
markwhitfeld, line 44: This comment refers to all the changes in the integration app...
We should have a signal and an observable variant in the implementations because both usages are still valid scenarios.

markwhitfeld (changes requested):
Thanks @artur. Great start to this!!!!
The biggest point here is about the new API only supporting typed selectors, and not exposing the `options` for the computed).
The other things that would still be needed in this PR are:
- Update the integration app to have both signal and observable variants that pass the integration tests
- Add unit tests for the new `selectSignal` function. We can probably copy-paste all of the type-safe tests for `store.select` as a baseline for this.

markwhitfeld (commented):
Looking good. Do you know what is going on with the types test?

arturovt:
> Looking good. Do you know what is going on with the types test?

`@angular/core/rxjs-interop` is using some latest TS features not supported by the TS version installed in `store/types`. I can't upgrade that TS version because `dtslint` uses `tslint` which is tied to lower TS version.

arturovt:
@markwhitfeld I switched from dtslint to ESLint that supports linting files with `$Expect` comments.

PR #2100, opened 2024-02-19
feat(storage-plugin): require explicit options when providing storage plugin

arturovt:
This commit updates the implementation of the register signature for the storage plugin.
It now always requires `options` to be provided. We want to ensure clarity regarding the
states that we're going to serialize. When no `options` are provided, we implicitly provide
the `@@STATE` key, which defines that all states should be serialized. This implementation
doesn't compromise extendability because we implicitly rely on the internal mechanism that
transforms options. We should always require options to be provided and require developers
to specify serializable states explicitly.

This would allow us to have feature states because the feature registration mechanism would
also know if the user explicitly specified all states to be serialized at the root level or not.

**Note:** This change is breaking. However, it's time to make breaking changes to move forward,
as the current implementation is stale and acts as a blocker.

### DIFF
--- a/docs/plugins/storage.md
+++ b/docs/plugins/storage.md
@@ -20,7 +20,7 @@ import { NgxsModule } from '@ngxs/store';
 import { NgxsStoragePluginModule } from '@ngxs/storage-plugin';
 
 @NgModule({
-  imports: [NgxsModule.forRoot([]), NgxsStoragePluginModule.forRoot()]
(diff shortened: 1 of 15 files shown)


Carniatto (approved):
LGTM

@arturovt Should we plan to add a migration schematics to ease the impact of these breaking changes?

rhutchison (approved):

rhutchison (approved):

arturovt:
> LGTM
> 
> @arturovt Should we plan to add a migration schematics to ease the impact of these breaking changes?

Yeah, I talked to @profanis .

PR #2103, opened 2024-03-01
ci: switch from `dtslint` to ESLint

arturovt:
This commit removes the `dtslint` dependency because it was deprecated a long time ago and is no longer maintained. The issue was with supporting newer TypeScript versions, as `dtslint` uses an older TypeScript compiler API. We now use `eslint-plugin-expect-type` to lint against `$Expect` comments.


rhutchison (approved):

markwhitfeld:
Great find! I'm glad we have a decent dtslint replacement now.

PR #2108, opened 2024-03-14
feat(storage-plugin): migrate to the property 'keys'

profanis:
This schematic migrates the following cases:

**From**
```ts
NgxsStoragePluginModule.forRoot()
```

**To**
```ts
   keys: '*' 
})
```

---

**From**
```ts
   optionProp: 'optionValue' 
})
```

**To**
```ts
   optionProp: 'optionValue',
   keys: '*' 
})
```
---

**From**
```ts
   optionProp: 'optionValue',
   key: 'keyValue'
})
```

**To**
```ts
   optionProp: 'optionValue',
   keys: ['keyValue'] 
})
```


packages/storage-plugin/schematics/src/ng-generate/keys-migration.spec.ts
```diff
@@ -0,0 +1,122 @@
...
+          AppComponent
+        ],
+        imports: [
+          BrowserModule,
+          AppRoutingModule,
+          NgxsStoragePluginModule.forRoot()
+        ],
+        bootstrap: [AppComponent]
+      })
+      export class AppModule { }
```
dmitry-stepanenko, line 41: can you please add cases like this 
```typescript
const imports = [
  NgxsStoragePluginModule.forRoot()
];


@Module({

imports,

})

export class MyModule { ... }
```

along with support for standalone applications

packages/storage-plugin/schematics/src/ng-generate/keys-migration.spec.ts
```diff
@@ -0,0 +1,122 @@
...
+    `;
+
+    appTree.overwrite('/projects/foo/src/app/app.module.ts', newContent);
+
+    const tree = await ngxsSchematicRunner.runSchematic('ngxs-keys-migration', {}, appTree);
+
+    const contentUpdate = tree!.readContent('/projects/foo/src/app/app.module.ts');
+    expect(contentUpdate).toContain(`keys: '*'`);
```
dmitry-stepanenko, line 49: this should probably be better
```suggestion
    expect(contentUpdate).toMatchSnapshot();
```

packages/storage-plugin/schematics/src/ng-generate/index.ts
```diff
@@ -0,0 +1,91 @@
...
+      const sourceFile = tsProject.createSourceFile(path, fileContent!);
+      const hasStoragePluginImported = sourceFile.getImportDeclaration(
+        importDecl => importDecl.getModuleSpecifierValue() === '@ngxs/storage-plugin'
+      );
+
+      // do not try migrating if the storage plugin is not imported
+      if (hasStoragePluginImported) {
+        sourceFile.forEachDescendant(node => {
+          if (
+            ts.Node.isCallExpression(node) &&
```
dmitry-stepanenko, line 65: I'd rather do something like
`node.getKindName() === 'CallExpression'`

That will be safer because under the hood this guy does something like `node.getKind() === typescript.SyntaxKind.CallExpression`, which is an enum that can return different numbers depending on minor typescript package versions. And ts-morph kindly does not set typescript as a dependency

markwhitfeld:
@profanis Note that the last conversion in the issue description is wrong. 
The valid values for the `keys` option is either `'*'` or an array with strings, or objects
It should be:
From
```
   optionProp: 'optionValue',
   key: 'keyValue'
  // or key: NovelsState
})
```
To
```
   optionProp: 'optionValue',
   keys: [ 'keyValue' ]
  // or keys: [ NovelsState ]
})
```
Note the second variant above where the StateClass is passed.
Also a StateToken is a valid value there.

We should also include this conversion where an array is already provided for the `key` prop. In this case we should just rename the `key` prop to `keys`: 
From
```
   optionProp: 'optionValue',
   key: ['keyValue', NovelsState, { key: 'novels', engine: MyEngineX } ]
})
```
To
```
   optionProp: 'optionValue',
   keys: ['keyValue', NovelsState, { key: 'novels', engine: MyEngineX } ]
})
```
@arturovt Am I correct in saying this?

arturovt:
Correct. `keys` is now always a list of values, so we shouldn't need to internally check whether it's a list or not and perform any unnecessary transformations, which I found cumbersome to deal with. After the migration is completed, `keys` should always be present. Whatever `key` is (if it exists), it should be turned into a list. Alternatively, if it's already a list, it should be renamed to `keys`. If `keys` doesn't exist, it should be added as `keys: *`.

packages/storage-plugin/schematics/src/ng-generate/keys-migration.spec.ts
profanis: I added a unit test for this case

```ts
const imports = [
  NgxsStoragePluginModule.forRoot()
];

@NgModule({

imports,

})

export class MyModule { ... }
```
The `imports` declaration is covered by the schematic since this ` NgxsStoragePluginModule.forRoot()` falls under the `CallExpression` and `PropertyAccessExpression`

profanis: As for the standalone applications, I was under the impression that the standalone will be released with 3.9.
@arturovt can you please advise?

packages/storage-plugin/schematics/src/ng-generate/keys-migration.spec.ts
profanis: Yes indeed, that's much better. I did this change

packages/storage-plugin/schematics/src/ng-generate/index.ts
profanis: Thanks for explaining this. I did the change

profanis:
> @profanis Note that the last conversion in the issue description is wrong. The valid values for the `keys` option is either `'*'` or an array with strings, or objects It should be: From
> 
> ```
> NgxsStoragePluginModule.forRoot({ 
>    optionProp: 'optionValue',
>    key: 'keyValue'
>   // or key: NovelsState
> })
> ```
> 
> To
> 
> ```
> NgxsStoragePluginModule.forRoot({ 
>    optionProp: 'optionValue',
>    keys: [ 'keyValue' ]
>   // or keys: [ NovelsState ]
> })
> ```
> 
> Note the second variant above where the StateClass is passed. Also a StateToken is a valid value there.
> 
> We should also include this conversion where an array is already provided for the `key` prop. In this case we should just rename the `key` prop to `keys`: From
> 
> ```
> NgxsStoragePluginModule.forRoot({ 
>    optionProp: 'optionValue',
>    key: ['keyValue', NovelsState, { key: 'novels', engine: MyEngineX } ]
> })
> ```
> 
> To
> 
> ```
> NgxsStoragePluginModule.forRoot({ 
>    optionProp: 'optionValue',
>    keys: ['keyValue', NovelsState, { key: 'novels', engine: MyEngineX } ]
> })
> ```
> 
> @arturovt Am I correct in saying this?

Thanks @markwhitfeld 
I did the change and added the relevant unit tests

markwhitfeld (approved):
Fantastic work!!!
Just a small question... Do we want to call this `ngxs-keys-migration`, or rather `ngxs-migrate-storage-plugin`?
I feel like the first is quite detail oriented, which may be irrelevant to the user.
The second makes it clear which plugin it is migrating.

Also a follow on question. I don't know how schematics works, but will this somehow be included in the migrations run for `ng update`? That could be a separate PR, because I'm sure there will be other schematics to run as well.

profanis:
> Fantastic work!!! Just a small question... Do we want to call this `ngxs-keys-migration`, or rather `ngxs-migrate-storage-plugin`? I feel like the first is quite detail oriented, which may be irrelevant to the user. The second makes it clear which plugin it is migrating.
> 
> Also a follow on question. I don't know how schematics works, but will this somehow be included in the migrations run for `ng update`? That could be a separate PR, because I'm sure there will be other schematics to run as well.

The command to run the schematic is against the storage-plugin, so it has the context.
`ng g @ngxs/storage-plugin:keys-migration`

I do not have a strong opinion of `keys-migration`, but I feel it's explanatory enough.

As for the `ng update` I think this should be a separate schematic where we will have to chain other schematics.

markwhitfeld:
@profanis Are you happy to merge this now?

profanis:
> @profanis Are you happy to merge this now?

Yes it's fine by me!

PR #2110, opened 2024-03-14
fix(store): `ctx.getState` should return `DeepReadOnly<>`

arturovt:
This commit modifies the return value of `StateContext.getState` from `<T>` to `ReadonlyDeep<T>`.
The `ReadonlyDeep` type is provided by the `type-fest` package, which is now included in the
dependencies of `@ngxs/store` and will be installed automatically with it.

This change is necessary because `getState` should return an immutable object, and users should
not have any reason to directly modify it.

Please note that this change is breaking and is part of the v4 release preparation.

### DIFF
--- a/packages/hmr-plugin/src/internal/hmr-lifecycle.ts
+++ b/packages/hmr-plugin/src/internal/hmr-lifecycle.ts
@@ -42,7 +42,7 @@ export class HmrLifecycle<T extends Partial<NgxsHmrLifeCycle<S>>, S> {
     if (typeof this.ngAppModule.hmrNgxsStoreBeforeOnDestroy === 'function') {
       state = this.ngAppModule.hmrNgxsStoreBeforeOnDestroy(ctx);
     } else {
-      state = ctx.getState();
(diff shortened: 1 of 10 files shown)


arturovt:
notes from meeting: let's use `type-fest`.

packages/hmr-plugin/src/internal/hmr-lifecycle.ts
```diff
@@ -42,7 +42,7 @@ export class HmrLifecycle<T extends Partial<NgxsHmrLifeCycle<S>>, S> {
     if (typeof this.ngAppModule.hmrNgxsStoreBeforeOnDestroy === 'function') {
       state = this.ngAppModule.hmrNgxsStoreBeforeOnDestroy(ctx);
     } else {
-      state = ctx.getState();
+      state = ctx.getState() as Partial<S>;
```
arturovt, line 45: Leaving as this because I don't wanna do any changes in the HMR plugin, just making compiler happy...

PR #2113, opened 2024-03-16
Revert "fix(store): `ctx.getState` should return `DeepReadOnly<>`"

arturovt:
Reverts ngxs/store#2110


markwhitfeld:
@arturovt what is the reason for the revert?

arturovt:
DMed in google chat

PR #2128, opened 2024-03-19
docs: update `concepts/select.md`

arturovt:
This commit completely updates the `select.md` to adopt a signal-first approach, as previously, all examples were based on observables. Most developers transitioning to NGXS would prefer to see examples using signals, as this concept is being heavily promoted by Angular.

Additionally, this commit removes examples using the `@Select` decorator since it's not usable and has critical issues due to its lack of integration with Angular's dependency injection.

Furthermore, the "special considerations" block has been removed, as the issues it addressed are no longer a concern.

### DIFF
--- a/docs/concepts/select.md
+++ b/docs/concepts/select.md
@@ -2,37 +2,9 @@
 
 Selects are functions that slice a specific portion of state from the global state container.
 
-In CQRS and Redux patterns, we keep READ and WRITE separated. This pattern also exists in NGXS.
-When we want to read data out of our store, we use a select operator to retrieve this data.
+In CQRS and Redux patterns, we maintain a separation between READ and WRITE operations. This pattern is also present in NGXS. When we need to retrieve data from our store, we utilize a select operator to access this data.
 
-In NGXS, there are two methods to select state, we can either call the `select` method on the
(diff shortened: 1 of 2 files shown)


docs/concepts/select.md
```diff
@@ -162,22 +133,8 @@ These options can also be provided through the `@SelectorOptions` decorator at a
 
 #### `injectContainerState`
 
-- `true` will cause all selectors defined within a state class to receive the container class' state model as their first parameter. As a result every selector would be re-evaluated after any change to that state.  
-  **NOTE:** _This is not ideal, therefore this setting default will be changing to `false` in NGXS v4._
+- `true` will cause all selectors defined within a state class to receive the container class' state model as their first parameter. As a result every selector would be re-evaluated after any change to that state.
 - `false` will prevent the injection of the container state model as the first parameter of a selector method (defined within a state class) that joins to other selectors for its parameters.
```
markwhitfeld, line 137: We could possibly state as the first point for this option that this property is only useful for migrating codebases from NGXS v3.x and that it is not recommended to keep this set as `true`.

We should also mention that the default is now `false`.
Under the details for `true` we should mention that this should only be used during migrations.

markwhitfeld (approved):

PR #2130, opened 2024-03-20
docs: update concepts, lazy loading guide and examples

arturovt:


markwhitfeld (approved):
Great updates!
I am wondering if we most likely need a page dedicated to showing the `NgModule` versions of  the APIs. If we remove modules totally from the docs, then those with older codebases might not have a handle on how to use NGXS in these codebases.
We could link to this "NgModule-based APIs" page throughout our docs wherever the standalone APIs are demonstrated.

PR #2134, opened 2024-03-21
feat(store): expose `withNgxsDevelopmentOptions` standalone API

arturovt:


docs/advanced/monitoring-unhandled-actions.md
```diff
@@ -24,42 +42,51 @@ Setting `warnOnUnhandledActions` to a truthy value will tell the logger to warn
...
+if (ngDevMode) {
+  providers.push({
+    provide: ENVIRONMENT_INITIALIZER,
+    multi: true,
+    useValue: () => {
+      const unhandledActionsLogger = inject(NgxsUnhandledActionsLogger);
       unhandledActionsLogger.ignoreActions(LazyAction);
     }
-  }
+  });
```
markwhitfeld, line 80: Unrelated to this PR, do you think we could provide a more convenient way of doing this type of thing?

packages/store/src/internal/state-factory.ts
```diff
@@ -60,7 +60,7 @@ import { ensureStateNameIsUnique, ensureStatesAreDecorated } from '../utils/stor
 import { ensureStateClassIsInjectable } from '../ivy/ivy-enabled-in-dev-mode';
 import { NgxsUnhandledActionsLogger } from '../dev-features/ngxs-unhandled-actions-logger';
 
-const NG_DEV_MODE = typeof ngDevMode === 'undefined' || ngDevMode;
+const NG_DEV_MODE = typeof ngDevMode !== 'undefined' && ngDevMode;
```
markwhitfeld, line 63: There are many variants of this throughout the codebase.
Which is correct?
image

markwhitfeld (approved):

packages/store/src/internal/state-factory.ts
arturovt: I forgot to revert.

PR #2137, opened 2024-03-21
feat(store): implement NGXS unhandled error handler

arturovt:


packages/store/src/internal/unhandled-rxjs-error-callback.ts
```diff
@@ -0,0 +1,40 @@
...
+    return true;
+  }
+  return false;
+}
+
+export function assignUnhandledCallback(error: any, callback: VoidFunction) {
+  // Since the error can be essentially anything, we must ensure that we only
+  // handle objects, as weak maps do not allow any other key type besides objects.
+  // The error can also be a string if thrown in the following manner: `throwError('My Error')`.
+  if (error !== null && typeof error === 'object') {
```
markwhitfeld, line 30: Do we need a strategy for handling non-object errors?

packages/store/src/ngxs-error-handler.ts
```diff
@@ -0,0 +1,29 @@
...
+  private _errorHandler: ErrorHandler = null!;
+
+  constructor(
+    private _ngZone: NgZone,
+    private _injector: Injector
+  ) {}
+
+  handleError(_error: any, _context: NgxsErrorContext): void {
+    // Retrieve lazily to avoid cyclic dependency exception.
+    this._errorHandler = this._errorHandler || this._injector.get(ErrorHandler);
```
markwhitfeld, line 18: Given the other changes to load this lazily in the StateFactory, we should move this back to the constructor.
This would also be closer to what a user may implement.

packages/store/tests/action.spec.ts
```diff
@@ -827,17 +827,17 @@ describe('Action', () => {
           'cancellableAction(1) - start',
           'action1 obs - subscribe',
           'Action 2 - dispatching',
-          'Action 1 - dispatch complete',
           'action1 obs - unsubscribe',
+          'Action 1 - dispatch complete',
```
markwhitfeld, line 831: Is this change because we unsubscribe before subscribing the real subscriber.
Would it go back to what it was if we unsubscribed after adding the real subscriber (in the fallback operator) ?

markwhitfeld (commented):
Awesome work here!
A few comments.
And just wondering if the original issue is represented in a test?

packages/store/tests/action.spec.ts
arturovt: This change is related to RxJS upgrade, seems like the sequence of some stuff has changed internally

packages/store/src/internal/unhandled-rxjs-error-callback.ts
arturovt: I don't think so. Here's why:

- the ecma 2022 standard specifies the `Error` class which is intended to represent errors
- while other types are technically allowed in `throw` statements, such as `throw some_string`, it's widely regarded as bad practice. This is because it doesn't include stack trace information, making debugging more difficult and less informative

arturovt:
> And just wondering if the original issue is represented in a test?

The original issue was that the error handler was not being called on each error. We've decided not to support this anymore because it's strange to log errors even if they're handled. For instance, if someone is using Rollbar and providing their custom `ErrorHandler` which logs errors to Rollbar, even handled errors would be logged to Rollbar, which is completely undesirable.

PR #2141, opened 2024-03-23
 feat(signals): provide `@ngxs/signals` package with signals utilities

arturovt:


rhutchison:
There are 3 utilities, select, produceActions, and produceSelectors

select:

```
@Component({...})
export class SomeComponent {
  user = select(AuthQueries.getUser);
}
```

produceActions/produceSelectors:

Allow users to quickly adapt ngxs store to @ngrx/signals (signalStore).  We can provide the following recipes in the documentation:

```
export function withSelectors<T extends SelectorMap>(selectors: T) {
  return signalStoreFeature(withComputed(() => produceSelectors(selectors)));
}

export function withActions<T extends ActionMap>(actions: T) {
  return signalStoreFeature(withMethods(() => produceActions(actions)));
}
```

Usage:

```
export const UserState = signalStore(
  withSelectors({
    user: AuthQueries.getUser
  }),
  withActions({
    login: AuthActions.login,
    logout: AuthActions.logout
  })
)
```

packages/signals/src/types.ts
```diff
@@ -0,0 +1,4 @@
+// There's no reason to call the following function with an empty object as
+// `produceActions({})` or `produceSelectors({})`. We should prevent invalid
+// or incorrect use cases.
+export type RequireAtLeastOneProperty<T> = keyof T extends never ? never : T;
```
rhutchison, line 4: if we're using type-fest, you can use `NonEmptyObject`

packages/signals/src/produce-actions.ts
```diff
@@ -0,0 +1,27 @@
+import { inject } from '@angular/core';
+import { ActionDef, Store } from '@ngxs/store';
+import { Observable } from 'rxjs';
+
+import { RequireAtLeastOneProperty } from './types';
```
rhutchison, line 5: `import type { NonEmptyObject } from 'type-fest';`

packages/signals/src/produce-selectors.ts
```diff
@@ -0,0 +1,24 @@
+import { Signal, inject } from '@angular/core';
+import { Store, TypedSelector, ɵSelectorReturnType } from '@ngxs/store';
+
+import { RequireAtLeastOneProperty } from './types';
```
rhutchison, line 4: `import type { NonEmptyObject } from 'type-fest';`

packages/signals/tests/produce-selectors.spec.ts
```diff
@@ -0,0 +1,39 @@
...
+
+    // Act
+    const selectors = runInInjectionContext(TestBed, () =>
+      produceSelectors({
+        counter: CounterState.getCounter
+      })
+    );
+
+    // Assert
+    expect(selectors.counter()).toEqual(0);
```
markwhitfeld, line 37: What do you think about wrapping this signal with a readonly property getter?

packages/signals/types/tests/produce-actions.lint.ts
```diff
@@ -0,0 +1,41 @@
...
+      constructor(readonly name: string) {}
+    }
+  
+    class ValidActionWithMultipleParameters {
+      static readonly type = 'Valid Action';
+
+      constructor(readonly name_1: string, readonly name_2: number) {}
+    }
+
+    produceActions({ action: ValidAction }); // $ExpectType { action: (name: string) => Observable<void>; }
```
markwhitfeld, line 38: What do you think about this action dispatching function returning a promise as opposed to an observable?

markwhitfeld (commented):
I think that this is a really great, simple utility. Nicely done!
Just needs docs ❤️

packages/signals/tests/produce-selectors.spec.ts
arturovt: Yep, makes sense

packages/signals/types/tests/produce-actions.lint.ts
arturovt: Is it to wrap it with `createDispatchFn().toPromise()` or smth?

arturovt:
questions:
- do we keep `select` name for the utility function?
- I added docs and placed it under `docs/concepts/select/signals.md`, I ain't sure if it should be under concepts/select. Any ideas?

packages/signals/src/produce-selectors.ts
```diff
@@ -0,0 +1,27 @@
...
+
+export function produceSelectors<T extends SelectorMap>(
+  selectorMap: RequireAtLeastOneProperty<T>
+) {
+  const store = inject(Store);
+
+  return Object.entries(selectorMap).reduce((accumulator, [key, selector]) => {
+    Object.defineProperty(accumulator, key, {
+      value: store.selectSignal(selector)
+    });
```
markwhitfeld, line 16: Great, but I was proposing that we implement a getter that calls the signal.
This would remove the need to use the brackets to get the value.

arturovt: do you mean the following?

```ts
const signal = store.selectSignal(selector);

defineProp(object, key, {
  get: () => signal()
})
```

arturovt: I believe this would deviate from the typical approach of using signals, especially in templates, where signals are typically accessed using brackets to retrieve their values.

markwhitfeld: Yes, I was meaning that, but now I'm thinking that get wrapping behaviour could be better acheived as another utility (maybe a future discussion).

packages/signals/types/tests/produce-actions.lint.ts
markwhitfeld: Yes. Because it makes the consumption of the result a lot simpler. There will only ever be one result. This is what promises are good at. What do you think?

arturovt: I’m not against promises, my only concern is the cancellation.

markwhitfeld:
> questions:
> - do we keep `select` name for the utility function?
> - I added docs and placed it under `docs/concepts/select/signals.md`, I ain't sure if it should be under concepts/select. Any ideas?

I think we can leave this utility function out for the first release of this😢.
There needs to be a final discussion on naming.
I think that would be the correct place in the docs.

rhutchison:
> > questions:
> > 
> > * do we keep `select` name for the utility function?
> > * I added docs and placed it under `docs/concepts/select/signals.md`, I ain't sure if it should be under concepts/select. Any ideas?
> 
> I think we can leave this utility function out for the first release of this. WDYT? There needs to be a final discussion on naming. I think that would be the correct place in the docs.

I think the utility function is super helpful, is inline with direction of replacing decorators with functions:

example @Input() => input, @Output => output, @ViewChild => viewChild, etc...

If we're looking to deprecate @Select it makes perfect sense to have select.

packages/signals/types/tests/produce-actions.lint.ts
markwhitfeld: hmmm... good point.

markwhitfeld:
> > > questions:
> > > 
> > > * do we keep `select` name for the utility function?
> > > * I added docs and placed it under `docs/concepts/select/signals.md`, I ain't sure if it should be under concepts/select. Any ideas?
> > 
> > 
> > I think we can leave this utility function out for the first release of this. WDYT? There needs to be a final discussion on naming. I think that would be the correct place in the docs.
> 
> I think the utility function is super helpful, is inline with direction of replacing decorators with functions, and times nicely with the latest [signal queries](https://angular.io/guide/signal-queries):
> 
> `@ContentChild` => `contentChild` and `@ContentChildren` => `contentChildren` `@ViewChild` => `viewChild` and `@ViewChildren` => `viewChildren`
> 
> If we're looking to deprecate `@Select` it makes perfect sense to introduce `select`.

This makes me so happy!!! There has been a lot of discussions about the naming of this utility, and I have always wanted to have it as `select`, advocating a signals first approach.
The hesitation has been that it doesn't match the method names of the `Store` methods.
Taking Angular's lead on this is a great motivation for this naming.
@arturovt I'm definitely happy for this utility to be called `select`.
@dmitry-stepanenko What do you think? You had raised the concern about alignment in the discussions.

markwhitfeld:
Just a subtle naming discussion point...
Do you think `produceSelectors` is the ideal name? It produces something from selectors. It produces "select"s. What do you think of the name `produceSelects`?
I'm not pushing for either, just looking for thoughts on it...
cc @arturovt @rhutchison @dmitry-stepanenko @Carniatto @joaqcid @profanis

markwhitfeld:
And, if we have a `select` utility to produce a single output as well as `produceSelectors` to produce a mapped output for selectors, do you think we should also have a `dispatch` utility for producing a single dispatch function from an action?
cc @arturovt @rhutchison @dmitry-stepanenko @Carniatto @joaqcid @profanis

markwhitfeld:
Given my last 2 questions, it leads me to ask:
What do you think about `produceDispatchers` as an alternative to `produceActions`?
cc @arturovt @rhutchison @dmitry-stepanenko @Carniatto @joaqcid @profanis

rhutchison:
> And, if we have a `select` utility to produce a single output as well as `produceSelectors` to produce a mapped output for selectors, do you think we should also have a `dispatch` utility for producing a single dispatch function from an action? cc @arturovt @rhutchison @dmitry-stepanenko @Carniatto @joaqcid @profanis

I did start to discuss with Artur.  `asDispatch` or `dispatch`?

```
export function asDispatch<TArgs extends any[]>(
  actionType: ActionDef<TArgs>
) {
  return (...args: TArgs) => inject(Store).dispatch(new actionType(...args));
}
```

```
@Component({...})
export class SomeComponent {
  user = select(AuthQueries.getUser);

  login = asDispatch(AuthActions.login);
}
````

rhutchison:
> Just a subtle naming discussion point... Do you think `produceSelectors` is the ideal name? It produces something from selectors. It produces "select"s. What do you think of the name `produceSelects`? I'm not pushing for either, just looking for thoughts on it... cc @arturovt @rhutchison @dmitry-stepanenko @Carniatto @joaqcid @profanis

> Given my last 2 questions, it leads me to ask:
What do you think about produceDispatchers as an alternative to produceActions?
cc @arturovt @rhutchison @dmitry-stepanenko @Carniatto @joaqcid @profanis

Maybe `toSelectMap` and `toDispatchMap`.  Would like to hear other's feedback before we merge.

arturovt:
> Given my last 2 questions, it leads me to ask: What do you think about `produceDispatchers` as an alternative to `produceActions`? cc @arturovt @rhutchison @dmitry-stepanenko @Carniatto @joaqcid @profanis

Since the function is generating dispatch functions rather than actions themselves a more appropriate name could be `produceDispatchers` or `createDispatchers`, or smth else.

rhutchison:
> I like `select` and `dispatch`. Although, including the `dispatch` here which returns an Observable seems a bit odd for a `@ngxs/signals` package. Would it rather belong in the core package?
> 
> If so, then what would that mean for the `produceDispatchers` (or whatever we call it). Does the same logic apply? Would that also shift to the main package?

dispatch should go in core, I think `produceSomething should stay here, `

> I like `select` and `dispatch`. Although, including the `dispatch` here which returns an Observable seems a bit odd for a `@ngxs/signals` package. Would it rather belong in the core package?
> 
> If so, then what would that mean for the `produceDispatchers` (or whatever we call it). Does the same logic apply? Would that also shift to the main package?

dispatch should definitely go in core, not sure about producer, but makes sense to question it

markwhitfeld:
In terms of naming, we have a few options:
- produceSelects
- makeSelectMap
- createSelectMap
- produceSelectMap
- asSelectMap

rhutchison:
> In terms of naming, we have a few options:
> 
> * produceSelects
> * makeSelectMap
> * createSelectMap
> * produceSelectMap
> * asSelectMap

I suggested toSelectMap and toDispatchMap, bc of the [rxjs-interop](https://angular.io/guide/rxjs-interop) naming convention

markwhitfeld:
> In terms of naming, we have a few options:
> - produceSelects
> - makeSelectMap
> - createSelectMap
> - produceSelectMap
> - asSelectMap

I think I prefer `createSelectMap` because it has similar naming to our other selector utilities (also starting with `create...`), which potentially makes it more discoverable too.

arturovt:
i'm fine with it too. so do we go with `createSelectMap` and `createDispatchMap`?

Carniatto:
Sorry I only saw the discussion now. Definitely the content of this PR is awesome, thanks Arthur for all the work. 

I was wondering if we should align with the terminology we already have for the selector utilities. See https://www.ngxs.io/advanced/selector-utils

dmitry-stepanenko:
Awesome PR! I think like the `createSelectMap` and `createDispatchMap` naming option.

> @dmitry-stepanenko What do you think? You had raised the concern about alignment in the discussions.

@markwhitfeld back in the [RFC](https://github.com/ngxs/store/discussions/1977#discussioncomment-7599500) we discussed that it will be `selectSignal`. At the same time I'm looking an it now and I think the `select` name aligns really well with all the new utilities angular has recently presented.

profanis:
Great work @arturovt!
My delayed input is that I like the `select` utility.
As for the dispatchers/actions, I agree with Artur that the function is generating Dispatchers and as such, I am more towards the `CreateDispatchers`.
I am afraid that the name `CreateActions` or `ProduceActions` is not that clear in terms of its underlying implementation.

Having said that, I would vote for the naming convention `Create*`

markwhitfeld (commented):
Great! So it looks like we have consensus on `select`, `createSelectMap` and `createDispatchMap`.
We are just missing tests for `createDispatchMap` in this PR.
The one other thing to be settled is the inclusion of the `dispatch` utility, and if both the dispatch utilities should exist in the `@ngxs/signals` package, or in the `@ngxs/store` package. 
Thoughts? @rhutchison @Carniatto @arturovt @joaqcid @dmitry-stepanenko @profanis

rhutchison:
> Great! So it looks like we have consensus on `select`, `createSelectMap` and `createDispatchMap`. We are just missing tests for `createDispatchMap` in this PR. The one other thing to be settled is the inclusion of the `dispatch` utility, and if both the dispatch utilities should exist in the `@ngxs/signals` package, or in the `@ngxs/store` package. Thoughts? @rhutchison @Carniatto @arturovt @joaqcid @dmitry-stepanenko @profanis

dispatch utility in separate PR https://github.com/ngxs/store/pull/2143

I am not sure where createDispatchMap should land, but I'm leaning to say keep where it is bc of the use-case with ngrx/signals (signalStore)

rhutchison:
I believe it's ready to merge

rhutchison (approved):

arturovt:
Merged. Can address any issues in subsequent PRs.

markwhitfeld:
I think we should include the `dispatch` utility in the `@ngxs/signals` package so that it is grouped together with the other utilities that follow the "signals style".
It is also easier to move something into the main package at a later stage, than it out of it.

PR #2146, opened 2024-03-28
docs: introduce deprecation root pages

profanis:
The reason for this change is the following:

Let's assume a developer works in a company where they decide to migrate to the version where the select decorator is deprecated. Although it's being deprecated they might need some time to convert their code. At the same time, they might need/want to read about the select decorator. If we completely remove the docs, it's harder for them to find the information they are looking for

image


arturovt (approved):

markwhitfeld (commented):
We need to update the docs references in the deprecation doc comments in the actual code too.

profanis:
> We need to update the docs references in the deprecation doc comments in the actual code too.

@markwhitfeld I am afraid I didn't get what needs to be fixed.

arturovt:
I think Mark refers to this code:

image

arturovt:
And for `children` too

profanis:
> We need to update the docs references in the deprecation doc comments in the actual code too.

I updated the doc references @markwhitfeld

PR #2149, opened 2024-03-29
fix(store): update `ActionDef` new to return `any`

arturovt:


markwhitfeld (commented):
Looks like this is still a work in progress (given the to do).
What is this PR for?

arturovt:
I removed `TReturn` from `ActionDef`, where it was `new (...args: TArgs): TReturn`, since it was redundant. By mistake, I changed it to `ThisType`, thinking it would infer the instance type. I now believe I just need to use `any` as the return type for `new`.

markwhitfeld:
What is the problem that you are looking to solve by removing the `TReturn`?
Or was is the motivation for removing it?

arturovt:
We have code where `ofActionDispatched` accepts a function parameter declared as shown in the test. Attempting to access `result.errorType` throws an error. Previously, `result` was inferred as `any` in this specific case. The use of `TReturn` wasn't resolving any issues, so I intend for `new()` to always return `any`.

markwhitfeld:
@arturovt this has regressed from previous type matching capabilities on the action type.
I will add a commit to fix this and demonstrate the capabilities I am referring to.

markwhitfeld:
@arturovt I have readded the correct typing and added tests that demonstrate the usages.
Hopefully this compiles, I did this in the github editor so I didn't have the build tooling.

PR #2150, opened 2024-03-29
fix(store): created maps properties should be enumerable

arturovt:


rhutchison (approved):

markwhitfeld (commented):
Yes! I had actually written a comment about this in the other PR, but didn't submit my review 😬

PR #2151, opened 2024-04-02
fix(store): do not rely on private token when checking whether zone is enabled

arturovt:


markwhitfeld (commented):
There is a code code change, and not just a docs change.

markwhitfeld (approved):

PR #2153, opened 2024-04-05
docs: explain `dispatch` utility

arturovt:


docs/concepts/store/README.md
```diff
@@ -114,6 +114,32 @@ export class ZooComponent {
...
+    route.params
+      .pipe(
+        map(params => params.invoiceId),
+        distinctUntilChanged(),
+        takeUntilDestroyed()
+      )
+      .subscribe(invoiceId => {
+        this.getInvoiceLines(invoiceId);
+      });
+  }
```
markwhitfeld, line 139: A simpler example would be better, so the utility is not lost in all the code.
Potentially:
```
// An action declared somewhere in your app
class Greet {
  static readonly type = 'Greet';
  constructor ( public greeting: string ) { }
}

// Then, in your component
export class MyComponent {
  greet = dispatch(Greet);

  constructor() {
    this.greet("Hello world!");
  }
}
```
It is good to mention the utility here, but I think it would be good to eventually have a page dedicated to these utilities.

markwhitfeld (approved):

PR #2154, opened 2024-04-05
docs: update life cycle guide and examples

arturovt:
Also removed some mentions about Angular internal stuff.

### DIFF
--- a/docs/concepts/state/life-cycle.md
+++ b/docs/concepts/state/life-cycle.md
@@ -4,7 +4,9 @@ States can implement life-cycle events.
 
 ## `ngxsOnChanges`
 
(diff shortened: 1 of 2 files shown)


docs/concepts/state/life-cycle.md
```diff
@@ -87,22 +88,20 @@ After creating the state by calling its constructor, NGXS calls the lifecycle ho
...
 
 ```ts
-// feature.module.ts
-@NgModule({
-  imports: [NgxsModule.forFeature([FeatureState])]
-})
-export class FeatureModule {}
+// some-data-access-library/index.ts
+export function provideDataAccessInvoiceLines() {
+  return provideStates([InvoiceLinesState]);
```
markwhitfeld, line 98: We have removed the example with the `NgxsModule.forFeature` usage.
We should possibly still maintain the Module version of the example (potentially as a collapsible section within the page?)

markwhitfeld (approved):

PR #2158, opened 2024-04-19
feat(store): improve action decorator types

ConnorUllmann:
## PR Checklist

## PR Type

```
[ ] Bugfix
[X] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

The `@Action` decorator does not enforce that the actions passed to the decorator match the payloads expected by the method it decorates.

A simple example is the following—the below action has a payload `{ value: string }` but the handler expects `{ value: number }` incorrectly. This does not produce an error.
```
class ActionString {
  static type = 'action1';
  constructor(public value: string) {}
}

...

@Action(ActionString)
handler(context: StateContext<any>, payload: { value: number }) {}
```

Issue Number: N/A

## What is the new behavior?

In the above example, a type error occurs. The case of multiple actions is also handled, requiring that a union of the payloads is accepted by the method.

```
[ ] Yes
[X] No
```

This change is likely to introduce type errors, but it is not intended to impact runtime behavior.

### DIFF
--- a/packages/store/src/decorators/action.ts
+++ b/packages/store/src/decorators/action.ts
@@ -14,9 +14,7 @@ type ActionToPayload<Action extends ActionType> =
  * Given a list of action classes, returns the union of their payloads.
  */
 type ActionsToPayload<Actions extends readonly ActionType[]> = {
-  [K in keyof Actions]: Actions[K] extends ActionDef<any, infer ActionPayload>
-    ? ActionPayload
-    : never;
+  [K in keyof Actions]: ActionToPayload<Actions[K]>;
 }[number];
 
 /**
--- a/packages/store/src/decorators/action.ts
+++ b/packages/store/src/decorators/action.ts
@@ -1,16 +1,59 @@
 import { ɵActionOptions, ɵensureStoreMetadata } from '@ngxs/store/internals';
 
-import { ActionType } from '../actions/symbols';
+import { ActionDef, ActionType } from '../actions/symbols';
 import { throwActionDecoratorError } from '../configs/messages.config';
+import { StateContext } from '../symbols';
 
 /**
- * Decorates a method with a action information.
+ * Given an action class, returns its payload.
  */
-export function Action(
-  actions: ActionType | ActionType[],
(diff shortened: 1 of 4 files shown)


packages/store/src/decorators/action.ts
```diff
@@ -1,16 +1,68 @@
...
+type ActionToPayload<Action extends ActionType> =
+  Action extends ActionDef<any, infer ActionPayload> ? ActionPayload : never;
+
+/**
+ * Given a list of action classes, returns the union of their payloads.
+ */
+type ActionsToPayload<Actions extends readonly ActionType[]> = {
+  [K in keyof Actions]: Actions[K] extends ActionDef<any, infer ActionPayload>
+    ? ActionPayload
+    : never;
```
markwhitfeld, line 19: Would this work here?
```suggestion
  [K in keyof Actions]: ActionToPayload<Actions[K]>;
```

packages/store/types/tests/action.lint.ts
```diff
@@ -390,3 +394,68 @@ describe('[TEST]: Action Operator Types', () => {
...
+
+  class ActionNumber {
+    static type = 'ACTION 2';
+    constructor(public c: number) {}
+  }
+
+  class ActionString2 {
+    static type = 'ACTION 3';
+    constructor(public a: string) {}
+  }
```
markwhitfeld, line 412: What this PR introduces is so much better than before, but if the method parameter has the class with the `static type` information then we could provide an even stricter matching.
Is it possible to also check if the correct action type is provided, not just the payload?
So it would be an error if we have:
```typescript
@Action(ActionString)
myAction(ctx: StateContext<any>, action: ActionString2) {
  // ....
}
```

Obviously if the parameter is specified with only the payload shape then we will have to fall back to the payload matching.

Note: We could do this even stricter typing as another PR after this.

markwhitfeld (commented):
Fantastic work!
My first comment is about a potential refactoring.
My second comment is a further idea, which I think would be another PR that build on this one.

markwhitfeld (approved):

PR #2174, opened 2024-06-11
docs: update `options.md` to add a warning about environments

arturovt:
This commit updates the options section of the documentation to include information on generating environments because they are no longer present by default.


markwhitfeld (commented):
👍

PR #2186, opened 2024-07-06
feat(store): add `withExperimentalNgxsPendingTasks`

arturovt:
In this commit, we provide the experimental app stability NGXS contributor as part
of the code. This is a fundamental feature that probably shouldn't be included in
recipes but should be part of our codebase. It is already being used in production
by apps that utilize zoneless server-side rendering. We expose it through the
`@ngxs/store/experimental` subpackage because exposing it from the core package would
be a breaking change, as it uses the experimental pending tasks service introduced in
Angular 18 (though we remain compatible with Angular 16).


markwhitfeld (approved):
Please add to the changelog

PR #2190, opened 2024-07-11
feat(store): debounce `selectSignal` by default

### LINKED ISSUE #2180: 🐞[BUG]: Dispatching an Action within an `effect` causes an error
### Affected Package
ngxs/store
### Is this a regression?

Yes, the previous version in which this bug was not present was: 3.8

### Description

When dispatching an action which will set the state from an angular signals `effect`, an error now occurs saying that   signals cannot be updated within an effect.

A workaround is to call the dispatch within an `untracked`, however the stack we get from angular is quite terrible, and having to do this for every dispatch is not a great devex.

## 🔬 Minimal Reproduction
https://stackblitz.com/edit/angular-17-starter-project-8ekglh?file=src%2Fmain.ts

## 🔥 Exception or Error
<pre><code>

NG0600: Writing to signals is not allowed in a `computed` or an `effect` by default. Use `allowSignalWrites` in the `CreateEffectOptions` to enable this inside effects.
</code></pre>

## Environment

<pre><code>
Libs:
- @angular/core version: 17.0.0
- @ngxs/store version: 18.0.0

Browser:
- [ ] Chrome (desktop) version XX
- [ ] Chrome (Android) version XX
- [ ] Chrome (iOS) version XX
- [ ] Firefox version XX
- [ ] Safari (desktop) version XX
- [ ] Safari (iOS) version XX
- [ ] IE version XX
- [ ] Edge version XX
 
For Tooling issues:
- Node version: XX  
- Platform:  

Others:

</code></pre>

arturovt:
In this commit, we make `selectSignal` debouncable by default. This means that changes originating from the state stream will not immediately update the signal but will instead occur as soon as possible (in the next microtask tick). By decoupling state signal updates from synchronous changes, we adhere to constraints on updates within effects. There are situations where actions dispatched within effects are synchronous, causing the state signal to also receive synchronous updates. This scenario can lead to an error indicating that signal writes are not permitted within effects.

Resolves #2180


packages/store/src/store.ts
```diff
@@ -84,7 +98,50 @@ export class Store {
...
+      this._selectableStateStream.pipe(debounceTime(0, asapScheduler)),
+      selectorFn
+    );
+
+    try {
+      assertInInjectionContext(() => {});
+      return toSignal(observable, { initialValue });
+    } catch {
+      return toSignal(observable, { initialValue, injector: this._injector });
+    }
```
arturovt, line 124: @markwhitfeld I don't think it's the right approach to do this. Basically, we're asserting the injection context, if the `assertInInjectionContext` doesn't throw an error, we're calling `toSignal` inside an injection context. Otherwise, we provide an explicit injector.

arturovt: Well, I actually would think that signals by default would require an injection context (not an explicit injector).

packages/store/src/store.ts
```diff
@@ -84,7 +98,50 @@ export class Store {
...
+      // are intended for use in templates or effects, as they always guarantee
+      // consistency with the latest signal value.
+      this._selectableStateStream.pipe(debounceTime(0, asapScheduler)),
+      selectorFn
+    );
+
+    try {
+      assertInInjectionContext(() => {});
+      return toSignal(observable, { initialValue });
+    } catch {
```
markwhitfeld, line 122: Why is this part necessary? 
Isn't it enough just to be explicit with the injector?
Was there something that failed?

packages/store/src/store.ts
markwhitfeld: I don't think that an injection context is necessary for the signals, but it would definitely be necessary for `toSignal` so that the `DestroyRef` can be used for unsubscribing.

packages/store/src/store.ts
arturovt: Not enough, because the store captures the "very root" injector, which doesn't resolve `DestroyRef`s tied to a component.

arturovt: Basically, when you do `selectSignal` in a AppComponent, it's using the root injector, which doesn't resolve the AppComponent's DestroyRef.

arturovt: Well, since the `injectDestroyRef` internal function is calling `getLView`, which is referring to the current LView being saved on the stack, I think I might be wrong... but just wanted to double check that exact scenario.

packages/store/tests/signal-per-state.spec.ts
```diff
@@ -0,0 +1,106 @@
...
+  })
+  class TestComponent {
+    value = signal(0);
+    number = this.store.selectSignal(NumberState.getNumber);
+
+    constructor(private store: Store) {
+      effect(
+        () => {
+          const value = this.value();
+          store.dispatch(new Set(value));
```
Carniatto, line 63: Maybe a silly comment, but calling this action `Set` can be confusing in terms of readability of this code. Because it can lead to confusion with the Set data structure of JavaScript. 

We could use `SetNumber`.

Please let me know if this kind of comment is irrelevant

packages/store/src/store.ts
markwhitfeld: What happens if you call `runInIjectionContext` and you are already in an injection context? 
Does it give you the same one? Maybe this would be the solution?

arturovt: `runInInjectionContext` requires the injector to be provided, which will be the same root injector.

packages/store/tests/signal-per-state.spec.ts
arturovt: I will update its name

packages/store/src/store.ts
```diff
@@ -116,12 +106,7 @@ export class Store {
       selectorFn
     );
 
-    try {
-      assertInInjectionContext(() => {});
-      return toSignal(observable, { initialValue });
-    } catch {
-      return toSignal(observable, { initialValue, injector: this._injector });
-    }
+    return toSignal(observable, { initialValue });
```
markwhitfeld, line 109: Oh nice, so this is able to determine the correct injection context at run time🎉.

markwhitfeld (approved):

simfyz:
@arturovt @markwhitfeld This is causing issues when using accessing signals inside dispatch. the signal value becomes null. Mostly when the default is null and values are set by an API call.

```
panda = select(ZooState.panda);

ngOnInit() {
    this.store.dispatch(new GetAnimals).pipe(takeUntilDestroyed(this.destroyRef)).subscribe(() => {
        const cutePanda = panda(); // value() is undefined here
    });
}
```

PR #2201, opened 2024-08-04
fix(store): use `computed` in `selectSignal`

arturovt:
Note: it is necessary to revert these changes and cease experimenting
with decoupling signal changes.

Previously, an issue was reported where calling `dispatch` from within an
effect caused an error indicating that signal writes are not allowed. We attempted
to resolve this by decoupling state signal changes through debouncing them, ensuring
that consumers always received signal updates asynchronously. However, we should no
longer pursue this approach because it fixes initial issues while introducing new ones.
It is the responsibility of developers to ensure a valid reactive context when updating
signals. The `dispatch` function can be wrapped in `untracked`, as it is common practice
to disallow updates to ANY signal within effects, not just NGXS state.

Debouncing signals disrupts zoneless server-side rendering because Angular is unaware
of tasks scheduled by the `debounceTime` operator. Consequently, content is serialized
before the `debounceTime` timer fires internally, which updates the signal and allows
the view to reflect the updated value.


markwhitfeld (commented):
I think there may be a better way to resolve this.
Let's pair on this to come to a good solution.

PS. Why were the tests for `packages/store/tests/select-signal-on-provider-levels.spec.ts` removed?

markwhitfeld:
I suspect that by reverting to this first solution,  updates to the state from an effect will not to propagate to the components when doing zoneless. This is because the use of `untracked` will prevent the triggering of the change detection from state changes... 
Could you add this as a test in your PR to verify?

PS. I have an idea that may work for all cases. I'll investigate tomorrow morning and then we can discuss.

PR #2219, opened 2024-09-17
perf(store): avoid going over states list every time action is dispatched

arturovt:
In this commit, I'm working on optimizing how we handle action dispatches by avoiding repeated traversal of the `states` list. Instead, we prepare a map each time a new state is added, allowing us to perform O(1) lookups by action type in the future. This approach reduces complexity and improves performance.

I've tested it with benchmark.js, and here are the results:

```ts
class Increment {
  static readonly type = 'Increment';
}

const states = Array.from({ length: 50 }).map((_, index) => {
  @State({
    name: `counter_${index + 1}`,
    defaults: 0,
  })
  @Injectable()
  class CounterState {
    @Action(Increment)
    increment(ctx: StateContext<number>) {
      ctx.setState((counter) => counter + 1);
    }
  }

  return CounterState;
});
```

```
store.dispatch() before changes x 3,435 ops/sec ±0.45% (65 runs sampled)
store.dispatch() after changes x 3,942 ops/sec ±1.21% (25 runs sampled)
```

### DIFF
--- a/.bundlemonrc.json
+++ b/.bundlemonrc.json
@@ -30,7 +30,7 @@
     },
     {
       "path": "./fesm2022/ngxs-store.mjs",
-      "maxSize": "100kB",
+      "maxSize": "101kB",
       "maxPercentIncrease": 0.5
     }
   ],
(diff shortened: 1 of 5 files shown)


markwhitfeld (approved):
Great improvement! I would love to know the performance change when you have 100 states and 10 actions each, and fire each of the 1000 actions. 

PS. I would like to chat with you about the direction I want to go with the action handler registration, because I think that it will bring much improvement to the code.

markwhitfeld:
You just need to update the max bundle size for the main bundle. Currently it is 100kb, but you can change it to 101kb to make this pass

PR #2224, opened 2024-10-01
refactor(store): add action registry

arturovt:
This adds a new internal construct, called the `ActionRegistry`.
The `ActionRegistry` facilitates the registration, and recall, of all handlers associated with an action type.
The default handler registered for each `@Action` within a state is now responsible for handling the return value of the user's handler within the state and the cancellation logic for the handler.


packages/store/src/actions/action-registry.ts
```diff
@@ -0,0 +1,40 @@
...
+@Injectable({ providedIn: 'root' })
+export class NgxsActionRegistry implements OnDestroy {
+  // Instead of going over the states list every time an action is dispatched,
+  // we are constructing a map of action types to lists of action metadata.
+  // If the `@@Init` action is handled in two different states, the action
+  // metadata list will contain two objects that have the state `instance` and
+  // method names to be used as action handlers (decorated with `@Action(InitState)`).
+  private readonly _actionTypeToHandlersMap = new Map<
+    string,
+    Set<InvokableActionHandlerMetaData>
```
markwhitfeld, line 18: This should be:
```suggestion
    Set<ActionHandlerFn>
```
Where the signature of the action handler is:
```typescript
type ActionHandlerFn = (action: Instance<ActionType>) => void | Promise<void> | Observable<unknown>;
```

packages/store/src/internal/state-factory.ts
```diff
@@ -424,29 +408,20 @@ export class StateFactory implements OnDestroy {
...
-      }
+      // This involves combining each individual action metadata with
+      // the state instance and the path—essentially everything needed
+      // to invoke an action. This eliminates the need to loop over states
+      // every time an action is dispatched.
+      const actionHandlers = actions[actionType].map(actionMeta => ({
+        ...actionMeta,
+        path,
+        instance
+      }));
```
markwhitfeld, line 420: And this would change to create the handler that encapsulates the creation of the context and the invocation of the function on the instance:

```suggestion
      const actionHandlers = actions[actionType].map(actionMeta => (action: Instance<ActionType>) => {
        const stateContext = this._stateContextFactory.createStateContext(actionMeta.path);
        return actionMeta.instance[actionMeta.fn](stateContext, action);
      });
```
So then all that needs to be done on dispatch is for the handlers for that action type to be called with the action instance as an argument.
The try catch and result handling will be the same, but just around the simpler handler call.

markwhitfeld (changes requested):
Great first step of the refactoring, but see my comments for the change in the handlers.

packages/store/src/actions/action-registry.ts
```diff
@@ -0,0 +1,117 @@
...
+    handlers.add(handler);
+    this._actionTypeToHandlersMap.set(type, handlers);
+
+    return () => {
+      const handlers = this._actionTypeToHandlersMap.get(type)!;
+      handlers.delete(handler);
+    };
+  }
+
+  private _createActionHandler(actionMeta: InvokableActionHandlerMetaData) {
```
markwhitfeld, line 60: As mentioned in the second comment of my previous review, the creation of the handler which in turn creates the state context, should stay outside of the action registry (in the StateFactory, for now).
Also the result handling logic should remain where it was in the StateFactory.

packages/store/src/actions/action-registry.ts
```diff
@@ -0,0 +1,117 @@
...
+import { StateContextFactory } from '../internal/state-context-factory';
+import { ɵNgxsInternalDispatchedActions } from './dispatched-actions';
+
+// `action: Instance<ActionType>`.
+type ActionHandlerFn = (action: any) => void | Promise<void> | Observable<unknown>;
+
+interface InvokableActionHandlerMetaData extends ɵActionHandlerMetaData {
+  path: string;
+  instance: any;
+}
```
markwhitfeld, line 25: This data structure should no longer be needed for the registry.
The registry should be super simple, only storing the handler functions, constructed and provided by the StateFactory.
The place in the StateFactory where the handler is created should contain enough access from its closure to be able to do all that it needs.

packages/store/src/actions/dispatched-actions.ts
```diff
@@ -0,0 +1,17 @@
...
+// This cancels the in-progress action handler once a new action of that
+// type is dispatched.
+export class ɵNgxsInternalDispatchedActions
+  extends Subject<ActionContext>
+  implements OnDestroy
+{
+  ngOnDestroy(): void {
+    this.complete();
+  }
+}
```
markwhitfeld, line 17: Should not be needed if we keep the logic in the StateFactory.

markwhitfeld (changes requested):
Please see comments in this review and the suggested code changes from the previous review.

markwhitfeld (approved):
Great work!!!!

hakimio:
@arturovt 
Sorry to bother you, just wanted to ask for a small favor.
[`@angular-ru/ngxs`](https://angular-ru.gitbook.io/sdk/ngxs/ngxs) library relies heavily on `ngxs/store` internals. After your recent `ngxs` refactoring, I have updated `@angular-ru/ngxs` to be compatible with `ngxs` `v18.1.3` (https://github.com/Angular-RU/angular-ru-sdk/pull/1527) but since I'm not too familiar with `ngxs` inner workings, I imagine the code might be suboptimal. When you have some time, could you take a look at [the PR](https://github.com/Angular-RU/angular-ru-sdk/pull/1527) and specifically at the [`@DataAction` decorator](https://github.com/Angular-RU/angular-ru-sdk/blob/845f53c08d352dc67d953446bb15daba37cf02e2/libs/ngxs/decorators/data-action/data-action.ts#L31) ([docs](https://angular-ru.gitbook.io/sdk/ngxs/data-action)) and let me know if the code could be improved?

arturovt:
I would first remove `NgxsDataInjector.ngZone?.run(...)` because it’s not required for NGXS. It doesn’t care whether dispatched events come from the Angular zone or not, as the state updates are decoupled from the dispatcher, and the new state always occurs within the Angular zone.

Since states are decorated with `@StateRepository`, that decorator could return a new constructor function, which is always called within the Angular injection context. You can capture the injector through `this[SomeInternalInjectorSymbol] = inject(Injector)`. The `descriptor.value` function now has access to `this[SomeInternalInjectorSymbol]` and can retrieve anything from the "app" injector as well as from the globally configured static injector. However, static properties wouldn't work properly for server-side rendered apps. If a new app is rendered for the first HTTP request and then a second request comes in, the newly bootstrapped app would overwrite static properties.

arturovt:
I think it might be better to find a way to register actions without relying on internal APIs, which are likely to change in the future. It seems that the `DataAction` decorator updates the metadata only when the method is called for the first time, but it should behave like the `Action` decorator and update the metadata immediately. This would eliminate the need to call `hydrateActionMetasMap`.

hakimio:
@arturovt thank you for your insights.
The `@Action` decorator is definitely a good example of how `@DataAction` could be implemented.

PR #2230, opened 2024-10-18
fix(store): memoize selectors which return NaN

arturovt:
In this commit, we update the memoize equality function to correctly check for `NaN`
values because we can't use the default `===` comparison operator, as `NaN` is not equal
to itself (`NaN === NaN` is false). Additionally, we have added a check using `Number.isNaN`.


markwhitfeld (commented):
Fantastic work!

PR #2231, opened 2024-10-18
fix(store): prevent writing to state once action handler is unsubscribed

arturovt:
In this commit, we update the implementation of action invocation. The key addition is that
we now prevent writing to the state whenever an action handler is unsubscribed (completed or
cancelled). Since we have a "unique" state context object for each action being invoked, we
can set its `setState` and `patchState` functions to no-ops, essentially making them do
nothing when invoked.

### DIFF
--- a/packages/store/src/internal/state-factory.ts
+++ b/packages/store/src/internal/state-factory.ts
@@ -23,8 +23,9 @@ import {
   filter,
   map,
   mergeMap,
-  shareReplay,
-  takeUntil
+  takeUntil,
+  finalize,
+  Observable
 } from 'rxjs';
 
 import { NgxsConfig } from '../symbols';
@@ -260,9 +261,9 @@ export class StateFactory implements OnDestroy {
   /**
    * Invoke actions on the states.
    */
-  invokeActions(action: any) {
+  private invokeActions(action: any): Observable<unknown[]> {
     const type = getActionTypeFromInstance(action)!;
-    const results = [];
+    const results: Observable<unknown>[] = [];
 
     // Determines whether the dispatched action has been handled, this is assigned
     // to `true` within the below `for` loop if any `actionMetas` has been found.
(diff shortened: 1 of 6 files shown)


packages/store/src/internal/state-factory.ts
```diff
@@ -380,12 +389,28 @@ export class StateFactory implements OnDestroy {
...
+              // was unsubscribed by `takeUntil` due to a new action being dispatched),
+              // we prevent writing to the state context.
+              finalize(() => {
+                stateContext.setState = noop;
+                stateContext.patchState = noop;
+              })
+            );
           } else {
-            result = of({}).pipe(shareReplay());
+            result = of(undefined).pipe(shareReplay());
```
arturovt, line 411: @markwhitfeld question: do we need `shareReplay()`?

packages/store/tests/issues/canceling-promises.spec.ts
```diff
@@ -0,0 +1,121 @@
...
+      await promiseAwaitReady;
+      recorder.push('after promise await ready');
+    }
+
+    @Action(MyActionThen, { cancelUncompleted: true })
+    handleActionThen() {
+      recorder.push('before promise then ready');
+      return promiseThenReady.then(() => {
+        recorder.push('after promise then ready');
+      });
```
markwhitfeld, line 47: What are we testing here?
There isn't anything that checks that operations on the state are nooped after cancellation.

markwhitfeld:
@arturovt I wasn't meaning to remove that test file, but rather to add tests that showcase the cancellation behavior by writing to the state. Potentially the action could write the log entries to the state as well as the recorder.
Then the normal test will show that they are the same, but then when you dispatch an action that results in cancellation, then you will see that the state doesn't receive the writes. 
You might want to add an `id` to the action so that the triggering action is apparent in the logs.

PS. I view the "don't write to state after unsubscribe" as a different set of tests to the cancellation tests (even though they use the same mechanism).

PR #2241, opened 2024-10-25
test(store): ensure state writes are allowed with fire & forget

arturovt:


markwhitfeld (commented):
I believe that the expected behaviour should be as follows...
(this is assuming that `cancelUncompleted: true` is present)
- If the fire and forget action is left to run without a new action coming in, then it should be allowed to do as many writes to the context that it needs to (ie. the StateContext remains writeable)
- but as soon as a second action comes in then the previous StateContext should become noop'ed so that it is no longer writeable.

Please add test cases to demonstrate both of these expected behaviours.

arturovt:
I'm afraid that might be difficult to implement. With your second approach, we would need to manually subscribe to the canceled observable:

```ts
const canceled = dispatched$.pipe(ofActionDispatched(action));

canceled.subscribe(() => {
  stateContext.setState = noop;
  stateContext.patchState = noop;
});
```

That is actually possible, but we need to unsubscribe from the canceled observable. However, the exact point at which we need to unsubscribe is unknown - perhaps when the action handler completes?

With the fire-and-forget approach, we never know if the action is still executing any functionality or when that functionality completes.

markwhitfeld:
The alternative here is that we could keep a reference to the previous
state context and do the noop when the next one comes though. Or, we could
have an abortController linked up to do the noop'ing and then keep a
reference to the last abort controller and call it when the new event comes
through.
(abort controller doesn't necessarily need to be exposed yet, but can be
used).

arturovt:
My thoughts on this question might be a bit disorganized, but I believe that fire-and-forget actions should not be cancellable. Instead, it should be the user’s responsibility to manage their cancellation by returning an observable, rather than subscribing to an observable and 'forgetting' it. I’d prefer not to capture previous state contexts because, if the fire-and-forget action never resolves, we wouldn't know when to release a reference to the previous state context.

PR #2244, opened 2024-11-02
feat(store): expose abort controller on state context

arturovt:


packages/store/src/internal/lifecycle-state-manager.ts
```diff
@@ -99,6 +99,7 @@ export class LifecycleStateManager implements OnDestroy {
   }
 
   private _getStateContext(mappedStore: MappedStore): StateContext<any> {
+    // Question: abort controller is not gonna be available for lifecycle hooks.
```
markwhitfeld, line 102: I think that the `abortSignal` could be available for lifecycle hooks.
For example, when we have a local state, the abort signal for the lifecycle hooks would be triggered when the state should be unloaded.

packages/hmr-plugin/src/internal/hmr-state-context-factory.ts
```diff
@@ -21,6 +21,7 @@ export class HmrStateContextFactory<T, S> {
    */
   public createStateContext(): StateContext<S> {
     return {
+      abortController: new AbortController(),
```
markwhitfeld, line 24: Not sure if this makes sense here, but we could just return:
```suggestion
      abortSignal: new AbortController().signal,
```

packages/store/src/internal/state-factory.ts
```diff
@@ -345,10 +346,14 @@ export class StateFactory implements OnDestroy {
     const { dispatched$ } = this._actions;
     for (const actionType of Object.keys(actions)) {
       const actionHandlers = actions[actionType].map(actionMeta => {
-        const cancelable = !!actionMeta.options.cancelUncompleted;
+        const abortController = new AbortController();
```
markwhitfeld, line 349: If cancellation is enabled we would retain a reference to the previous `abortController` for the handler and when the next action hits this handler we would call `abortController.abort()` on the previous `abortController` to trigger the cancellation of the previous handler and of any related work that has used the abort signal from the context.
When the abortSignal fires for some work in progress in the handler (like `fetch`), the handler would throw an error, we should catch this error and not let it propagate to the action error handlers.

markwhitfeld (changes requested):

PR #2245, opened 2024-11-03
refactor(store): use `Object.is` as default equality check

arturovt:
In this commit, we default to using `Object.is` as the equality check function (this function is also used by other frameworks). `Object.is` supports checking for `NaN` values internally. We have also removed the reset function, as it was unused. If needed in the future, it can be re-added.

### DIFF
--- a/packages/store/internals/src/memoize.ts
+++ b/packages/store/internals/src/memoize.ts
@@ -43,5 +43,10 @@ export function ɵmemoize<T extends (...args: any[]) => any>(
     lastArgs = arguments;
     return lastResult;
   }
+  (<any>memoized).reset = function () {
+    // The hidden (for now) ability to reset the memoization
+    lastArgs = null;
+    lastResult = null;
+  };
   return memoized as T;
 }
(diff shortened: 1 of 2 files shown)


packages/store/internals/src/memoize.ts
```diff
@@ -59,10 +43,5 @@ export function ɵmemoize<T extends (...args: any[]) => any>(
     lastArgs = arguments;
     return lastResult;
   }
-  (<any>memoized).reset = function () {
-    // The hidden (for now) ability to reset the memoization
-    lastArgs = null;
-    lastResult = null;
-  };
```
markwhitfeld, line 66: The removal of this function is unrelated to this PR.
And besides, what is the motivation to remove it?

PS. The PR also requires an update to the changelog

markwhitfeld (commented):
Great find!!!

packages/store/internals/src/memoize.ts
arturovt: I can remove it in another PR, but that part should not be included in the code unless it’s ever used. Extra work is bad, so unless there’s a valid argument for including that extra code, it shouldn’t be there.

PR #2246, opened 2024-11-03
refactor(store): tree-shake internal state tokens

arturovt:
In this commit, we remove internal injection token providers from the root providers list. Instead, we expose a function that provides these tokens.

### DIFF
--- a/packages/store/src/standalone-features/root-providers.ts
+++ b/packages/store/src/standalone-features/root-providers.ts
@@ -1,15 +1,9 @@
 import { APP_BOOTSTRAP_LISTENER, Provider, inject } from '@angular/core';
-import {
-  ɵStateClass,
-  ɵNGXS_STATE_CONTEXT_FACTORY,
-  ɵNGXS_STATE_FACTORY,
-  ɵNgxsAppBootstrappedState
-} from '@ngxs/store/internals';
+import { ɵStateClass, ɵNgxsAppBootstrappedState } from '@ngxs/store/internals';
 
 import { PluginManager } from '../plugin-manager';
 import { StateFactory } from '../internal/state-factory';
 import { CUSTOM_NGXS_EXECUTION_STRATEGY } from '../execution/symbols';
-import { StateContextFactory } from '../internal/state-context-factory';
 import { NgxsModuleOptions, ROOT_STATE_TOKEN, NGXS_OPTIONS } from '../symbols';
 
 /**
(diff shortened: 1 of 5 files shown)


markwhitfeld (commented):
I'm trying to understand the motivation here?
What is the practical benefit of the tree shaking? What problem are we solving?

arturovt:
Tokens are removed from the bundle.

arturovt:
Merging because I want to proceed with the next patch release. This change only reduces the bundle size by removing tokens that are not used in standard apps. These tokens were previously exposed as part of the ngxs-labs/data package.

packages/store/src/standalone-features/root-providers.ts
```diff
@@ -1,15 +1,9 @@
 import { APP_BOOTSTRAP_LISTENER, Provider, inject } from '@angular/core';
-import {
-  ɵStateClass,
-  ɵNGXS_STATE_CONTEXT_FACTORY,
-  ɵNGXS_STATE_FACTORY,
```
hakimio, line 5: @arturovt  `ɵNGXS_STATE_FACTORY` is used by `@angular-ru/ngxs` ☹

arturovt: They will need to use the `ɵprovideNgxsInternalStateTokens` in their providers list.

arturovt: You can still inject the factory through the token.

hakimio: Through which token?

arturovt: I can’t find any docs regarding angular-ru/ngxs, it’s all empty. Is there any link?

arturovt: What if we expose `StateFactory` as `ɵɵStateFactory` from the `@ngxs/store`? Those tokens will be removed since you'll then might be able to inject the `ɵɵStateFactory` directly.

hakimio: The is not about providing the factory. The library is using and injecting `ɵNGXS_STATE_FACTORY`:
https://github.com/Angular-RU/angular-ru-sdk/blob/845f53c08d352dc67d953446bb15daba37cf02e2/libs/ngxs/internals/services/ngxs-data-injector.service.ts#L22

How do you inject it without having access to the token?

arturovt: I don't understand the phrase "having access to the token." Why don't you have access to the token?

hakimio: ngxs no longer exports `ɵNGXS_STATE_FACTORY` and `ɵNGXS_STATE_CONTEXT_FACTORY` tokens after this PR was merged.

hakimio: Ok, my bad. For some reason instead of `import`, I thought `export` was removed. The tokens are still exported from `@ngxs/store/internals`. I'll try to use `ɵprovideNgxsInternalStateTokens()` to see if it fixes the issue.

hakimio: `ɵprovideNgxsInternalStateTokens()` fixes the issue. Sorry for the confusion.

PR #2247, opened 2024-11-04
perf(store): replace `instanceof Function` with `typeof`

arturovt:
`typeof` avoids prototype chain checks and type coercion, which is faster:

```
instanceof Function x 104,665,819 ops/sec ±4.90% (44 runs sampled)
typeof function x 169,867,932 ops/sec ±6.42% (55 runs sampled)
```

Btw fastest benchmarked checks were:

```js
1.
const returnValueType = typeof returnValue;
returnValueType[0] === 'f'

2.
const F_CHAR_CODE = 102;
const returnValueType = typeof returnValue;
(returnValueType.charCodeAt(0) ^ F_CHAR_CODE) === 0;

3.
const F_CHAR_CODE = 102;
const returnValueType = typeof returnValue;
returnValueType.charCodeAt(0) === F_CHAR_CODE;
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ $ npm install @ngxs/store@dev
 ### To become next patch version
 
 - Fix(store): Prevent writing to state once action handler is unsubscribed [#2231](https://github.com/ngxs/store/pull/2231)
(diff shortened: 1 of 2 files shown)


markwhitfeld (approved):
Nice perf tweak. You just need to update the changelog in this PR too.

PR #2248, opened 2024-11-09
refactor(router-plugin): mark selectors as pure

arturovt:
In this commit, we first remove the `@Selector()` decorator because we can use `createSelector` directly. This also removes `__decorate` calls for the `state` and `url` properties. Secondly, we mark these properties as 'pure.' The `/* @__PURE__ */` hint tells the bundler that `createSelector` has no side effects and that the output of this expression can be safely removed if unused. Therefore, if `state` or `url` is never accessed anywhere in our code, it can be tree-shaken.


Carniatto (approved):
LGTM

Issue #2268, opened 2024-11-20
🚀[FEATURE]: Support for Angular 19


rupekeshan:
### Relevant Package
This feature request is for @ngxs/store

### Description
Since Angular 19 was released on the 19. Nov, it should be supported by this package.

AndreiZaman:
Hello, when is planned to have the support of angular 19?

shadow1349:
Looks like the work of ready to go, it just needs a review.

AndreiZaman:
It will be nice to have it sooner rather than later

Sina7312:
So we will have it soon? We can't migrate to 19 because of it.

weewoo-meow:
It's merged on master, but on npm it is still not updated.

weewoo-meow:
> So we will have it soon? We can't migrate to 19 because of it.

Try version `18.1.6-dev.master-0ae2066`. It is bleeding-edge and you may encounter errors with the devtools package (as far as I tested it), but it would be a good start for you until the stable release appears.

arturovt:
NGXS v19 has been released.

shadow1349:
Thanks guys, this is awesome!

PR #2269, opened 2024-11-21
build: add Angular 19 support

arturovt:
Issue: #2268


rhutchison (approved):

.circleci/config.yml
```diff
@@ -178,6 +178,19 @@ jobs:
           paths:
             - app/integrations/hello-world-ng18/dist-integration
 
+  integration_ng19_tests:
```
markwhitfeld, line 181: This job needs to be added to the `build-workflow`.

markwhitfeld (changes requested):
Great work, but it seems that in the following file:
`packages\store\experimental\src\pending-tasks.ts`
We import `ExperimentalPendingTasks` from `@angular/core`.
In Angular 19, this is now available as `PendingTasks` from `@angular/core`.

This means that we can move the `withExperimentalNgxsPendingTasks` utility from the `@ngxs/store/experimental` sub package to `withNgxsPendingTasks` in the main `@ngxs/store` package (unless you can think of a more ideal place?).
Related to this change would be a version bump of NGXS v18 to NGXS v19 (especially because of this breaking change)🎉

arturovt:
I agree.

arturovt:
But we should wait until the Nx allows migrating to Angular 19 in order to update the workspace version too.

markwhitfeld:
What is the dependency on NX which would be blocking us?
Can we not release an NGXS v19 independently of the version of NX?
Surely we can do the workspace update separately.

PS. We should still keep the experimental subpackage around. 
We can export the `withExperimentalNgxsPendingTasks` with a deprecation notice. This function would just use the `withNgxsPendingTasks` function that has moved to the main package.

arturovt:
We can but not with the pending tasks change. The workspace is using Angular 18 which doesn't expose `PendingTasks` from the core.

markwhitfeld:
Ok, so we are blocked until this NX PR is merged:
https://github.com/nrwl/nx/pull/28847
That NX PR is, in turn blocked by a number of their dependencies, waiting for upgrade.

Is there any way that we can decouple the angular version that our libraries reference from the version of NX?
I'm wondering if there is something we can do with overrides in the `package.json`, or if we could use the aliasing functionality of tsconfigs to "redirect" the referenced angular version use in our libraries?
What we build with doesn't necessarily determine what our code references. Any ideas?

arturovt:
> We can export the withExperimentalNgxsPendingTasks with a deprecation notice.

This would throw an error that ` Entry point @ngxs/store has a circular dependency on itself.`

markwhitfeld:
The experimental subpackage already references things from the
***@***.***/store` library.
We would be moving the functionality to the core lib and referencing it
there (having the same direction of reference as the existing refs).

arturovt:
I have looked at what NgRx did, they upgraded workspace version w/o waiting for the Nx upgrade.

I upgraded the workspace version and had to update unit tests because components are standalone by default now.

markwhitfeld:
Great work!
There may be some usefulness to still have some components as `standalone:
false` where we are testing things about modules, but I think that most can
be converted.

arturovt:
@markwhitfeld what do we with ng16/17/18 integrations? They're not compatible anymore because we're using `PendingTasks`. Do we remove them completely because we want to bump the version to `ngxs@19` and make it compatible with `angular >= 19.0.0`?

tejasv-tagline:
Is anyone taking care of changes for this PR? Many users are waiting for Angular 19 support.

packages/store/experimental/src/pending-tasks.ts
```diff
@@ -1,31 +1 @@
...
-        const removeTaskFn = actionToRemoveTaskFnMap.get(ctx.action);
-        if (typeof removeTaskFn === 'function') {
-          removeTaskFn();
-          actionToRemoveTaskFnMap.delete(ctx.action);
-        }
-      }
-    });
-  });
-}
+export { withNgxsPendingTasks as withExperimentalNgxsPendingTasks } from '@ngxs/store';
```
markwhitfeld, line 1:
```suggestion
import { withNgxsPendingTasks } from '@ngxs/store';

/**
 * Required for correct functioning of SSR apps. 
 *
 * @deprecated
 * This experimental export is deprecated in favour of the non-experimental export. This is no longer an experimental feature, but is now available as `withNgxsPendingTasks` in the main `@ngxs/store` package.
 */
export function withExperimentalNgxsPendingTasks () {
  return withNgxsPendingTasks(); 
}
```

packages/store/src/pending-tasks.ts
```diff
@@ -0,0 +1,32 @@
+import { inject, PendingTasks } from '@angular/core';
+
+import { Actions, ActionStatus } from './actions-stream';
+import { withNgxsPreboot } from './standalone-features/preboot';
+
+/**
+ * This feature that contributes to app stability, * which is required during
```
markwhitfeld, line 7:
```suggestion
 * This feature that contributes to app stability, which is required during
```

packages/store/src/pending-tasks.ts
```diff
@@ -0,0 +1,32 @@
...
+ * server-side rendering. With asynchronous actions being dispatched and handled,
+ * Angular is unaware of them in zoneless mode and doesn't know whether the app is
+ * still unstable. This may prematurely serialize the final HTML that is sent to the client.
+ */
```
markwhitfeld, line 11:
```suggestion
  * Including `withNgxsPendingTasks` in your `provideStore` for your SSR app will resolve the above issue.
  */
```

packages/store/experimental/tests/preboot-wait-stable.spec.ts
markwhitfeld, line 1: We should add this test to the main package's tests.

markwhitfeld (changes requested):

markwhitfeld (approved):
Great work!!!!
Let's get this released!

PR #2278, opened 2024-12-17
fix(store): add root store initializer guard

arturovt:
In this commit, we are adding the root store initializer guard, which ensures that
`provideStore()` is not accidentally called multiple times (e.g., in feature-level
providers such as route providers).


markwhitfeld (approved):
🎉 LGTM

Squixx:
Nice! Thanks for adding this

PR #2294, opened 2025-01-12
refactor(router-state): change `@Selector` to `createSelector`

arturovt:
In this commit, we switch from using the selector decorator to allow it to be dropped in production
for apps that don't use it at all.

Breaking change: To get the `RouterState.state`, it is now required to call the `state`
function — `select(RouterState.state())`. Previously, it was possible to provide a generic type for
the router state, e.g., `select(RouterState.state<CustomRouterState>)`, but this is not possible
if `state` were a property.

### DIFF
--- a/packages/router-plugin/src/router.state.ts
+++ b/packages/router-plugin/src/router.state.ts
@@ -10,7 +10,7 @@ import {
   NavigationEnd,
   Event
 } from '@angular/router';
-import { Action, Selector, State, StateContext, StateToken, Store } from '@ngxs/store';
+import { Action, createSelector, State, StateContext, StateToken, Store } from '@ngxs/store';
 import {
   NavigationActionTiming,
   ɵNGXS_ROUTER_PLUGIN_OPTIONS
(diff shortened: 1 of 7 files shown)


packages/router-plugin/src/router.state.ts
```diff
@@ -85,17 +85,15 @@ export class RouterState {
...
-  static state<T = RouterStateSnapshot>(state: RouterStateModel<T>) {
-    // The `state` is optional if the selector is invoked before the router
-    // state is registered in NGXS.
-    return state?.state;
+  static state<T = RouterStateSnapshot>() {
+    return createSelector([ROUTER_STATE_TOKEN], (state: RouterStateModel<T>) => {
+      // The `state` is optional if the selector is invoked before the router
+      // state is registered in NGXS.
+      return state?.state;
+    });
```
markwhitfeld, line 93: @arturovt Isn't this a breaking change because it is updating the public API?

markwhitfeld: Oh, I see that now that this is mentioned in the PR description. I think that there needs to be a note in the changelog about this breaking change.
Or do you think that there is an alternative like a separate `getRouterState<T>()` dynamic selector exposed by the plugin, and we leave the `RouterState.state` as a property.
OR is there the possibility of:
`RouterState<RouterStateSnapshot>.state` where the T on `RouterState` has a default so that `RouterState.state` still works.

markwhitfeld: Hmmm.... I think I have figured out how to make this a non breaking change:
```suggestion
  static get state() {
    return createSelector([ROUTER_STATE_TOKEN], <T = RouterStateSnapshot>(state: RouterStateModel<T>) => {
      // The `state` is optional if the selector is invoked before the router
      // state is registered in NGXS.
      return state?.state;
    });
```
I made a StackBlitz example here:
https://stackblitz.com/edit/vitejs-vite-rrj1efj4?file=src%2Fcounter.ts

arturovt: Doesn't work, unfortunately:

image

PR #2305, opened 2025-02-17
feat(store): action status "PreHandler"

XavierDupessey:
Before this change, it was not possible to prepare for an action before it was processed by NGXS.

This is because the DISPATCHED event is emitted only after the action handlers have been executed, which is unclear from the documentation: ofActionDispatched: This will ONLY grab actions that have just been dispatched

Since changing this behavior would break the public API, this update introduces a new event, PreHandler, which allows reacting to an action before any handler is executed.

## PR Checklist

Doc will be adapted if the PR is accepted (page to update is https://www.ngxs.io/concepts/actions/action-handlers)

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Currently, it was not possible to prepare for an action before it was processed by NGXS.

## What is the new behavior?

The action PreHandler is emitted before NGXS executes any handler.

```
[ ] Yes
[x] No (except if triggering more ofAction(MyAction) is a breaking change?
```

### DIFF
--- a/packages/store/src/actions-stream.ts
+++ b/packages/store/src/actions-stream.ts
@@ -9,6 +9,7 @@ import { InternalNgxsExecutionStrategy } from './execution/internal-ngxs-executi
  * Status of a dispatched action
  */
 export const enum ActionStatus {
+  PreHandler = 'PRE_HANDLER',
   Dispatched = 'DISPATCHED',
   Successful = 'SUCCESSFUL',
   Canceled = 'CANCELED',
--- a/packages/store/src/internal/dispatcher.ts
+++ b/packages/store/src/internal/dispatcher.ts
@@ -62,6 +62,8 @@ export class InternalDispatcher {
     const prevState = this._stateStream.getValue();
     const plugins = this._pluginManager.plugins;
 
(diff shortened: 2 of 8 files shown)


XavierDupessey:
FYI @markwhitfeld this is the new feature I mentioned earlier :)

markwhitfeld (commented):
Hi @XavierDupessey 
Have you thought about implementing this callback through a dedicated plugin?
This plugin could expose a `BeforeDispatch` stream that is similar to the `Actions` stream.
The Plugin API allows you to intercept each action call before it calls the handlers.

XavierDupessey:
No, I hadn't considered implementing it as a plugin first. But you're right, it's very straightforward using the `handle` function provided by `NgxsPlugin` from `@ngxs/store/plugins`.

If you believe there's no need to expose this event in the core, I'm totally fine with going the plugin route instead! 🙂

markwhitfeld:
I'll have a chat with the team about it, but for now I think that the plugin route is best.
If you would like to contribute a page to our "Recipes" section of the docs with the code that you create and some instructions, that would be a great way to share the pattern with the community.

PR #2317, opened 2025-04-09
chore: migrate executionStrategy

profanis:
In this PR I introduced the migration scripts and moved some schematics utils to a separate directory


packages/store/package.json
```diff
@@ -5,7 +5,12 @@
   "sideEffects": false,
   "peerDependencies": {
     "@angular/core": ">=19.0.0 <20.0.0",
-    "rxjs": ">=7.0.0"
+    "rxjs": ">=7.0.0",
+    "ts-morph": "21.0.1"
```
profanis, line 9: @arturovt is it OK to have ts-moprh as a peer dep or would it be better to get rid of this dependency completely?

packages/store/migrations/src/19_1_0/index.ts
```diff
@@ -0,0 +1,218 @@
...
+  ) {
+    const args = callExpression.getArguments();
+    if (!args.length) {
+      _context.logger.info(`Migrating empty forRoot in ${path}`);
+      migrateEmptyForRoot(callExpression);
+    } else if (args.length === 1) {
+      _context.logger.info(`Migrating forRoot with states in ${path}`);
+      migrateForRootWithStates(callExpression);
+    } else if (args.length === 2) {
+      _context.logger.info(`Migrating forRoot with states and exiting properties in ${path}`);
```
arturovt, line 63:
```suggestion
      _context.logger.info(`Migrating forRoot with states and existing properties in ${path}`);
```

packages/store/migrations/src/19_1_0/index.ts
```diff
@@ -0,0 +1,218 @@
...
+    const args = callExpression.getArguments();
+
+    if (args.length === 1) {
+      // args.length === 1 --> provideStore([])
+      _context.logger.info(`Migrating provideStore with states in ${path}`);
+      migrateForRootWithStates(callExpression);
+    } else if (args.length === 2) {
+      // args.length === 2 --> provideStore([], {foo:'bar'})
+      _context.logger.info(
+        `Migrating provideStore with states and exiting properties in ${path}`
```
arturovt, line 96:
```suggestion
        `Migrating provideStore with states and existing properties in ${path}`
```

packages/store/package.json
arturovt: I think it's better to remove it as I'm pretty sure people wouldn't be happy having it as a peer.

packages/store/migrations/src/19_1_0/index.ts
profanis: fixed

packages/store/migrations/src/19_1_0/index.ts
profanis: fixed

packages/store/package.json
profanis: I removed the ts-morph

packages/store/migrations/src/19_1_0/index.ts
```diff
@@ -0,0 +1,275 @@
...
+    // args.length === 1 --> provideStore([])
+    context.logger.info(`Migrating provideStore with states in ${path}`);
+    migrateForRootWithStates(provideStoreImport.arguments[0].end, path, changes);
+  } else if (args.length === 2) {
+    // args.length === 2 --> provideStore([], {foo:'bar'})
+    context.logger.info(
+      `Migrating provideStore with states and existing properties in ${path}`
+    );
+    migrateForRootWithExistingOptions(provideStoreImport.arguments, path, changes);
+  }
```
arturovt, line 90: Would it also migrate `provideStore` with plugins after state?

```ts
provideStore([], withNgxsRouterPlugin(), withNgxsStoragePlugin(), ....)
```

profanis: hmm I don't think this will work.
Is this how the migrated code should look like?

**Before**
```ts
provideStore([], withNgxsRouterPlugin(), withNgxsStoragePlugin())
```

**After**
```ts
provideStore([], withNgxsRouterPlugin(), withNgxsStoragePlugin(), {executionStrategy:DispatchOutsideZoneNgxsExecutionStrategy})
```
---
**Before**
```ts
provideStore([], withNgxsRouterPlugin(), withNgxsStoragePlugin(), {foo:'bar'})
```

**After**
```ts
provideStore([], withNgxsRouterPlugin(), withNgxsStoragePlugin(), {foo:'bar', executionStrategy:DispatchOutsideZoneNgxsExecutionStrategy})
```

arturovt: It should be:

```ts
proideStore([], OPTIONS, ....plugins)
```

profanis: I did some changes to take under consideration the plugins. 
I am not sure if having a function to provide the options would make the migration easier, but for sure it would be better in terms of API.

If we want to go that route, we should also migrate the existing options to this function. This could be a separate migration which could run after the executionStrategy migration

arturovt (approved):

PR #2329, opened 2025-05-13
feat(store): implement `attachAction`

arturovt:


markwhitfeld (commented):
Great work so far.
Why have you gone for the API `experimentalBuildActionHandler(...).attach()` as opposed to `attachAction(...)`?

arturovt:
I didn’t really understand how attach action API should look like. Like if it’s a function that returns a function (attach) that also returns a function (detach) seems complex.

joaqcid:
here's what the attach-action plugin looks like currently

```
attachAction(STATE, ACTION, (_stateContext, action) => {
return // can return sync or async to control action flow
});
```

```
const META_OPTIONS_KEY = 'NGXS_OPTIONS_META';

export function attachAction<S, A>(
  storeClass: any,
  action: ActionType,
  fn: (ctx: StateContext<S>, action: A) => any,
  options?: ActionOptions
): void {
  if (!storeClass[META_OPTIONS_KEY]) {
    throw new Error('storeClass is not a valid NGXS Store');
  }

  const methodName = getActionMethodName(action);

  storeClass.prototype[methodName] = function(_state: any, _action: any): any {
    return fn(_state, _action);
  };

  Action(action, options)({ constructor: storeClass }, methodName, null as any);
}

const getActionMethodName = (action: ActionType) => {
  const actionName = action.type.replace(/[^a-zA-Z0-9]+/g, '');
  return `${actionName}`;
};
```

arturovt:
This implementation is not patching the class, but rather uses the internal DI classes to resolve the state class path in the state graph and creates a state context for that state. The `attach()` is lazy because it also requires an injection context to run.

markwhitfeld:
@arturovt I was thinking of alternative implementations:
1) have a public `ActionRegistry` service exposed (different to our internal action registry), and it exposes the `attachAction` function as previously described.
2) alternatively, the `attachAction` utility function could send the arguments it has received (state, action and handler function) to a static queue. Then the next time a dispatch is called, it checks to see if any items are in this queue and registers them 'just in time', removing them from that queue, before the dispatch is run.

What do you think?

arturovt:
I like the second approach with the queue and registration when the dispatch is called.

markwhitfeld:
I think that the following API would make things much simpler and more explicit:
```ts
const actionDirector = inject(ActionDirector);
// somewhere appropriate in the code
actionDirector.attachAction(
  CountriesState, 
  AddCountry, 
  (ctx: StateContext<string[]>, action: AddCountry) => {
    ctx.setState(countries => [...countries, action.country]);
  }
);
```
and to build on this example with more a complex usage and side effects:
```ts
const actionDirector = inject(ActionDirector);
const countryRepo = inject(CountryRepo);
// somewhere appropriate in the code
const detachMe = actionDirector.attachAction(
  CountriesState, 
  AddCountry, 
  async (ctx: StateContext<string[]>, action: AddCountry) => {
    // Note: this is not run inside an injection context, just like in the states
    ctx.setState(countries => [...countries, action.country]);
    await countryRepo.Add(action.country);
  }
);
const destroyRef = inject(DestroyRef);
destroyRef.onDestroy(() => detachMe());
```

We could also provide another function (`effect`) that is only for side effects that don't influence state:
```ts
const actionDirector = inject(ActionDirector);
const countryRepo = inject(CountryRepo);
// somewhere appropriate in the code
const stopEffect = actionDirector.effect(
  AddCountry, 
  async (action: AddCountry) => {
    // Note: this is not run inside an injection context, just like in the states
    await countryRepo.Add(action.country);
  }
);
const destroyRef = inject(DestroyRef);
destroyRef.onDestroy(() => stopEffect());
```

With respect to the naming of these functions, I have thought of a few options:
1. `attachAction`, `effect`
2. `onAction`, `effect`
3. `reaction`, `effect`

What do you think?

arturovt:
I think `effect` is an actual _effect_ (like we practically tend to think of), not for state mutations. `reaction` is conceptually nice, but too abstract. `attachAction` is a bit low-level sounding. Previously, I had `registerHandler` in mind because it's basically what it does. But I don't have any balance between all of them.

packages/store/tests/build-action-handler/build-action-handler.spec.ts
```diff
@@ -0,0 +1,32 @@
+import { runInInjectionContext } from '@angular/core';
+import { TestBed } from '@angular/core/testing';
+import { provideStore, Store } from '@ngxs/store';
+
+import { AddCountry } from './fixtures/actions';
+import { CountriesState } from './fixtures/state';
+
+describe('buildActionHandler', () => {
```
markwhitfeld, line 8: The test file and folder should be renamed to `action-director`, and the test describe adjusted:
```suggestion
describe('ActionDirector', () => {
```

packages/store/tests/build-action-handler/fixtures/add-country-action-handler.ts
```diff
@@ -0,0 +1,13 @@
+import { inject } from '@angular/core';
+import { ActionDirector } from '@ngxs/store';
+
+import { COUNTRIES_STATE_TOKEN } from './state';
+import { AddCountry } from './actions';
+
+export const addCountryActionHandler = () => {
+  const actionDirector = inject(ActionDirector);
```
markwhitfeld, line 8: Just a thought regarding your implementation of this lazy loaded action handler...
You could pass the reference to the `ActionDirector` as an argument to this function. Then it would remove the need to run it in an injection context.
Up to you whether you change it or leave it. Doesn't matter to me.
```suggestion
export const addCountryActionHandler = (actionDirector: ActionDirector) => {
```

markwhitfeld (changes requested):

markwhitfeld:
I have made some changes and added documentation for this feature.
@joaqcid and @profanis please could you have a read through the docs I created for this and give me your feedback?

profanis:
> I have made some changes and added documentation for this feature.
> @joaqcid and @profanis please could you have a read through the docs I created for this and give me your feedback?

The explanation is easy to follow 💯

docs/concepts/actions/dynamic-action-handlers.md
```diff
@@ -0,0 +1,96 @@
+# Dynamic Action Handlers
+
+Sometimes you need to attach action handlers dynamically after state initialization. For example, you might want to:
+
+- Add action handlers conditionally based on runtime conditions
+- Add action handlers for lazy-loaded modules
+- Add temporary action handlers that can be removed later
```
joaqcid, line 7: -  Add action handler to fine-grain control when an action is completed

joaqcid:
> I have made some changes and added documentation for this feature. @joaqcid and @profanis please could you have a read through the docs I created for this and give me your feedback?

i added a couple of comments, but overall looks great!

thanks!!

docs/concepts/actions/dynamic-action-handlers.md
```diff
@@ -0,0 +1,96 @@
...
+The `ActionDirector` allows you to attach action handlers to a state at any point after initialization and gives you the ability to detach them when no longer needed.
+
+### Attaching an Action Handler
+
+```ts
+import { ActionDirector, createSelector } from '@ngxs/store';
+import { inject, Injectable } from '@angular/core';
+
+// State token
+const COUNTRIES_STATE_TOKEN = new StateToken<string[]>('countries');
```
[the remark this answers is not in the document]
markwhitfeld: Only state tokens, because it is type safe.

You can hack it for your needs by creating an object that satisfies the StateToken interface and the returns the same name as the state class from a getName() function.

PR #2359, opened 2025-07-08
feat(store): add `withNgxsNoopExecutionStrategy`

arturovt:


markwhitfeld:
@arturovt do you think we should name this `withNgxsZonelessExecutionStrategy`
or even `withNgxsZoneless` to be more explicit about when it should be used?

markwhitfeld:
@arturovt in addition to the above, do you think that we should also re-institute some of the tests as well as the dev mode warning removed by PR #2325 ?

arturovt:
It's not related to zoneless, because they want to use it in a zone.js app.

arturovt:
In zoneless both strategies act like a "noop".

PR #2399, opened 2026-03-16
feat(store): support async/await syntax for dispatch

arturovt:
Wraps the result of `dispatch()` in a dual-natured return type that extends `Observable` and implements `PromiseLike<void>`. This allows callers to either subscribe reactively or use `await` without any API change — the dispatch function signature stays the same and the caller's syntax alone determines the behavior.


rhutchison (approved):

markwhitfeld (approved):
Nice feature 👍

PR #2424, opened 2026-03-31
fix(store): return original reference from updateItems when no elements match

arturovt:
updateItems always returned a new array clone, even when no element matched the predicate. This caused memoized selectors to see a changed reference and re-evaluate on every no-op call.

Track whether any element was actually updated and return the original array reference unchanged when the predicate matched nothing.


nineninesevenfour:
Hi! You might have a look at `operators/tests/update-items.spec.ts`, which seems to have started failing with this MR.

PR #2426, opened 2026-04-01
feat(store): expose ɵgetTypedNgxsStateFactory for internal third-party usage

arturovt:
Adds `ɵgetTypedNgxsStateFactory`, a type-safe accessor that surfaces `StateFactory`'s private fields (`_states`, `_statesByName`, `_statePaths`) without requiring consumers to cast or use bracket notation themselves.

The return type is declared explicitly on the function signature so that third-party libraries can rely on a stable, typed contract rather than the inferred shape of private internals.


markwhitfeld (commented):
What is this for?

PR #2449, opened 2026-06-08
ci: add on-demand dev publish workflow with manual approval gate

arturovt:


markwhitfeld (changes requested):
Good start, but this won't work because npm trusted publishing has been configured and that requires the publishing to only happen from a single workflow.
I will make the adjustments to the PR.

markwhitfeld (approved):

PR #2474, opened 2026-07-21
refactor(router-plugin): move derived selectors to tree-shakable exports, add named-outlet variants

arturovt:
Static class properties always ship in the bundle once the owning class is referenced anywhere (RouterState always is, via withNgxsRouterPlugin()), so bundlers can't tree-shake individual unused selectors off it. queryParams/queryParamMap/fragment/params/ paramMap/data/title were added post-v22 and aren't released yet, so move them off RouterState into router-selectors.ts as standalone routerXxx exports instead.

Also add new selectors for named outlets — routerParamsForOutlet, routerParamMapForOutlet, routerDataForOutlet, routerTitleForOutlet — which resolve params/paramMap/data/title for a given outlet's branch, since the existing primary-outlet selectors only ever followed the first child at each level.

Mark the top-level createSelector() calls /* @__PURE__ */ so unused ones can actually be dropped. RouterState.state()/.url are untouched since those predate this and are already released.

### DIFF
--- a/packages/router-plugin/src/public_api.ts
+++ b/packages/router-plugin/src/public_api.ts
@@ -1,6 +1,19 @@
 export { NgxsRouterPluginModule, withNgxsRouterPlugin } from './router.module';
 export { ROUTER_STATE_TOKEN, RouterState, type RouterStateModel } from './router.state';
 export {
+  routerQueryParams,
+  routerQueryParamMap,
+  routerFragment,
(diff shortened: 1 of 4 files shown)


markwhitfeld (approved):
Would it help if we marked `createSelector` itself as pure? 
Also, I need to chat to you about the naming of the selectors

